PR #64, opened 2015-03-10
Fixes to Array.prototype models

msridhar:
These were missing from prologue.js


msridhar:
I also added a fixed model for `Array.prototype.pop`

msridhar:
The tests failed due to some issue with the Dalvik tests.  I confirmed that all the JS tests pass on my machine.

msridhar:
FWIW, I think the build failures are due to Travis and/or Android issues; the JS tests pass for me.

PR #65, opened 2015-04-02
Use Arrays.toString() instead of the default array toString() method.

collinw:
The former will include the contents of the array, while the latter only includes the object's identity.

This will allow WALA to be compiled using Google's error-prone compiler (https://github.com/google/error-prone).


msridhar:
It looks like we have some flakiness on Travis that is causing the CI build to fail.  This change looks good to me, so I'll merge.

PR #70, opened 2015-05-27
Fixed JUnit launcher of com.ibm.wala.cast.js.test

Sheak90:
The JUnit launcher of com.ibm.wala.cast.js.test could not access the resource because the runtimeClasspathEntry did not linked to the data package.


msridhar:
Thanks!

PR #73, opened 2015-06-15
changes to Android support over the last months

mohrm:
Hi!

Here are our fixes and changes to the com.ibm.wala.dalvik project which we performed since since the merge in August '14.
We will switch our fork from jgf/WALA to this one so that it is really a fork of the main WALA repo, which should make contribution a lot easier. This could be a first step to eventually switching over to the main WALA repo.


mohrm:
Yes, I already saw - there seem to be some more issues. I'll try again...

PR #75, opened 2015-06-18
Changes from the Joana team over the last months

mohrm:
This is the second trial, hopefully it works now :)

Hi!

Here are our fixes and changes (mainly) to the com.ibm.wala.dalvik project which we performed since since the merge in August '14.
We will switch our fork from jgf/WALA to this one so that it is really a fork of the main WALA repo, which should make contribution a lot easier. This could be a first step to eventually switching over to the main WALA repo.


msridhar:
Looks like we still have some Maven build errors, possibly due to dependencies on com.google libraries that aren't being pulled in (maybe shouldn't be pulled in?).

mohrm:
As far as I see, the problem is that com.ibm.wala.scandroid needs the Google guava lib und relies on com.ibm.wala.dalvik providing it. com.ibm.wala.dalvik does not need guava, so in one of my commits I removed the dependency, unfortunately forgetting to run maven on WALA as a whole...
For now, I am adding this dependency again but we maybe should think about moving this dependency from *.dalvik to *.scandroid...

msridhar:
@mohrm maybe you can just move the dependency now?  Might as well do the cleanup while we remember.

mohrm:
Good idea! I will do that

PR #84, opened 2015-07-30
Remove com.ibm.wala.cast's dependence on Eclipse

msridhar:
This should make creation of Maven Central JARs for the JS frontend easier.

PR #85, opened 2015-07-31
Remove Eclipse dependence from cast.js(.rhino)

msridhar:
Similar to PR #84.  @juliandolby I changed the JSDT code in a way that seems to work (the regression tests pass); loading files through its own `Activator` type instead of the old `JavaScriptPlugin` class seems to find `prologue.js` just fine.  But if you could eyeball the changes that would be great.


juliandolby:
Manu,

  The changes look nice.  I have struggled with those regression tests a
bunch of times, and if using the JSDT plugin's activator works, that is
great.

	-- Julian

Issue #100, opened 2016-06-03
SecurityExceptions when analyzing JARs with broken/expired signatures


sewe:
AFAICT, WALA uses the single-argument constructors of `JarFile` and `JarInputStream` throughout. Unfortunately, this causes a `SecurityException` whenever WALA reads a JAR with broken or expired signatures. The latter is particularly annoying when the JAR was correctly signed but the certificate has since expired and no [signature timestamp](http://docs.oracle.com/javase/7/docs/technotes/guides/security/time-of-signing.html) has been included.

IMHO, WALA should not verify any signatures.

msridhar:
I agree about not doing signature verification. A pull request would be
welcome! Otherwise I'll look when I can.

sewe:
> I agree about not doing signature verification. A pull request would be welcome! Otherwise I'll look when I can.

I’ll prepare a pull request soon. We already have a patch that fixes this for our 1.3.4-ish fork of WALA, which hopefully will apply cleanly.

sewe:
Created the pull request (sorry about the noise with 21c5971f9da770339363ad440e5bc1fc50fcb839; its commit message still referred to our internal bug tracker) for commit e2e5bcf43567ad2871d15a22095d63d8f0e54164.

I went through all uses of `JarFile` and `JarInputStream` constructors and tried to figure out whether the JAR being opened is a JAR that is about to be analyzed (which was always the case); hence, every constructor invocation in the codebase was changed.

msridhar:
Fixed by e2e5bcf43567ad2871d15a22095d63d8f0e54164

PR #101, opened 2016-06-08
SecurityExceptions when analyzing JARs with broken/expired signatures

sewe:
See <https://github.com/wala/WALA/issues/100>


msridhar:
Looks good to me.  Thanks again!

PR #105, opened 2016-09-21
Added build script step to copy dx.jar from Android SDK.

mherzberg:
The proposed additions to the ant script automatically copy the dx.jar from the local Android SDK folder into the `com.ibm.wala.dalvik.test/lib` folder, if ANDROID_HOME is set. This is basically the procedure from [here](https://groups.google.com/forum/#!msg/wala-sourceforge-net/cBYsfEvYVG0/Ua52dyQQU-YJ). As the dx.jar is needed for a clean compile of the whole project, the user does not have to copy it manually anymore.


msridhar:
LGTM.  Thanks!

PR #106, opened 2016-10-19
Nodejs Support

Pfarrer:
Two projects are introduced, one contains the source files and the other one the tests and test data.

Class NodejsRequireTargetSelector is used to detect require calls. Using string propagation, the parameter of the require call is retrieved. Next, the corresponding file is loaded using either an instance of class NodejsRequiredCoreModule or NodejsRequiredSourceModule. Nodejs core libraries are included in the dat/core-modules folder.

Any file is wrapped into the dat/module-wapper.js which behaves exactly like the actual Nodejs wrapper. When a JSON file is required, dat/json-wapper.js is used.


juliandolby:
If we can download the code in a script, that would be simplest.  However, given the flexible license of nodejs, we also ought to be able to apply the same trick we used for contributions from Berkeley: we re-released the code under our license, keeping both that license and the original in the file.

Pfarrer:
All right, there are no Nodejs source included anymore. Instead, a build.xm Ant script is included that will download and extract those files.

msridhar:
@juliandolby do you have time to take a look at this? It will be a few days before I can look.

juliandolby:
downloading the nodejs code with Ant will definitely simplify issues with licenses and IP.  This is great.  @msridhar, I have looked at this code, indeed I talked about it with @Pfarrer while he was writing it.  @Pfarrer, did you add this new project to our maven scripts?  If not, I can do that.  Then I'll merge this.

Pfarrer:
No I did not. In fact, I did not touch any file outside of the two projects I created...

MaxSchlueter:
Great contribution! Just a remark on circular dependencies: if node.js encounters one it returns an unfinished copy of the required module back to the caller (https://nodejs.org/api/modules.html#modules_cycles). I don't think you are doing this right now.

Also dynamic arguments to require calls are a problem. I tested your code on this module:
```
var name = (Math.random() < 0.5) ? './a.js' : './b.js';
var mod = require(name);
mod();
```
(modules a.js and b.js simply return a function).
The CallGraph only contained module a.js but it should also contain a call to b.js.

I mean these things aren't happening too often in practice :)

PS: I had the same problem as you but instead of writing the code in WALA, I wrote a tool in node.js that does some source code transformation (https://github.com/MaxSchlueter/bundler). I also didn't write it to handle core modules in node.js but rather CommonJS modules according to the spec (http://wiki.commonjs.org/wiki/Modules/1.1). It can't really handle dynamic arguments to require calls or circular dependencies. The CallGraphs of your other test files look very similiar to mine :)

Pfarrer:
The arguments to require calls are resolved using WALA points-to information.
That is the function that is used to figure out what values the require argument might have:
https://github.com/wala/WALA/pull/106/files#diff-e85b78662dd882c8a50eaa47292959e6R140

Is there something wrong?

MaxSchlueter:
Yes, when you iterate over the possible call targets you only return the first one: https://github.com/wala/WALA/pull/106/files#diff-e85b78662dd882c8a50eaa47292959e6R122

Pfarrer:
Ohh, right... Is there a way to return multiple possible call targets in getCalleeTarget?

MaxSchlueter:
Hmm, I guess not. You'll have to ask the maintainers for that.

Pfarrer:
@juliandolby Any ideas for that?

msridhar:
Yeah, if we want to handle multiple possible targets for a `require` call, we'll have to do something a bit more complex.  The `MethodTargetSelector` is not intended to return multiple methods.  It's essentially designed to mimic whatever logic the language uses at runtime to figure out which method to invoke.  E.g., in Java, for a virtual call, at runtime a method is chosen based on the concrete type of the receiver; this logic is reflected in [ClassHierarchyMethodTargetSelector](https://github.com/wala/WALA/blob/30bc2858c4008b042fbc0d77209e114c7767c038/com.ibm.wala.core/src/com/ibm/wala/ipa/callgraph/impl/ClassHierarchyMethodTargetSelector.java).

The analogous role for the `MethodTargetSelector` here would be: once a single string argument is determined for a `require` call, the target selector would return the appropriate module.  But the target selector should not be determining the set of possible targets.

The "right" way to do this would be something like how we handle `Function.prototype.apply`:

* The [`JavaScriptFunctionApplyTargetSelector`](https://github.com/wala/WALA/blob/master/com.ibm.wala.cast.js/source/com/ibm/wala/cast/js/ipa/callgraph/JavaScriptFunctionApplyTargetSelector.java) returns a dummy target method for calls to `apply`.
* A corresponding [context selector](https://github.com/wala/WALA/blob/master/com.ibm.wala.cast.js/source/com/ibm/wala/cast/js/ipa/callgraph/JavaScriptFunctionApplyContextSelector.java) forces this dummy target method to be analyzed in contexts reflecting each possible function that might actually be invoked by `apply`.
* Finally, a [context interpreter](https://github.com/wala/WALA/blob/master/com.ibm.wala.cast.js/source/com/ibm/wala/cast/js/ipa/callgraph/JavaScriptFunctionApplyContextInterpreter.java) generates specialized IR for the dummy method given a particular context, that invokes the true target function with the given arguments.

This case is slightly different, since `require` doesn't actually get passed the function to be invoked; it gets passed a string.  If the relevant function is stored in a global variable somewhere (at least from WALA's point of view), the context interpreter could read this variable into an SSA local, and then do an invoke with that local as the function argument.  Maybe @juliandolby has alternate ideas here.  In any case, if this approach were taken, a special `NodeJsRequireTargetSelector` would no longer be required, since WALA would just do the right thing for the IR generated by the context interpreter.

Sorry for all this complexity, and for the slow response.  If we're ok with just getting one target for a `require` call, I am ok with merging the current approach, assuming the limitation is documented.

juliandolby:
I merged this, but have not done the merge of its maven files into our overall maven build.

PR #109, opened 2016-11-27
Ignore missing non-null-by-default annotations in Eclipse

liblit:
In general, the WALA code base is not really ready for nullness checking.  It would be nice if we got there some day, but I’m not planning to take that on now or any time soon.  Until then, it’s not useful to warn about missing `@NonNullByDefault` declarations on WALA packages.

See also older commit 7b6811b.


msridhar:
Thanks!

PR #111, opened 2016-11-27
Enforce restrictions on the model behavior class statically

liblit:
If a client violates these restrictions, I prefer that their code fail at compile time instead of run time.  Changing a few key types from `Class` to `Class<? extends AbstractAndroidModel>` gives us precisely the static enforcement we need and lets us remove an `AbstractAndroidModel.class.isAssignableFrom` run-time check.

However, this does change the public API of `AndroidEntryPointManager` in two ways.  The `getModelBehavior` and `setModelBehavior` methods now respectively accept and return `Class<?  extends AbstractAndroidModel>` instead of `Class`.  Is tightening up a public API in this manner considered OK?


msridhar:
I am fine with the tightening of the public API.  @juliandolby?

PR #112, opened 2016-11-28
Eliminate all Eclipse HTML validation warnings

liblit:
Most of these commits involve excluding specific HTML documents from validation. We no longer validate HTML that comes from third-party components, since we are not in a position to fix those anyway. Likewise, some invalid HTML generated by non-WALA tools is now excluded. I am also excluding some invalid HTML documents that are part of test suites. Some of these are clearly intended to be invalid. For others, it is not so clear. I have decided to assume that hand-edited HTML test inputs are intended to be exactly as they are, even when they are not valid HTML.

This pull request is not all about exclusions, though. We have some genuine fixes here too, all involving invalid HTML that arises from Javadoc comments. Most of this required properly escaping “<” and “>”. There were a few other cases of invalid HTML embedded in Javadoc comments, none common enough to
be worth detailing here.

None of these commits should be especially controversial. The closest any of these come to touching real Java code are the Javadoc comment fixes.


liblit:
[The Travis CI failure](https://travis-ci.org/wala/WALA/builds/179522959#L233) is an intermittent problem unrelated to these changes: 

> "/home/travis/build/wala/WALA/com.ibm.wala.core.testdata/build.xml:83: Can't get http://archive.apache.org/dist/jakarta/bcel/binaries/bcel-5.2.tar.gz to /home/travis/build/wala/WALA/com.ibm.wala.core.testdata/temp.folder/bcel-5.2.tar.gz".

I recommend telling Travis to restart the test. I had to restart the test twice earlier today before it finally succeeded.

msridhar:
Awesome.  Thanks so much for doing this!

liblit:
You're welcome!

PR #115, opened 2016-11-30
Use consistent Java versions, usually 1.7

liblit:
Previously, the various Eclipse projects’ Java configurations used mixtures of 1.6, 1.7, and 1.8.  Many were internally inconsistent, such as requiring 1.7 in `MANIFEST.MF` but 1.6 in the Eclipse JDT build preferences.  The Travis-CI configuration tests against both 1.7 and 1.8, but does not test against 1.6.

Across all projects, the most common version was 1.7.  So I’m going to assume that 1.7 is the intended build target.  This commit makes 1.7 the selected version nearly everywhere.

`com.ibm.wala.core.testdata` is the one exception.  This specific project uses a few features only found in 1.8, such as lambda expressions.  Previously, `com.ibm.wala.core.testdata` used 1.7 in some aspects of its configuration but 1.8 in others.  Now it consistently targets 1.8.  I wish this one project didn’t need to be inconsistent with the rest of WALA, but at least now it’s consistent with itself.

(Personally, I’d be happy to target 1.8 only.  But my impression across all of these configuration files is that the WALA developers still want to be compatible with 1.7.  If that is no longer a requirement, let me know and I will adjust these changes accordingly to target 1.8 only.)

This change eliminates 11 “`There is no 'jre.compilation.profile' build entry and the project has Java compliance preferences set`” warnings and 13 “`The JRE container on the classpath is not a perfect match to the 'JavaSE-1.7' execution environment`” warnings.  However, it also adds 450 “`Redundant specification of type arguments <...>`” warnings and 17 “`Resource '...' should be managed by try-with-resource`” warnings.  So this seems like a net step backward in my wish to reduce WALA warnings.  However, those new warnings concern Java 1.7 language features that we were not previously using to good effect in projects that targeted 1.6.  If we all agree that we can now target 1.7 instead, then we can use these helpful features as the newly-added warnings suggest.  So I call that a step in the right direction.


msridhar:
@juliandolby do we still need to stick with targeting Java 1.7, or can we target 1.8?

Issue #107, opened 2016-11-07
Changes in JarFileModule lead to ill behaviour


mohrm:
Commit 73747cbb86df0c2448a4fd264556752ef73b3558 changed the implementation of JarFileModule dramatically and introduced a subtle error.

* Before this commit, JarStreamModule took a JarInputStream and read the class files directly from there.
* Commit 73747cbb86df0c2448a4fd264556752ef73b3558 made JarStreamModule inherit from JarInputStream. The JarInputStream it takes is not stored as a field anymore, but passed to the super constructor. This is where the problem comes in. JarInputStream's constructor takes an arbitrary input stream and assumes that it gets the zipped data from that stream. Consequently, JarStreamModule also makes this assumption now. But the constructor still takes a JarStreamModule, hence the data will be unzipped twice which leads to the problem that WALA cannot read class files in certain situations anymore, e.g. from a jar within a jar, or from a jar by using JarStreamModule.

I suggest to [change the constructor ](https://github.com/joana-team/WALA/commit/971c8c188d763ab6ea59b10e5ba4ec4c5d9f63f8) of JarStreamModule to accept any InputStreams. If I see correctly, the only place of WALA which needs to be adapted is [a place in FileProvider](https://github.com/wala/WALA/blob/master/com.ibm.wala.core/src/com/ibm/wala/util/io/FileProvider.java#L178): You only need to get rid of the outer JarInputStream. I think, this code is never executed anyway, since if it were actually used to construct a Module, WALA would crash while constructing the ClassHierarchy because it cannot load any classes.

msridhar:
This is a bad bug; we really need to be able to read a jar from within a jar.  @juliandolby can we just revert the changes to JarFileModule from 73747cb?

msridhar:
I assume this was fixed by #120

Issue #88, opened 2015-08-15
Different call graphs for APK on subsequent runs


mherzberg:
I am struggling to build Java call graphs from APK files in a reproducable manner. I expect the call graph to be the same every time I build it, but the number of nodes differ up to a couple hundred between subsequent runs, although the code and circumstances do not seem to differ. Does anyone have an idea of what I am doing wrong?

I added some sample code below. It runs the ZeroCFABuilder 100 times on the "de.guh.finanzchecker" Android App I downloaded from the Google Play Store, although I encountered this problem with many other apps, too. Additionally I took the android.jar from API level 19. I then print the number of nodes and write nodes from the call graph into a file for debugging purposes. When I run this on my Arch Linux machine using "openjdk version 1.8.0_51" and the latest master branch of WALA, I got at least 10 different call graphs, ranging from 6984 to 7027 nodes. I appended the delta nodes from those two examples below the code.

``` Java
public static void main(String[] args)
        throws IOException, ClassHierarchyException, IllegalArgumentException, CallGraphBuilderCancelException {
    for (int i = 0; i < 100; i++) {
        AnalysisScope scope = DalvikTestBase.makeDalvikScope(null, new File("android19.jar"),
                new File("de.guh.finanzchecker.apk").getCanonicalPath());
        ClassHierarchy cha = ClassHierarchy.make(scope);
        AnalysisCache cache = new AnalysisCache(new DexIRFactory());
        Set set = new HashSet();
        set.add(LocatorFlags.INCLUDE_CALLBACKS);
        set.add(LocatorFlags.EP_HEURISTIC);
        set.add(LocatorFlags.CB_HEURISTIC);
        AndroidEntryPointLocator eps = new AndroidEntryPointLocator(set);
        AnalysisOptions options = new AnalysisOptions(scope, eps.getEntryPoints(cha));
        options.setReflectionOptions(ReflectionOptions.FULL);
        CallGraphBuilder cgb = Util.makeZeroCFABuilder(options, cache, cha, scope);
        CallGraph cg = cgb.makeCallGraph(options, null);


        File out = new File(cg.getNumberOfNodes() + "nodes.txt");
        if (out.exists())
            out.delete();
        System.out.println(cg.getNumberOfNodes());
        FileWriter fw = new FileWriter(out);
        List<String> nodes = new ArrayList<String>();
        for (Iterator<CGNode> it = cg.iterator(); it.hasNext();) {
            nodes.add(it.next().toString());
        }
[...]

PR #120, opened 2016-12-08
Pull request  misc bugfixes

mhecker:
This is a series of small fixes we found useful for JOANA.


mohrm:
Additional side-note: commit dff20ac49c57b59fe8f2343245e32feee63703da is a fix proposal for issue #107.

mohrm:
While preparing this pull request, we already did a travis run for it: 

https://travis-ci.org/joana-team/WALA/builds/182236756

Comparing the commit history, I come to the conclusion that the code under test was the same in both test runs. When I look at the log of the failing travis run, I see that there was only one failing test, where a call graph of an android app has not the expected shape. Is this maybe issue #88 striking again?

Issue #125, opened 2016-12-15
Fetching dexlib-1.3.4.jar does not work anymore


mohrm:
While setting up WALA, the build script of com.ibm.wala.dalvik tries to obtain the file

https://apktool2.googlecode.com/archive/c1659e62c414c96882c49afa72032b504c921b5a.zip

which is an old version of apktool2. It then unpacks that zip file and extracts dexlib-1.3.4.jar.

The link is dead now. I tried to build versions 1.4.0 and 1.4.3 of baksmali (see https://github.com/JesusFreke/smali -- smali contains baksmali and dexlib, more closely than apktool) but that fails with some NullPointerException while processing ANTLR grammars.

For our fork, I came up with the "solution" of just to fetch the library from SCanDroid (see https://github.com/joana-team/WALA/commit/f9f663821b7e9f8410788829065562451f09e1bc). But I think we should consider porting the code from dexlib to dexlib2 in the near future...
@juliandolby I think we already mailed about this at the end of june. Has there been progress since then at the IBM side? Maybe we can ask the smali developer for some advice regarding the transition dexlib -> dexlib2.

msridhar:
@mohrm I tried cherry-picking your commit to update the `build.xml` but got a whole bunch of compile errors.  I guess there are other changes in `com.ibm.wala.dalvik` that have not yet been merged that are required to use this other jar?

mohrm:
@msridhar In https://github.com/joana-team/WALA/commit/f9f663821b7e9f8410788829065562451f09e1bc I used the wrong location - the ant script then did not fetch the actual jar file, but a webpage referring to it.
I fixed that in https://github.com/joana-team/WALA/commit/fcb9c57e40cf576e9992d237058059fe9ea87ad0 but did not update this issue - sorry for that!

mohrm:
We leave this issue open until we have a proper solution.

msridhar:
FYI, in finding a way to get the Dalvik frontend onto Maven Central, I created a maven repo to host the current dexlib jar.  It's used in the [WALA-start build.gradle](https://github.com/wala/WALA-start/blob/master/build.gradle#L16).  But I don't want to upload that one-off jar to Maven Central.  Probably the right fix here is to port to dexlib 2.

Nugatp:
Hey Guys, I am sure this might not be the best place to post, but I got a questions.
[...]

PR #130, opened 2017-01-03
Dexlib fix

mohrm:
fix for #125 - at least for now...


msridhar:
Awesome, thanks!  Let's leave #125 open for getting a proper solution

Issue #133, opened 2017-01-26
using HashCodeComparator in class loading may and does cause instabilities


mohrm:
In `ClassLoaderImpl`, the method `getClassFiles` is responsible for retrieving all the module entries from a module and returning them as a set. For this, first the iterator is converted into a list and then this list is inserted into a TreeSet, using a comparator which compares two module entries by hash code (!).
I think this is highly fragile. The [Java API doc](https://docs.oracle.com/javase/8/docs/api/java/util/TreeSet.html) clearly says that `TreeSet` relies on the used `compareTo` method [being consistent with](https://docs.oracle.com/javase/8/docs/api/java/lang/Comparable.html) `equals`. In particular, `TreeSet` uses `compareTo == 0` as eqality check. The used `HashCodeComparator` is *not* consistent with `equals` as there may be collisions. So, using a `HashCodeComparator` in `getClassFiles` may accidentally delete module entries.

I also think that there is a connection of this issue to #88. The dalvik front-end class `DexClassLoaderImpl` contains a method which is similar to `getClassFiles` (`getDexFiles`). The bad thing is that the `DexModuleEntry` class it processes does not implement `hashCode` or `equals`. I think this causes the flakiness, since now `System.identityHashCode` is used which is completely unpredictable.
One simple fix would be to implement `hashCode` and `equals` consistently for `DexModuleEntry`.

Another possible solution would be to not use `TreeSet` and `HashCodeComparator` but a `LinkedHashSet`.

I am in favor of the latter solution because it is more robust. I also don't think that it is really necessary to first unroll an iterator into a list and then insert all list items into a set, so removing this code would not only make it more robust but also faster and simpler.

@juliandolby @msridhar What do you think?

juliandolby:
Martin,

 This seems like an important issue; thanks for finding it.  I think making sure equals, hahsCode and compareTo are all consistent is good to do anyway, so that seems like the right fix to me.

  — Julian

mohrm:
@juliandolby What do you think about getting rid of HashCodeComparator, as proposed in #134? This not only seems more robust to me but also simpler and faster.

juliandolby:
That seems like a good idea too.

msridhar:
Fixed by #134

PR #134, opened 2017-01-26
get rid of HashCodeComparator and all its usages

mohrm:
This addresses both https://github.com/wala/WALA/issues/133 and https://github.com/wala/WALA/issues/88


msridhar:
So the `HashCodeComparator` cods is really old and I can't tell from version control history why it was introduced.  I'm guessing that at some point, module entries were being returned in a non-deterministic order, and this was an attempt to iterate through them deterministically.  But I agree that it's not the right way to do it, at least not anymore.  Seems like Julian is also OK with this change, so merging.  Thanks for finding this!

PR #136, opened 2017-02-11
IFDS Null Analysis files

d2an3:
This pull request contains an IFDS-based analysis for detecting null-pointer dereferences. The analysis tracks access paths that either may be null or may-not be null.


juliandolby:
It would be great if there was a test of this analysis, so we can make sure it keeps working.

juliandolby:
I merged this before seeing Manu's comment.  I then reverted the merge in my own clone before pushing, so the code is not there even though it looks closed.  I guess we may need a new pull request once the approvals are in place.

msridhar:
Yeah sorry for confusion. We'll open a fresh request once things are in order.

PR #140, opened 2017-03-12
Remove execute bits from many non-executable (i.e., non-script) files

liblit:


msridhar:
Can I ask how you ended up even noticing this? 😄 Also why is it important?  No objections to the PR, just curious.

liblit:
I primarily work under Linux, where such bits matter. My “`ls`” command is configured to color executable files green and add a “`*`” after their names. These bits also have an effect on tab completion in shells. So things that are marked as executable catch my attention. If they are marked as executable but are not actually executable, then that is an unnecessary distraction, suggesting an affordance that does not exist. I noticed one such file, assumed there would be others, and therefore systematically fixed them all.

msridhar:
Sounds good to me.

msridhar:
Another thought: if we see this problem cropping up again, and you know something about git commit hooks, feel free to contribute one of those that prevents `.java` and `.js` files from being committed with execute bits set (or maybe even just unsets them).

liblit:
Commit hooks sound perfect for this. I know nothing about them in `git` specifically, but I’m sure I can figure them out if the problem continues to reappear (and continues to bug me). Good idea.

PR #145, opened 2017-03-15
Additional refactorings for IAnalysisCacheView.

khatchad:
Other places where IAnalysisCacheView can be used. May not be inclusive.


msridhar:
@juliandolby can you look at this one?  I'm not super familiar with the new `IAnalysisCacheView` type and when it should be used.

khatchad:
Right. Now, some methods return IAnalysisCacheView instead of AnalysisCache. And, there are methods for which I was sending the result that now should also change their signatures.

msridhar:
Ok, sounds good.

PR #148, opened 2017-03-17
Automatically download “dx.jar” when building

liblit:
This should make it easier for newcomers to get WALA up and running from sources. One no longer needs to manually find `dx.jar` using instructions given on WALA’s [wiki](http://wala.sourceforge.net/wiki/index.php/UserGuide:Getting_Started#Building_the_code) or [mailing list](https://groups.google.com/forum/#!msg/wala-sourceforge-net/cBYsfEvYVG0/Ua52dyQQU-YJ). Those instructions were already out-of-date, anyway.

If this pull request is accepted, someone should update the WALA wiki accordingly. In particular, I recommend changing [this text](http://wala.sourceforge.net/wiki/index.php/UserGuide:Getting_Started#Building_the_code) from this

> First, manually download dx.jar and place it in the com.ibm.wala.dalvik.test/lib folder, as described [here](https://groups.google.com/d/msg/wala-sourceforge-net/cBYsfEvYVG0/Ua52dyQQU-YJ). (Unfortunately, we cannot easily download this jar automatically.) Then, in your root directory…

to this

> In your root directory…

I would make this change myself, but I do not have a WALA wiki account and cannot figure out how to create one.

Given that an official release of `dx.jar` is available in Maven, it would be nice to have Maven itself manage that dependency. Unfortunately, it seems that WALA in general does not use Maven for dependency management. As far as I can tell, the Maven configuration just calls out to Ant for doing builds, including having Ant download other needed components. If WALA ever moved to using Maven more fully, downloading `dx.jar` could easily come along with that.


msridhar:
This is awesome!  Thanks for figuring this one out.

msridhar:
And we should be able to have Maven manage this dependence, at least redundantly with Ant.  I'll look into it.

liblit:
You're welcome. I'm very glad to finally get this itch scratched once and for all, for me and for everyone else. Please remember to update the "Getting Started" wiki instructions as well, as suggested in my [original description of this pull request](https://github.com/wala/WALA/pull/148#issue-215077892)

I definitely don't know much about Maven, Ant, and how they both interact with Eclipse. If you can figure out a way to have Maven manage this, more power to you!

msridhar:
I fixed the "Getting Started" page; thanks!  (We really need to port all our docs to the wiki on GitHub...)

I reconsidered on getting Maven to manage the dependence.  At some point, I'd like to move away completely from being so Eclipse-centric, but for now managing is Ant is better from an Eclipse compatibility standpoint.

Issue #151, opened 2017-03-22
"type mismatch" in com.ibm.wala.escape.FILiveObjectAnalysis


mamarsh:
We've been using a fork of WALA, and with the new release I was hoping to move us back to the mainline.  I'm getting an error when compiling, though:

```
[error] /a/file/path/MyFile.scala:27: type mismatch;
[error]  found   : com.ibm.wala.analysis.pointers.HeapGraph[_$1] where type _$1 <: com.ibm.wala.ipa.callgraph.propagation.InstanceKey
[error]  required: com.ibm.wala.analysis.pointers.HeapGraph[com.ibm.wala.ipa.callgraph.propagation.InstanceKey]
[error] Note: _$1 <: com.ibm.wala.ipa.callgraph.propagation.InstanceKey, but Java-defined trait HeapGraph is invariant in type T.
[error] You may wish to investigate a wildcard type such as `_ <: com.ibm.wala.ipa.callgraph.propagation.InstanceKey`. (SLS 3.2.10)
[error]     val liveAnalysis = new FILiveObjectAnalysis(GlobalHeap.cg, GlobalHeap.heap, false)
```

It might be worth checking all of the changes from commit fde65340d22f153914343a33bbaf6f5497ce67f4 (where this was introduced) for similar issues.

msridhar:
@liblit can you take a look?

msridhar:
@mamarsh do you have a pure Java example to reproduce?

mamarsh:
@msridhar I don't, unfortunately.  I can try to put one together, but I'm not sure how long that'll take me.

msridhar:
@mamarsh ok, if you have time to actually explain the problem here, that'd be helpful 😄 I'm not an expert in Scala compile warnings.

liblit:
@mamarsh, thank you for tracking this down to a specific commit, which turns out to have come from me. I am sorry for the disruption those changes are causing you.

Like @msridhar, I am no Scala expert. Doing my best to decode the error message and infer what might be happening, here is what I see. Scala is finding a value (in your code) whose static type is what Java would call `HeapGraph<? super InstanceKey>`. However, what it was actually expecting was `HeapGraph<InstanceKey>`. So that's a static type mismatch. Apparently the code that leads to this error is the following Scala fragment:

``` scala
val liveAnalysis = new FILiveObjectAnalysis(GlobalHeap.cg, GlobalHeap.heap, false)
```

[...]

PR #153, opened 2017-03-22
Generalize a generic type parameter even more

liblit:
As it turns out, I should have been using "`? extends InstanceKey`" rather than "`? super InstanceKey`".  But really, we can just use "`?`" here since `HeapGraph` itself constrains its own type parameter appropriately.

This should fix WALA bug #151, which was first introduced by [this change](https://github.com/wala/WALA/commit/fde65340d22f153914343a33bbaf6f5497ce67f4#diff-6f505606d6841bf44858283ce76371ceL71) in commit fde65340d22f153914343a33bbaf6f5497ce67f4.

### DIFF
--- a/com.ibm.wala.core/src/com/ibm/wala/escape/FILiveObjectAnalysis.java
+++ b/com.ibm.wala.core/src/com/ibm/wala/escape/FILiveObjectAnalysis.java
@@ -48,7 +48,7 @@ public class FILiveObjectAnalysis implements ILiveObjectAnalysis {
   /**
    * Graph view of pointer analysis results
    */
-  private final HeapGraph<? super InstanceKey> heapGraph;
+  private final HeapGraph<?> heapGraph;
 
   /**
    * Cached map from InstanceKey -> Set<CGNode>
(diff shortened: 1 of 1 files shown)


liblit:
The Travis CI failure seems to be an unrelated network outage:

```
ERROR] /home/travis/build/wala/WALA/com.ibm.wala.core.testdata/build.xml:84: Can't get http://archive.apache.org/dist/jakarta/bcel/binaries/bcel-5.2.tar.gz to /home/travis/build/wala/WALA/com.ibm.wala.core.testdata/temp.folder/bcel-5.2.tar.gz
```

I recommend waiting a bit, then asking Travis CI to restart the build.

msridhar:
Tried restarting the Travis build (weird, I can download that URL locally).  

@liblit just to double-check, there will be a compile-time error if someone tries to instantiate `HeapGraph` with a type parameter that is not a subtype of `InstanceKey`, independent of this `?`, right?  Java generics are confusing 😕

liblit:
@msridhar: correct, that will result in a compile-time error. For example, I just tried the following small code fragment:

```java
new FILiveObjectAnalysis(null, new HeapGraph<String>(), true);
```

You might think this would work, since the second argument type is `HeapGraph<?>`. However, the constraints on the `HeapGraph` class itself make `HeapGraph<String>` a statically invalid type. Eclipse reports the following errors:

>Bound mismatch: The type `String` is not a valid substitute for the bounded parameter `<T extends InstanceKey>` of the type `HeapGraph<T>`
>Cannot instantiate the type `HeapGraph<String>`

I expect that any other Java compiler would (indeed, *must*) reject this in a similar manner.

msridhar:
Travis is being a pain this afternoon :-(  To play it safe, I'm not going to merge until I can get the Travis build to go through.

liblit:
I agree completely.

PR #155, opened 2017-03-23
Fix 161 Eclipse Javadoc warnings

liblit:
There are a small number of suppression annotations in here for things there is just not a good fix for. But the vast majority of these changes are real fixes to real Javadoc glitches.

This set of changes, combined with those in #154, eliminate all Eclipse Javadoc warnings. Let's keep it that way, shall we?


liblit:
The Travis CI failure seems to be an unrelated network outage:

> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:1.7:run (default) on project com.ibm.wala.core.testdata: An Ant BuildException has occured: The following error occurred while executing this line:
> [ERROR] /home/travis/build/wala/WALA/com.ibm.wala.core.testdata/build.xml:37: java.net.ConnectException: Connection timed out
> [ERROR] around Ant part ...<ant antfile="/home/travis/build/wala/WALA/com.ibm.wala.core.testdata/build.xml" target="build.update.jar"/>... @ 5:111 in /home/travis/build/wala/WALA/com.ibm.wala.core.testdata/target/antrun/build-main.xml

I recommend waiting a bit, then asking Travis CI to restart the build.

msridhar:
I'll be honest that these days I don't use Javadoc that much for open-source projects; I just browse the docs+source in my IDE.  If we were serious about not regressing on the Javadoc warnings, we'll have to introduce some tooling around it that is independent of Eclipse, probably.  It's not a huge priority for me, so someone else would have to work on that.  I am not opposed to keeping the Javadoc comments clean, though, if it's not too painful.

liblit:
Understood. Personally I'm in favor of turning off warnings one doesn't realistically plan to fix, but also being very aggressive about fixing anything not turned off. If the right thing to do is to change Eclipse configurations to warn about fewer things, that's fine with me. But certainly any Javadoc that *is* present should at least be valid. That's what this pull request is about: not documenting everything exhaustively, but making sure the documentation that exists works as its authors intended.

PR #156, opened 2017-03-23
Fix 265 Eclipse warnings about unnecessary code

liblit:
The changes proposed here fix about 265 Eclipse warnings about unnecessary code: unused imports, useless casts, etc. I consider these changes to be uncontroversial: each seems like an obvious fix without larger implications. I welcome skeptical outside review, of course! But these should all be pretty innocuous.

That being said, I do have a second batch of "unnecessary code" fixes pending that will be more controversial. So let's take care of these easy ones before we get into trickier material that may or may not be wanted.


com.ibm.wala.cast.js/source/com/ibm/wala/cast/js/loader/JavaScriptLoader.java
```diff
@@ -263,7 +262,7 @@ public SSAPutInstruction PutInstruction(int iindex, int ref, int value, String f
             byte[] utf8 = field.getBytes("UTF-8");
             return PutInstruction(iindex, ref, value, 
                 FieldReference.findOrCreate(JavaScriptTypes.Root, Atom.findOrCreate(utf8, 0, utf8.length), JavaScriptTypes.Root));
-          } catch (UnsupportedEncodingException e) {
+          } catch (@SuppressWarnings("unused") UnsupportedEncodingException e) {
```
msridhar: These are weird.  I think there are plenty of legitimate cases where you catch an exception and the handling does not make use of the exception object.  Do we really need to suppress these?  Can we shut off the Eclipse warning instead?

liblit: Yes, we can definitely turn off the Eclipse warning for unused exception parameters, and get rid of those per-location suppression declarations. Given that there are 157 such cases, that seems perfectly reasonable. I'll make that change and merge it into this existing pull request.

PR #163, opened 2017-03-29
Disable Eclipse warnings about missing version constraints

liblit:
Specifically, we're turning off Eclipse warnings about missing version constraints on required bundles (`Require-Bundle`), exported packages (`Export-Package`), and imported packages (`Import-Package`). We're not turning these off absolutely everywhere, though: only in packages where one or more such warnings were actually being reported. So if a given package was already providing all version constraints for, say, package imports, then we've kept that warning on in that package.

Honestly I don't entirely understand the practical implications of these warnings.  However, there were **355** of them across many WALA subprojects.  I take this as evidence that the WALA developers do not consider these version constraints to be important.  Therefore, we may as well stop warning about something we have no intention of fixing.

That being said, if we *do* want to fix some or all of these, I welcome any advice on what those fixes should look like.  I am rather ignorant about all things OSGi.


msridhar:
You're right, we don't care 😄

PR #173, opened 2017-04-28
Optimize PDG to only check callers for number of callee params when n…

msridhar:
…ecessary

This extra step should only be needed for languages that let you pass extra
parameters at call sites, like JavaScript.


msridhar:
@juliandolby I'm a bit concerned we don't have a unit test to cover this code in PDG that looks through all the call sites to figure out the number of callee parameters.  Do you have anything handy you could add to this branch?

msridhar:
Fixed by #191 instead

PR #146, opened 2017-03-15
Fix Eclipse warnings about type safety and raw types

liblit:
Previously, Eclipse reported 140 warnings in the “Type Safety and Raw Types” category. The commits in this pull request bring that count to zero.

PR #174, opened 2017-05-03
Remove unnecessary `klass` argument and propagate back to callers

liblit:
Special thanks to @khatchad for [asking why this argument was needed](https://github.com/wala/WALA/pull/146#commitcomment-21999159). It isn't!


msridhar:
👍

PR #178, opened 2017-05-07
Add @Override annotations wherever possible

liblit:
This resolves nearly 300 Eclipse code style warnings about methods that override but are not declared (annotated) as doing so.


msridhar:
Just curious, how did you insert these annotations? I hope not by hand :-) Does not Eclipse let you autofix in bulk?

liblit:
This specific batch of `@Override` annotations were all fixable in a single bulk operation. For some of the other warning fixes I have pushed, that has not been available.

In theory, Eclipse does let you apply the same “Quick Fix” to multiple instances of the same problem. In practice, Eclipse often treats problems as being distinct instances when a human would consider them the same. So I do sometimes end up with dozens or hundreds of near-identical fixes that need to be applied one at a time. This is a fine activity to occupy one while flying across an ocean, though.

Even for the bulk `@Override` insertions, Eclipse sometimes re-indented near the insertion in a way that introduced gratuitous white-space changes. When this happens in a given file, though, it is generally quite systematic and therefore easily fixed with a throw-away Emacs macro or two.

msridhar:
Looks like this one now somehow has a merge conflict @liblit

liblit:
Conflict resolved.

msridhar:
Thanks!  Eventually it'd be nice to have a way to enforce this bug doesn't get re-introduced.  We use [Error Prone](http://errorprone.info/bugpattern/MissingOverride) at Uber, maybe we would eventually get to the point where we run it in CI for WALA builds.

liblit:
I agree. That [Error Prone MissingOverride check](http://errorprone.info/bugpattern/MissingOverride) is perfect for this specific issue, if we can get it hooked up in CI.

In general, though, I am fixing many warnings but we have nothing in place to prevent new instances. If we wanted to get strict about that, I guess we’d need a programmatic way for CI to (1) run Eclipse’s code checking logic over WALA, (2) count how many warnings arise in each category, and (3) fail the CI test if the warning count in any category increases. I know nothing at all about using Eclipse in a headless, automated way like that.

PR #179, opened 2017-05-11
Fix or suppress 373 Eclipse Code Style warnings relating to static declarations

liblit:
Broadly speaking, these changes are of three forms:

1. Declare methods `static` where this is both possible and safe. Here “safe” means that the method cannot possibly be overridden by third-party code outside of the WALA tree. `private` methods are safe in this manner, as are `final` methods and methods of `final` classes.

1. Suppress Eclipse suggestions about making methods `static` where Eclipse is simply wrong. Some methods that seem suitable for `static` in isolation need to be non-`static` because they have overrides elsewhere and rely on dynamic dispatch. Also, JUnit `@Test` methods cannot be `static`.

1. Access `static` fields directly through the class that declares them, not through a subclass or an instance.


liblit:
Closing for now. An improved version will come along later, but there a few other warning categories I prefer to target first.

PR #181, opened 2017-05-15
Simplify management of virtual frame buffer X server

liblit:
Using the `xvfb-run` script gives us automatic lifetime management for the virtual X server: it will start before the `mvn` test script and will terminate when that test script exits.  The `xvfb-run` script also sets `$DISPLAY` properly in the `mvn` test script, so we no longer need to manage that setting ourselves.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,11 @@
 language: android
 jdk: oraclejdk8
 before_install:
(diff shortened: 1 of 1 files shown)


msridhar:
Thanks!

liblit:
You’re welcome, @msridhar!

PR #182, opened 2017-05-15
Fix all Eclipse warnings about unused locals and parameters

liblit:
Taken together, these commits fix or suppress 201 Eclipse warnings relating to unused local variables and method parameters. A few of these commits warrant closer attention:

- d9d471df6b7d17e36cc69f0d0b6a79e1e3fd3b4b comments out several large chunks of code that were only manipulating local variables that were otherwise unused. A typical example would be code that loops over a list of warnings, but does not actually print or log the warnings. I decided to be cautious here and comment out the code rather than removing it entirely. Perhaps someone will decide to do something more significant with this code in the future, such as actually logging those warnings. It is also possible that I accidentally commented out something with important side effects, though I made a conscious effort to watch for and avoid that.

- b4eaa8dc36f2e510904f2731068cc3b3954cd221 and a9e4c2e3f7bd0fc8d9eb9afaa1ac799caf1139ab respectively remove many calls to `AnalysisCache.getIR()` and `ClassHierarchy.resolveMethod()` whose result is never subsequently used. My impression is that neither of these calls has side effects we need here. If I am mistaken, then it should be easy enough to back all of these changes out _en masse_.

- d43cecff3ca0f3a14d8947f5ca601c79ebe6fabb removes unused method arguments; some of these changes could affect public API visible to third-party WALA code.

### DIFF
--- a/com.ibm.wala.cast.java.ecj/src/com/ibm/wala/cast/java/translator/jdt/JDTJava2CAstTranslator.java
+++ b/com.ibm.wala.cast.java.ecj/src/com/ibm/wala/cast/java/translator/jdt/JDTJava2CAstTranslator.java
@@ -509,12 +509,12 @@ public abstract class JDTJava2CAstTranslator<T extends Position> {
         else if (met.getParameterTypes().length > 0)
           memberEntities.add(createDefaultConstructorWithParameters(met, n, context, inits));
         else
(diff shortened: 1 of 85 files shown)


msridhar:
This all LGTM.  @juliandolby any objections re: the public API changes?  If not I will merge.

msridhar:
Thanks @liblit!

liblit:
Glad to help.

PR #184, opened 2017-05-27
Fix or suppress all Eclipse code style warnings relating to static methods and fields

liblit:
Taken together, these commits fix or suppress 238 Eclipse code style warnings relating to `static` methods and accesses to `static` fields.

Many of these fixes consist of declaring methods `static` wherever it is safe to do so. I’ve made that change only in cases where the affected method cannot possibly be overridden by third-party code: `private` methods (40359dfb2baadc68de999b0c3eb110ea2ee64e63), `final` methods (9e9a861d8cd32d0435f3c566119f27875c7479b6), and methods of `final` classes (79209e96fe81ac82773bc5564a78625d7f014a04). An additional group of fixes involve accessing static fields directly through the classes that declare them, rather than through some subclass or instance (91d85ab9ef418559bd850a69645042a41ac304a4).

In other cases, we suppress Eclipse’s recommendations because we know things Eclipse does not. For example, methods that third-party code can see and potentially override should not be changed, since the third-party code may rely on dynamic dispatch of calls to these methods (423d80835659430aebac9df02732d6bef66d5a71). Also, JUnit `@Test` methods must not be static, but Eclipse doesn’t know about that rule (b458dd568bbc91b37d471d278303b2e8034f08df). Sometimes test inputs intentionally do seemingly-goofy things that we would want to fix in regular code, but which we should silently allow in test code because that’s the point of the test (6bea42fab206a6838889b4daa774ef8f87e484f2).

### DIFF
--- a/com.ibm.wala.cast.java.ecj/src/com/ibm/wala/cast/java/translator/jdt/JDTJava2CAstTranslator.java
+++ b/com.ibm.wala.cast.java.ecj/src/com/ibm/wala/cast/java/translator/jdt/JDTJava2CAstTranslator.java
@@ -396,7 +396,7 @@ public abstract class JDTJava2CAstTranslator<T extends Position> {
 
   }
 
-  private boolean isInterface(AbstractTypeDeclaration decl) {
+  private static boolean isInterface(AbstractTypeDeclaration decl) {
     return decl instanceof AnnotationTypeDeclaration ||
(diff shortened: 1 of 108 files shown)


msridhar:
Thanks @liblit!

liblit:
My pleasure, @msridhar!

PR #186, opened 2017-06-07
When converting JDT AST to IR, a do-while loop in a case statement will throw NPE

caiusb:
This fixes this issue.


msridhar:
Can you add a unit test?

caiusb:
Added. I'm not 100% it's in the right place. If you want me to move it somewhere else, please let me know.

msridhar:
Awesome, thanks for the fix!!

PR #196, opened 2017-06-30
Fix Eclipse warnings about missing hashCode() methods

liblit:
Add a `hashCode()` method to a class that already has `equals()`. Generally, overriding one means you should be overriding the other too. Also, configure Eclipse to treat any similar cases as errors, rather than merely warnings.

Ignore Eclipse warnings about missing `hashCode()` methods in tests. In regular application code, these warnings should be taken seriously and fixed.  But for test code, it's better to keep things simple and not add any methods that aren’t strictly needed by the test.


msridhar:
I noticed you only made this an error in two (non-test) projects.  What about the other ones?

liblit:
I believe that every other non-test project *already* treats missing `hashCode()` methods as errors. Except for the intentional exclusions of `com.ibm.wala.cast.java.test.data` and `com.ibm.wala.core.tests`, every `.settings/org.eclipse.jdt.core.prefs` file already contains `org.eclipse.jdt.core.compiler.problem.missingHashCodeMethod=error`.

@msridhar, do you see some that I missed?

msridhar:
Nah, just didn't realize that it was already another error for all the other projects.

PR #103, opened 2016-06-27
Build fixes for Eclipse Neon

liblit:
The fixes needed for Eclipse Neon are very similar to those previously added for Luna (c9ad359d656463eccea08417d9855d4eb8f4478b + 0e55cd2eecdd830428a4fa16ab99ef7ca98f9df4, fixing #56) and Mars (6df8d1b32d1b6a9750bdb368efaecab8c1edca8c, fixing #98).

Issue #98, opened 2016-03-28
Eclipse Mars build failure for FakeExceptionTypeBinding


liblit:
Class `com.ibm.wala.cast.java.translator.jdt.FakeExceptionTypeBinding` fails to build under Eclipse Mars (4.5.2). One error is reported for line 58 of `com.ibm.wala.ide.jdt/source/com/ibm/wala/cast/java/translator/jdt/FakeExceptionTypeBinding.java`: `The type FakeExceptionTypeBinding must implement the inherited abstract method ITypeBinding.getDeclaringMember()`.

`ITypeBinding` here refers to `org.eclipse.jdt.core.dom/ITypeBinding`, which presumably added this `getDeclaringMember()` method in recent Eclipse releases.

This is similar to issue #56, although in that case it was Eclipse Luna that had problems.

msridhar:
Thanks for the report!

Issue #56, opened 2014-06-22
Update to compile with Eclipse Luna


msridhar:
I'm running Eclipse Kepler SR2 with the patched JDT to handle Java 8, and I get the following compile errors in com.ibm.wala.ide.jdt:

```
Description	Resource	Path	Location	Type
The type FakeExceptionTypeBinding must implement the inherited abstract method ITypeBinding.getTypeAnnotations()	FakeExceptionTypeBinding.java	/com.ibm.wala.ide.jdt/source/com/ibm/wala/cast/java/translator/jdt	line 58	Java Problem
The type FakeExceptionTypeBinding must implement the inherited abstract method ITypeBinding.getFunctionalInterfaceMethod()	FakeExceptionTypeBinding.java	/com.ibm.wala.ide.jdt/source/com/ibm/wala/cast/java/translator/jdt	line 58	Java Problem
```
I'm guessing these errors will persist under Eclipse Luna.  Nothing urgent, just noting for future reference.

msridhar:
This is fixed now

denis-bogdanas:
I still get a similar error with wala 1.3.8:

The type FakeExceptionTypeBinding must implement the inherited abstract method ITypeBinding.getDeclaringMember()

liblit:
@denis-bogdanas, I am seeing the same problem with `ITypeBinding.getDeclaringMember()` under Eclipse Mars. I have reported this as a separate WALA issue: #98.

denis-bogdanas:
As workaround you can add an empty implementation of that method. The other methods in that class don't do much either.

PR #197, opened 2017-07-07
Enable Eclipse Oxygen’s new “unlikely argument types” diagnostics

liblit:
Eclipse Oxygen offers [two new diagnostics](https://www.eclipse.org/eclipse/news/4.7/jdt.php#unlikely-argument-types) for situations where argument types are strictly correct, but unlikely to be what the programmer intended. One of these diagnostics covers various collection methods, and the other covers `Object.equals()`.  For each of these diagnostics, I’ve set the severity level to “warning” in projects that have some instances of the suspicious code, or to “error” in projects that have no instances of the suspicious code.

I am pleased to report that no other changes were needed to get WALA up and running under Eclipse Oxygen. That’s a nice change from the porting work we had to do for [Neon](https://github.com/wala/WALA/pull/103), [Mars](https://github.com/wala/WALA/issues/98), and [Luna](https://github.com/wala/WALA/issues/56).


msridhar:
@liblit have you had a chance to inspect the warnings at all? Also, I am unclear about what ECJ compiler version is used by Tycho. Could you check if introducing an error here actually causes a Maven / Travis build failure?

liblit:
@msridhar, yes I have looked at the warnings. There are only four, one of which appears twice. In my judgement, all four point out code that likely does not do what its authors intended:

Occurrences|Problem Description|My Opinion
-|-|-
2|Unlikely argument type `CAstNode` for `containsKey(Object)` on a `Map<Pair<CAstNode,CAstBasicRewriter.NoKey>,CAstNode>`|The warning seems valid; the code seems wrong. The given map will never contain any `CAstNode` as a key, because the map's keys are always `Pair<...>`. The code in question is asserting non-containment, so that `assert` will always succeed, but surely is not checking what the author intended to check.
1|Unlikely argument type `Integer` for `remove(Object)` on a `Collection<HyperNode<Integer>>`|The warning seems valid; the code seems wrong. The removal will always fail, because we are trying to remove an `Integer` from a collection of `HyperNode<Integer>`. I am not sure what the author intended, but this doesn't seem right.
1|Unlikely argument type `TypeName` for `get(Object)` on a `Map<String,String>`|The warning seems valid; the code seems wrong. The `get` will always fail, because `TypeName` is not a `String` and the map only contains items keyed by `String`. As in the first row of this table, the code in question is asserting that the `get` fails, which it always will. But surely this is not what the author intended to check.

Regarding which ECJ compiler version is used by Tycho, I am also confused about this. I raised the severity of the above from warnings to errors, then ran `mvn clean install -DskipTests`. To my surprise, this succeeded, even though the Eclipse GUI was definitely showing four errors. So I do not completely understand how Tycho, ECJ, and the Eclipse GUI interact, and in what ways their diagnostics configurations are shared or not.

msridhar:
Re: Tycho and ECJ, I think we need to set the [target platform](https://wiki.eclipse.org/Tycho/Target_Platform); not sure what it's set to now.  I'm a bit hesitant to bump the target platform to Eclipse Oxygen so soon after release.  @juliandolby are there any IBM constraints around Eclipse target platform version?

liblit:
I am not sure whether changing the target platform would cause Tycho to report errors where Eclipse is configured to do so. I would try this out to see what happens if I could figure out how, but I cannot make sense of that documentation.

Either way, I think there is no harm in enabling these errors while not changing the target platform to Oxygen. Older Eclipse releases should just quietly ignore the settings for diagnostics they do not understand, while those of us using Oxygen will benefit from the additional checks.

msridhar:
OK i'll merge. @liblit can you open an issue on updating the target platform in Tycho, so we don't forget?

msridhar:
Also @liblit if there are obvious fixes for those warnings then fixes would be welcome ☺️

liblit:
@msridhar, I prefer that you add the issue for target platform updating. I don't even understand what that means.

PR #200, opened 2017-07-11
Fix nearly all Eclipse warnings about using raw types

liblit:
Along the way, I also converted many `for (;;)` loops into modern `for (:)` loops. I didn’t systematically look for all opportunities to do this, though. I merely made this change where I was already converting raw `Iterator` uses into modern `Iterator<...>` uses.

Better use of generics also allowed many casts to become statically redundant. I have removed all such redundant casts.

Only three raw-types warnings remain after this batch of fixes. All three involve raw uses of `CallGraphBuilder`. I tried to fix these too, but it quickly snowballs into a cascade of changes that may or may not eventually reach a statically-type-safe fixed point. I may give these last few problem areas another go in the future. For now, though, the hundreds of other fixes seem worth keeping even if there are a few stragglers.

This commit may change some public APIs, but only by making weaker type signatures stronger by replacing raw types with generic types. For example, we may change something like `Set` into `Set<String>`, but we’re not adding new arguments, changing any underlying (post-generics-erasure) types, etc.

<hr/>

This commit also upgrades raw-types-usage warnings to errors where possible. In general, these diagnostics are now errors in projects for which all such warnings have been fixed. There are three unfixed warnings in two projects, so this diagnostic remains a warning (not an error) in those projects.

There are also many places where raw-types-usage warnings have been locally suppressed using `@SuppressWarnings` annotations. I haven’t systematically revisited those to see if any can be fixed properly. So for those projects this diagnostic must also remain a warning (not an error), since `@SuppressWarnings` does not work on things Eclipse is configured to treat as errors.


msridhar:
Wow!  This sounds great.  Quick question.  For the loop conversion, did you only convert loops where the original loop was using an `Iterator`?  I ask because at least for arrays and lists, if the original loop was of the form `for (int i = 0; i < len; i++)`, we might want to preserve that form if the loop is performance-critical (I think it's still faster than allocating an `Iterator`).

liblit:
Yes, I only converted loops to `for (:)` if they were already using `Iterator`. I did not systematically look for other opportunities to convert loops; I only made this change where I was changing a pre-existing raw `Iterator` type to an `Iterator<...>` type with proper generic type parameters.

com.ibm.wala.core.testdata/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -81,7 +81,7 @@ org.eclipse.jdt.core.compiler.problem.potentiallyUnclosedCloseable=error
 org.eclipse.jdt.core.compiler.problem.rawTypeReference=warning
 org.eclipse.jdt.core.compiler.problem.redundantNullAnnotation=error
 org.eclipse.jdt.core.compiler.problem.redundantNullCheck=error
-org.eclipse.jdt.core.compiler.problem.redundantSpecificationOfTypeArguments=warning
+org.eclipse.jdt.core.compiler.problem.redundantSpecificationOfTypeArguments=ignore
```
msridhar: The actual build of the testdata jar file is done with ant, via the `build.xml` file at the root directory.  But since this is testdata input anyway, I don't think we care about fixing these type of warnings.

com.ibm.wala.cast/source/c/Makefile.configuration
```diff
@@ -1,5 +1,5 @@
-JAVA_SDK = /usr/lib/jvm/jdk1.8.0_111/
-DOMO_AST_BIN = /home/vagrant/WALA/com.ibm.wala.cast/target/classes/
-JAVAH_CLASS_PATH = :/home/vagrant/WALA/com.ibm.wala.cast/target/classes/
+JAVA_SDK = /usr/lib/jvm/java-1.8.0-openjdk-1.8.0.131-1.b12.fc25.x86_64/
```
msridhar: @liblit what's up with these changes?  everything still passes on Travis afterward?  I have no understanding of what this file is doing.

juliandolby: That file is definitions used to compile the native interface for WALA.  It is generated anyway on Travis during the build, so it is really just an example here.

com.ibm.wala.core.testdata/.settings/org.eclipse.jdt.core.prefs
liblit: I am fine with not fixing these warnings in a `testdata` project. But if the decision is being made to ignore those issues, then we should tell Eclipse to be silent about them. That is exactly what this change does: it reclassifies redundant type arguments as `ignore` instead of `warning`.

com.ibm.wala.cast/source/c/Makefile.configuration
msridhar: If this file is auto-generated, maybe it shouldn't be checked in to the repo?  In any case, @liblit can you revert your changes to this file in this PR for clarity?

liblit: This change is harmless, as @juliandolby explained, but also pointless. I will amend this pull request to remove this unnecessary churn.

msridhar: Thx @liblit.  In a separate commit / PR, I think we should mark this file as ignored.

msridhar:
Thanks again @liblit!  I will merge once Travis is green.

PR #203, opened 2017-07-15
Fix native build `Makefile.configuration` churn

liblit:
This pull request contains a few small tweaks, but its main purpose is to eliminate the churn that results from embedding site-specific paths in either of the two `Makefile.configuration` files used during native builds.  In particular:

* In `com.ibm.wala.cast.test`, the `cbuild.sh` script is gone entirely.  Instead, `pom.xml` directs Maven to run `make` directly.  We still have a `Makefile.configuration` file in this project, but this file is now independent of where WALA is being built and of where Java is installed.

* In `com.ibm.wala.cast`, a small `cbuild.sh` remains, to do some special processing involving Visual Studio variables under Windows.  When not building under Windows, `cbuild.sh` now simply runs `make`.  It may well be possible to hoist the special Windows stuff up into this subproject’s `pom.xml`, or to change that `pom.xml` to run `cbuild.sh` only when on Windows, and to run `make` directly otherwise.  I don’t know `pom.xml` stuff very well, though.  We still have a `Makefile.configuration` file in this project, but this file is now independent of where WALA is being built and of where Java is installed.

The changes I’ve made in both `Makefile.configuration` files use GNU make extensions.  I assume that’s safe because `Makefile.definitions` already relies on GNU make.

**Important disclaimer:** I have not actually tested these changes under Windows, and I think that Travis-CI doesn’t test under Windows either. It would be wise for someone with a suitable setup to try this out under Windows before merging this pull request.


msridhar:
Nice!  @juliandolby can you look this one over and merge if you don't see any issues?

juliandolby:
looks good!  the previous files were hackish in many ways, really just enough to get stuff to pass.  This looks much better.  Thanks!

juliandolby:
Oh, yeah, _no one_ _ever_ tests on Windows, and we do not expect it :)

PR #207, opened 2017-07-27
Remove "throws XYZ" declarations where XYZ cannot be thrown

liblit:
Unnecessary `throws` declarations tend to cascade.  If `foo()` calls `bar()` and `bar()` falsely declares that it might throw `IOException`, that often leads a programmer to declare that `foo()` might throw `IOException` as well.  Fixing the `bar()` `throws` declaration then reveals that we can fix the `foo()` `throws` declaration too.  By the time we reach a fixed point with cleaning these up, we have removed roughly 320 unnecessary `throws` declarations.

In a few cases, this cleanup even lets us remove entire `try` ... `catch` statements where the only thing being caught was an exception that we now statically know cannot be thrown.  Nice!

In Eclipse project configurations, upgrade any future such shenanigans from warnings to errors.  Now that we’ve fixed this, we don’t want it coming back again.

There is a potential drawback to this change.  Conceivably some public WALA API entry point might have declared that it could throw some exception merely to reserve the *option* of throwing that exception in third-party code that subclasses and overrides the API entry point in question.  I have no idea whether this is a significant concern in practice, though.


msridhar:
Thanks!

PR #208, opened 2017-07-30
Fix or suppress all 242 Eclipse warnings about unused parameters

liblit:
Also, treat unused parameters as errors, not merely warnings. We’ve fixed or suppressed all such warnings, except in projects containing test inputs.  Let’s make sure no more appear in the future.

In general, my approach was to try to eliminate each unused parameter using Eclipse’s “Change Method Signature” refactoring.  That did not always succeed: a parameter may be unused in some base class method, but then be used in subclass’s override of that method.  In cases where refactoring to eliminate a parameter failed, I instead annotated the parameter with `@SuppressWarnings("unused")` to silence the warning.

Note: this group of changes creates a significant risk of incompatibility for third-party WALA code.  Some removed parameters change externally-visible APIs.  Furthermore, these changes do not necessarily lead to Java compilation errors.  For example, suppose third-party code subclasses a WALA class or interface, overrides a method, but does not annotate that method as `@Override`.  Removing a parameter means that the third-party method no longer overrides.  This can quietly change code behavior without compile-time errors or warnings.  This is exactly why one should use `@Override` wherever possible, but we cannot guarantee that third-party WALA users have done that.


juliandolby:
Reading Ben's message makes me wonder: do we (or should we) make the no-override-annotation warning an error?

liblit:
@juliandolby: yes, we already treat missing `@Override` as an error in all WALA Eclipse projects. This was part of fb9042d3a6c9e746f277a8ba1a86fec8c0cb8156, created during the [PLDI 2017 WALA Hack-A-Thon](http://pldi17.sigplan.org/event/pldi-2017-workshops-and-tutorials-wala-hack-a-thon). We have no control over `@Override` use in third-party code that builds upon WALA, though.

msridhar:
This is master; let's push ahead and merge in spite of the tricky issue with third-party method overrides.  If/when we create a new release, we should mention in the release notes.

PR #212, opened 2017-08-03
Suppress an Eclipse warning about lacking XML grammar info

liblit:
The single “No grammar constraints (DTD or XML Schema) referenced in the document” warning arises in a generated file. I doubt that we can change the generation process to include grammar information. Even if we could, I don’t mind omitting validation here assuming that we can trust the generator tool to be correct. Validation is much more important for human-authored XML; for tool-authored XML, we can live without it.

Warnings of this kind will be treated as errors in the future.

This was the only remaining Eclipse warning in the “Other” category.


msridhar:
Thanks @liblit!

PR #227, opened 2017-08-16
Rename `...-feature` to `..._feature` in subdirs and features

liblit:
This change affects both top-level subdirectory names as well as Eclipse plug-in feature names.  Perhaps it would have been possible to change only the latter, but I don’t like the idea of the two being different.

These name changes fix three Eclipse plug-in warnings of the form:

> Illegal value '...-feature' for attribute 'id'. Legal token characters are "a-z", "A-Z", "0-9", "_".  Tokens must be separated by "."

I’ll be the first to admit that I know nearly nothing about Eclipse plug-in development.  If changing these plug-in feature IDs has broader implications that the automated regression tests won’t detect, then I probably overlooked them too.  I would greatly appreciate skeptical review of this change by someone who knows Eclipse plug-in development well.

Note that personal Eclipse workspaces may need some manual adjustment after this change.  The three `...-feature` Eclipse projects should be removed from the workspace, and the three corresponding `..._feature` Eclipse projects should be added.  If you do your git pull using Eclipse’s team features, perhaps it is smart enough to do this for you?  I don’t know, but it wouldn’t surprise me if fixing things manually were still needed even in that case.


msridhar:
@juliandolby think you added these projects; can you take a quick look at this PR to make sure it doesn't break anything?

msridhar:
@juliandolby gentle reminder on this one

khatchad:
> I’ll be the first to admit that I know nearly nothing about Eclipse plug-in development. If changing these plug-in feature IDs has broader implications that the automated regression tests won’t detect, then I probably overlooked them too. I would greatly appreciate skeptical review of this change by someone who knows Eclipse plug-in development well.

Oh, wish I saw this earlier. This is not for developing *on* Eclipse, it's for developing *for* Eclipse. It will break any Eclipse plug-ins that rely on WALA as a plug-in in Eclipse (i.e., as a dependency). I'll make a PR that suggests adding this back in some form.

msridhar:
@khatchad rather than relying on WALA as a plug-in, could you switch to pulling in the WALA jars from Maven Central?  That will be the (much) better supported option going forward.

khatchad:
Hm, maybe ... That would certainly work for compilation, but I am wondering about the run time dependency. Indeed, if there are multiple Eclipse plug-ins that rely on WALA, each would have to have their own version of WALA. That WALA already supported OSGi-style packaging, this made it very easy for Eclipse plug-ins to integrate with WALA. I'll need to think about this ...

msridhar:
On WALA master, the `com.ibm.wala.ide` project relies on the core WALA jars, and then re-exports relevant packages to other plug-ins.  So hopefully you could just add a dependence from other plug-ins to `com.ibm.wala.ide` and then everything will work?  That would still require cloning and building WALA from source.  Alternately, you could follow the same pattern that `com.ibm.wala.ide` uses,  but add it to one of your own plug-ins.  I think it's a standard pattern with OSGi to have a single plug-in that manages and re-exports third-party dependencies.

Happy to help / advise more to get this working for you.  If we need to make changes, it would be really awesome to also have regression tests to ensure important things don't get broken by future changes.  We've been hampered by a lack of sufficient testing for our Eclipse support.

khatchad:
Oh, I did not consider that. I think that would be possible; in fact, that is sort of what we have been doing now. Except, we haven't pulled from upstream is a while and are now hamped with all of the changes to merge in.

msridhar:
If it's helpful to you, I can cut a release with the current master branch.  Then you can update to that version and not have to deal with further coming changes to master.

khatchad:
Thanks, we'll keep this in mind! First, though, I need to send a PR with some enhancements. I'll do that now.

khatchad:
> @khatchad rather than relying on WALA as a plug-in, could you switch to pulling in the WALA jars from Maven Central? That will be the (much) better supported option going forward.

OK, I have a plug-in that was relying on WALA as another plug-in. I think I'll need to convert that plug-in to a maven project in order to use WALA as a dependency from maven central.

msridhar:
If your plug-in relies on `com.ibm.wala.ide`, that is still a plug-in and also re-exports a lot of core WALA packages:

https://github.com/wala/WALA/blob/master/com.ibm.wala.ide/META-INF/MANIFEST.MF

khatchad:
Ah, that helped a lot, thanks! Unfortunately, I am still missing some. I'll probably still need to use maven.

khatchad:
Does this also mean that com.ibm.wala.ide.* packages are not available on Maven Central?

msridhar:
> Does this also mean that com.ibm.wala.ide.* packages are not available on Maven Central?

They are not.  I assumed that would not be useful.  Don't Eclipse plug-ins need to be served from a P2 repository or something like that?

khatchad:
Yes, I think you are correct.

khatchad:
OK, com.ibm.wala.ide.* is not currently included in Maven Central, but it is also not hosted on a p2 site either, correct?

msridhar:
Correct.  If there is some Maven-Central equivalent for p2 sites we can
look into hosting.  Or maybe we can host one on the WALA web site itself.
Happy to take a contribution :-)

--Manu

khatchad:
:). We host ours on GitHub. However, I think that the tricky part would be to minimize the burden of the maintainers. There would be an extra step involved in deployment. I would need to look into how to fit that into the build process. Thanks for the info!

PR #221, opened 2017-08-11
Fix for default methods, etc.

khatchad:
- Fix #219 
- Test for #219.
- Minor efficiency improvements.

PR #228, opened 2017-08-17
Minor performance improvements to methods in TypeInference class

khatchad:
But not sure why GitHub is bringing in code from #221.


msridhar:
We need to clean this up before merging.  If the `TypeInference` changes are in their own commits, try creating a new branch from `master` and then doing `git cherry-pick` for those commits.  If that's not going to work, last option is to copy out the `TypeInference.java`, make a new branch from `master`, and then copy it back.

khatchad:
This looks like a bug with GitHub. Have a look at https://github.com/wala/WALA/blob/master/com.ibm.wala.core.tests/src/com/ibm/wala/core/tests/callGraph/DefaultMethodsTest.java#L84-L91. It's exactly the same as what is in https://github.com/wala/WALA/pull/228/files.

msridhar:
It may well be a GitHub bug, but I'd still appreciate if you could clean up the commit history.  I'd like to be fully confident in what exactly is getting merged before merging it.  It shouldn't take more than a few minutes to clean it up.  Thanks!

khatchad:
OK I understand now. Done. https://github.com/wala/WALA/pull/230.

PR #231, opened 2017-08-21
Fix Eclipse warnings about resource leaks

liblit:
These changes add `try`-with-resources statements to ensure timely closure of various `AutoCloseable` instances.

A few leaks remain, all involving `JarFile` instances used by `JarFileModule` instances. Plugging those leaks will be a significant challenge, and may or may not be worth the code complexity it will entail. But we may as well fix these easy ones now; no need to wait for perfection.


msridhar:
Thanks @liblit!  Can you open an issue on the remaining resource leaks?  Would like to tackle this problem at some point.

liblit:
@msridhar: [done](https://github.com/wala/WALA/issues/236).

PR #232, opened 2017-08-24
Add missing void type.

khatchad:
You can see from: https://github.com/wala/WALA/blob/68f30f949d60b290a75bbd4d52956a412f01f91c/com.ibm.wala.core/src/com/ibm/wala/analysis/typeInference/JavaPrimitiveType.java#L62 that it's missing. In other words, That line would never print since the corresponding constant is missing.


msridhar:
Thanks!

PR #238, opened 2017-08-28
Use Collections.emptySet() instead of Collections.EMPTY_SET

liblit:
`Collections.emptySet()` returns `Set<T>` for any `T`, whereas `Collections.EMPTY_SET` merely returns
a raw `Set`.  Using the former instead of the latter fixes four Eclipse warnings about using raw generics.

In modern Java 1.8 and later, `Collections.emptySet()` is all we would need to write. Unfortunately, Android is behind the times, so this specific project targets Java 1.5. That older Java definition does not do the inference we would want here, so we have to be explicit instead and use `Collections.<TypeReference>emptySet()`.


msridhar:
Sorry, what do you mean that Android is behind the times?  I thought we are already leveraging the recent Android Java 8 support?  I don't see why the Dalvik project would need to use source level 1.5.  @juliandolby?

liblit:
My first attempt at this fix used `Collections.emptySet()`, not `Collections.<TypeReference>emptySet()`. The Travis-CI run failed. While investigating why, I came across the following lines in `com.ibm.wala.dalvik/build.xml`:

```xml
<property name="javacSource" value="1.5"/>
<property name="javacTarget" value="1.5"/>
```

I assumed that these were there for the purpose of selecting Java 1.5 for compilation, and that whoever did that did it for good reason.

Changing `Collections.emptySet()` to `Collections.<TypeReference>emptySet()` fixed the Travis-CI failures. We use `Collections.emptySet()` just fine in some other WALA Eclipse projects, but apparently `com.ibm.wala.dalvik` is still being built using an earlier version of the Java language.

msridhar:
I can't think of any reason for that setting.  @liblit can you change both those values to `"1.8"` and see if you can erase the explicit type parameters then?

liblit:
@msridhar: I tried your suggestion, but as you can see, it does not work. Inference on `Collections.emptySet()` works fine in at least some other projects, so there must be something that is configuring this one to use a pre-1.8 version of Java. I have no idea what that might be.

msridhar:
@liblit yeah the `pom.xml` for the Dalvik project is a mess.  I have an idea for how to fix, but let's separate that.  You can revert the change I suggested and then we'll land this PR (with the explicit type annotations in the dalvik project) for now.

juliandolby:
The reasons for this are historical.  In the past, we used versions of Android tools that did not support modern Java, since they were available from maven central.  

Now we are using the latest and greatest Android tools anyway, so there is no longer any need for any of these hacks.

  — Julian

liblit:
@msridhar: done.

PR #240, opened 2017-08-28
Missing source class loader reference.

khatchad:


khatchad:
Nevermind, found it here: `com.ibm.wala.cast.java.ipa.callgraph.JavaSourceAnalysisScope.SOURCE`.

PR #248, opened 2017-09-17
Enable ErrorProne on wala.util project

msridhar:
Since @liblit was running out of warnings to fix... 😄 This PR enables [Error Prone](http://errorprone.info/) for building the Maven jar of wala.util.  I fixed most of the warnings it emitted, but turned a couple checks off since the issues needed more careful study.

The downside of this change is that Error Prone doesn't run during the standard Eclipse build.  Locally, you can run it by running `./build-maven-jars.py`, but you need to remember to do that.  Since the `wala.util` code is quite stable I think it's reasonable to enable here and experiment.

As you can see by the commented lines in `mvncentral.xml`, another goal is to try [NullAway](https://github.com/uber/NullAway) out on WALA to build up more robustness against NPEs.  But one step at a time.

### DIFF
--- a/com.ibm.wala.util/mvncentral.xml
+++ b/com.ibm.wala.util/mvncentral.xml
@@ -69,11 +69,43 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
-        <version>3.2</version>
+        <version>3.5</version>
         <configuration>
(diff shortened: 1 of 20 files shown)


msridhar:
@liblit comments welcome

liblit:
The changes you have made all seem reasonable to me.

msridhar:
Ok, let's give it a shot

PR #264, opened 2017-11-23
Update Maven Central configurations to also use Java 8

liblit:
E-mail exchanged with @juliandolby suggests that this is the right thing to do, and that it should have been done back when we converted other parts of the build configuration to Java 8.


msridhar:
LGTM, thanks!

PR #265, opened 2017-11-26
Update more parts of the build configuration to use Java 8

liblit:
@juliandolby assures me that WALA is now supposed to be using Java 8 everywhere.  This pull request covers nearly all remaining places that I can find where I *think* that an earlier Java version was still being used.  (The few exceptions are places where switching to Java 8 causes test failures. I'll address those separately, probably by reaching out to the WALA maintainers for help.)


msridhar:
@liblit have you confirmed that all of these `build.xml` files are actually used?  I know some get invoked from the Maven build, but I am wondering if some can just be deleted.

liblit:
No, I have not confirmed that they are all used. I had not considered the possibility that some or all of this is dead content. If you wish, I can revise this pull request to remove any `build.xml` that does not seem to be invoked by the corresponding `pom.xml`. Certainly Travis-CI should be able to catch me if I accidentally recommend removing something that is still needed.

msridhar:
If you have time that would be awesome 👍

PR #266, opened 2017-11-27
Remove "build.xml" files not used by corresponding "pom.xml" files

liblit:
@msridhar requested this clean-up before proceeding with #265. I will revisit the latter after this pull request is merged.


msridhar:
Note to self: once this merges we should update the getting started guide to indicate that running Maven is mandatory before opening WALA in Eclipse.

Issue #268, opened 2017-11-27
Java 8 not used for some fragments of `wala.core.testdata`


liblit:
@juliandolby assures me that WALA should now be using Java 8 throughout. Recent pull requests #264 and #265 updated several parts of the build configuration to be consistent with this intent. However, a few fragments of `com.ibm.wala.core.testdata` are still using earlier Java versions. My attempts to update these lead to regression test failures. I'm documenting these here and explicitly asking for help from WALA maintainers or anyone else who understands the build system more completely than I do.

<hr/>

The first pre-Java-8 area can be seen on [line 38 of `com.ibm.wala.core.testdata/build.xml`](https://github.com/wala/WALA/blob/5ac8bf881da01d1e056e54ba42b750ed0515784a/com.ibm.wala.core.testdata/build.xml#L38):

``` build.xml
<javac target="1.4" source="1.4" includeAntRuntime="no" srcdir="${temp.folder}" destdir="${temp.folder}"/>
```

The second pre-Java-8 area appears on [lines 168–169 of the same file](https://github.com/wala/WALA/blob/5ac8bf881da01d1e056e54ba42b750ed0515784a/com.ibm.wala.core.testdata/build.xml#L168):

``` build.xml
<param name="jsv" value="1.6"/>
<param name="jtv" value="1.6"/>
```

I have [changed these instances of `1.4` and `1.6` to `1.8`](https://github.com/wala/WALA/compare/master...liblit:java-8-in-wala.core.testdata) on an experimental branch. Unfortunately, Travis CI reports that this build fails. In the Travis CI logs, we see two different kinds of problems.

First, we have [a series of assertion failures](https://travis-ci.org/liblit/WALA/builds/307992060#L1154) of the form

```
java.lang.AssertionError: <Application,Lannotations/AnnotatedClass1> not found
```

Then we have [some sort of Java VM crash or failure](https://travis-ci.org/liblit/WALA/builds/307992060#L1365):

```[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.14.1:test (test) on project com.ibm.wala.dalvik.test: Execution test of goal org.apache.maven.plugins:maven-surefire-plugin:2.14.1:test failed: The forked VM terminated without saying properly goodbye. VM crash or System.exit called ?```

<hr/>

Can anyone figure out how to get these last few pieces caught up to Java 8 like the rest of WALA?

liblit:
[...]

PR #269, opened 2017-11-27
Use Java 8 lambdas and method references wherever possible

liblit:
Eclipse’s automated code clean-up tool did most of the heavy lifting here: it specifically has a clean-up option for converting functional interfaces to lambdas. I merely had to revert the automated changes for a single enumeration class for which it produced invalid results, and for a few test inputs that apparently [aren't set up to be compiled with Java 8](https://github.com/wala/WALA/issues/268).

Surprisingly, Eclipse does *not* have an automated clean-up option to use method references instead of lambdas where possible, so that part I [did by hand](https://github.com/wala/WALA/commit/74e0640771c3f5a701abed3681ad0a4b10730e54).


msridhar:
Sorry typing on my phone. @juliandolby can you check that this works in the browser use cases you're using?

com.ibm.wala.cast.java.test/src/com/ibm/wala/cast/java/test/JavaIRTests.java
```diff
@@ -82,103 +82,78 @@ public JavaIRTests() {
   @Test public void testTwoClasses() throws IllegalArgumentException, CancelException, IOException {
     runTest(singleTestSrc(), rtJar, simpleTestEntryPoint(), Arrays.asList(
 
-    new IRAssertion() {
-
-      @Override
-      public void check(CallGraph cg) {
-        final String typeStr = singleInputForTest();
+    (IRAssertion) cg -> {
```
msridhar: @liblit Why do we need the `(IRAssertion)` cast here?

liblit: We don’t. I will remove the unnecessary cast here and at several other places in this same file. Thank you!

PR #267, opened 2017-11-27
Gradle build scripts

SunghoLee:
Enable WALA being built by Gradle.

PR #278, opened 2017-12-26
Build WALA using Gradle instead of Maven

liblit:
This pull request represents work-in-progress to convert WALA’s build system from Maven to Gradle. It also adds support for Eclipse’s Buildship plugin for better Eclipse/Gradle integration.

We still have much left to do before Gradle can truly replace Maven. However, @msridhar [suggested](https://github.com/wala/WALA/pull/267#issuecomment-353740080) that a premature pull request would still be useful “for visibility / easy commenting,” so that is what we have here. I will keep working on this as time permits, but I make no exclusive claim to it. I welcome pull requests and constructive comments from any and all! I am very new to Gradle, and there are surely better ways to do some of what I am doing here. Please help me make things better.

Big thanks to @SunghoLee for getting us started down this path!


liblit:
I have opened [several GitHub issues](https://github.com/liblit/WALA/milestone/1) in my own `gradle-and-buildship` branch to record things that I know need to be improved before Gradle can replace Maven. This is certainly not an exhaustive list; it is just the things I am aware of right now.

None of these issues are assigned to me specifically, and that all of them are tagged `help wanted`. These are all excellent targets for anyone looking to contribute to the Gradle-based WALA effort. Please let me know if you intend to take any of these on; I will assign the appropriate issue(s) to you so we can avoid duplicated work.

From December 30 through January 9, I will have occasional e-mail, but no desktop, no laptop, and no physical keyboard of any kind. Expect delays in my responses to any pull requests or other nontrivial tasks. On the other hand, this is an *excellent* opportunity for anyone else to come along and contribute useful improvements without worrying about conflicting with my efforts.

SunghoLee:
@liblit, how can commit some codes to your repository? I tried to commit something, but the access is denied because of permission.

liblit:
@SunghoLee, please submit [pull requests](https://help.github.com/articles/about-pull-requests/) for your changes. This should be similar to what you did to offer your original Gradle work to WALA, except that you'll be submitting pull requests for my `gradle-and-buildship` branch rather than WALA’s `master` branch.

Thanks!

liblit:
173 commits later, we now have a delightful Gradle-based build system. It’s not yet a drop-in replacement for Maven, but it has several significant advantages. For the known limitations, I am at a point where I will really *need* help from other experts in order to make progress. It is hard to attract the attention of such experts as long as this work is separated from mainstream WALA development. Therefore, I propose that we now merge this Gradle work back into WALA master.

Such a merge should not be done lightly! WALA maintainers (@msridhar, @juliandolby, etc.): please read the [WALA+Gradle README](https://github.com/liblit/WALA/blob/gradle-and-buildship/README-Gradle.md) that describes advantages, limitations, new Eclipse import procedures, and other important highlights. Given what’s written there, do you think that it's time to accept this pull request? Or should some specific other work be done first, and if so, who can help me out with that?

msridhar:
@liblit this looks amazing!  I am in favor of merging this PR.  I am hoping we can get rid of the Maven build ASAP, or at least disable testing on CI.  One thing that I will take on after this merges is getting rid of the `mvncentral.xml` files and adding support for uploading MVN Central artifacts via the Gradle build.  The other major-ish thing seems to be the Eclipse plugin stuff.  This is not relevant to me, and I'm not sure how much @juliandolby needs it.  I am also not even sure how much Travis actually tests that things are being done correctly here.  Hopefully we won't need to keep Maven builds on Travis just for this.

Anyway, I think @juliandolby should sign off before we merge back, but I approve 😄

juliandolby:
@liblit @msridhar I agree this looks cool!  I would like to keep maven for now, partly for the Eclipse plugin stuff and partly so I have some time to port the various child projects to gradle.  But I am also in favour of merging.  

I will next try following the build instructions, but assuming that works, then let's merge.

juliandolby:
I just checked out WALA master, merged this pull request, and ran "./gradlew build" on my Mac, and everything seemed to work, with tests passing!

And, furthermore, it seems that my latest WALA work with analyzing machine learning code will be helped by integrating WALA into PyCharm, which more-or-less requires gradle.  So this is likely to be useful to me right away.  (Although, for the time being, I still need Maven as well)

I think we should merge it.

liblit:
That’s welcome news, @juliandolby.

Would either you or @msridhar care to try [the Eclipse instructions](https://github.com/liblit/WALA/blob/gradle-and-buildship/README-Gradle.md#eclipse) before pulling the lever on this merge?

msridhar:
@juliandolby can you try the Eclipse instructions?  I'm not set up with Eclipse at the moment.

Regarding merging, I've gotten in the habit of rebasing before merging in other repos, to maintain a more linear commit history (makes things like bisecting much easier).  I usually also squash commits before merging, under the assumption that each individual PR should be small enough to fit in a single commit.

In this case, the PR is fairly large 😄 @liblit how interested are you in preserving the commit history?    If you don't need to preserve the commit history, we could just squash and then merge as one large commit (which I'm ok with in this case).   If you strongly prefer to preserve commit history, the best thing would be to rebase on master first.  But that might be pretty painful as it goes commit by commit in the branch, so you might end up fixing the same conflicts repeatedly. If we want to preserve commit history *and* rebasing is too painful, we can fall back on using a merge commit.  IMO squashing is the best option.

liblit:
Four months of development history and commit messages may have some value; I don’t like the idea of squashing and thereby throwing all of that history away. Perhaps rebasing onto `master` is the best compromise between keeping things simple and preserving history as it really happened.

I have successfully rebased `gradle-and-buildship` onto `master`. [All Travis CI tests still pass.](https://travis-ci.org/liblit/WALA/builds/355144978) For now, I have retained [the non-rebased branch](https://github.com/liblit/WALA/tree/gradle-and-buildship-before-rebase) too just in case we need (or decide) to switch back to that instead.

msridhar:
Cool, the "Rebase and merge" button is now green for me 😄

liblit:
Unfortunately I just noticed that two tests are failing in the Travis CI build of my `gradle-and-buildship` branch, only when built by Maven:

* [`com.ibm.wala.cast.java.test.JDTJava15IRTests`](https://travis-ci.org/liblit/WALA/jobs/355657100#L1036)
* [`com.ibm.wala.cast.java.test.JDTJavaIRTests`](https://travis-ci.org/liblit/WALA/jobs/355657100#L1386)

Travis reports that this job succeeded, even though it clearly ran into problems. So I need to (1) fix these two failures and also (2) figure out why Travis incorrectly treated the job as successful. We should not merge until I have resolved this.

That being said, @juliandolby should still feel free to try out [the Eclipse instructions](https://github.com/liblit/WALA/blob/gradle-and-buildship/README-Gradle.md#eclipse) whenever he has the time. No need to wait until I correct these Maven regressions.

liblit:
I’m done. I give up.

The two failing tests I [mentioned earlier](https://github.com/wala/WALA/pull/278#issuecomment-374459680) are among the three that I have [never been able to run successfully under Gradle either](https://github.com/liblit/WALA/issues/5). It is unlikely that I will be able to get them working again under Maven, about which I know even less. I have been banging my head against this for a few days now with no success. When I try to reproduce this problem on my own machine, these two tests fail regardless of which earlier version I try, even going all the way back to WALA master. This means that I cannot even bring tools like `git bisect run` to bear in helping me isolate the regression-inducing change. So I do not know how to make these tests work in _any_ revision or configuration, even though they apparently do work in Travis CI tests of WALA master.

Does anyone else on this pull request have the knowledge needed to diagnose and fix these failing tests? If so, go for it. Submit a pull request on [my `gradle-and-buildship` branch](https://github.com/liblit/WALA/tree/gradle-and-buildship) that shows me what I broke and how to fix it. If nobody steps up to help me with this, then I guess we just don’t get to have a shiny new build system after all, and these last four months of Gradle effort on my part were nothing more than a throwaway learning exercise.

I have corrected the meta-bug that allowed Travis CI tests of the Maven build to succeed even though these tests were failing. So you can now see that the `gradle-and-buildship` branch [is indeed broken](https://travis-ci.org/liblit/WALA/jobs/356643549#L1038). If you want to see it fixed one day, then somebody needs to step up and lend me a hand.

juliandolby:
This is looking great.  I merged with WALA master, and then built using Gradle, imported into Eclipse, and then ran tests using Maven.  Everything seems to work.

I am not worried any more about Eclipse project packaging, and anyway I tested that we can reference the Gradle Eclipse projects in a new Plugin project.  So even if someday we want to package stuff in Eclipse, we ought to be able to.

My only remaining question is about the various GitHub projects that use WALA and are tested using Maven build scripts on Travis.  After the merge, will simply cloning WALA and calling 'mvn clean install -DskipTests" be sufficient to install our maven packages for child projects to use?

liblit:
@juliandolby, thank you for giving this your attention.

If the [failures of some Eclipse-related tests under Maven](https://github.com/wala/WALA/pull/278#issuecomment-375141177) do not concern you, then I propose that we disable these tests in the appropriate `pom.xml` files. That way Maven Travis CI testing goes back to passing status for the tests that remain. I am happy to implement this if you agree that I should.

Those Eclipse-related concerns aside, yes: to the best of my knowledge, `mvn clean install -DskipTests` will still work as before. Per the second paragraph of [my discussion of known shortcomings](https://github.com/liblit/WALA/blob/gradle-and-buildship/README-Gradle.md#known-shortcomings), either Gradle or Maven can be used to build WALA from the command line. Building in Eclipse does not offer both options, but the command line does. I believe that I have kept `mvn clean install -DskipTests` working as it always did. That being said, it certainly wouldn’t hurt to actually try out some of these “various GitHub projects that use WALA and are tested using Maven build scripts on Travis” to see whether they still work if changed to use my `gradle-and-buildship` branch instead of WALA `master`.

Two other thoughts if we may finally be on the cusp of merging this pull request:

1. This may warrant bumping the minor version number from `1.4.4-SNAPSHOT` to `1.5-SNAPSHOT`. That leaves you the option of doing future `1.4.5+` releases from of a pre-Gradle branch if these merged Gradle changes turn out to be more disruptive than expected. I can make this change if you agree that I should.

2. @msridhar [proposed rebasing these changes on the latest WALA `master`](https://github.com/wala/WALA/pull/278#issuecomment-374028801), which [I did](https://github.com/wala/WALA/pull/278#issuecomment-374063853). However, `master` has continued to evolve since then. I am happy to rebase again if we really are just about ready to merge. I prefer not to rebase over and over, though, as doing so (1) requires moderate effort and (2) creates some disruption due to invalidation of hashes in existing clones.

juliandolby:
I found a weird issues with our nodejs support when running mvn clean install after importing with buildship, but i seem to have fixed it.  And all the Eclipse-based tests---JDT, JSDT and ECJ---passed for me with maven, after I imported with buildship.  So now, at least for me, everything seems to work!

Do you know if it is possible to run buildship from the command line?  If so, we could run it during integration, which should create the metadata we need.  Otherwise, can we have the Eclipse metadata in github only for the com.ibm.wala.cast.{java,js}.test.data projects?  They are used by Eclipse-based tests, and in some sense are meant to be proper Eclipse plugins.  I suspect that the Eclipse-based tests need it, since they explicitly try to read Eclipse project state.

liblit:
_Very_ interesting! Thank you for this additional information, @juliandolby. It seems you are right that some part of the Eclipse metadata is needed in order for those tests to pass. Given this insight, I see three options:

# Use Gradle to Generate Eclipse Metadata Before or During Maven Builds

We cannot run Buildship from the command line. However, we can ask Gradle to generate Eclipse project metadata, specifically the various `.project` and `.classpath` files. That is really what happens during Buildship imports anyway. So we could do that before a Maven build, or perhaps even have Maven run Gradle automatically when needed.

If Gradle creates `com.ibm.wala.cast.{java,js}.test.data/.{classpath,project}` correctly right out of the box, great! If we need to tweak the contents of those files, that can be done too as part of these projects’ `*/build.gradle` configurations. We are already applying several such tweaks, in fact. Look for mentions of `eclipse.project` in existing `*/build.gradle` files, such as [here in `com.ibm.wala.cast.java.test/build.gradle`](https://github.com/liblit/WALA/blob/96be02d77c583c49f964ffaf6b70b4b282123c80/com.ibm.wala.cast.java.test/build.gradle#L4)

The **advantage** of this approach is that the Gradle-generated metadata should exactly match what would result from an interactive Buildship import. That keeps things simple and consistent. Another **advantage** is that this may very well fix the [remaining IDE-related test failures under Gradle](https://github.com/liblit/WALA/issues/5) too. The **disadvantage** of this approach is that it makes Gradle an essential dependency for a Maven build: you can no longer build using Maven alone without Gradle being involved in some small way.

# Commit Selected Eclipse Metadata to Git

If any `.classpath` and/or `.project` files already exist, then importing using Buildship will read their contents, merge in any other Gradle-derived information it thinks should be there, and write the modified files back out. This could be an annoyance for version control, since then we have a file acting as both input and output. So we would want to ensure that the metadata files in the repository are identical to the rewritten-in-place ones after a Buildship import. Otherwise we end up with bad version-control churn.

Looking at `com.ibm.wala.cast.{java,js}.test/.{classpath,project}`, I see only one thing that could be problematic. My `com.ibm.wala.cast.js.test/.project` file, generated locally by Buildship, includes the following lines:
``` xml
<linkedResources>
  <link>
    <name>examples-src</name>
    <type>2</type>
    <location>/home/liblit/Documents/research/joust/WALA/com.ibm.wala.cast.js.test.data/examples-src</location>
  </link>
</linkedResources>
```

Clearly that absolute path is not suitable for committing to a shared repository. These lines arise due to the following piece of the corresponding `com.ibm.wala.cast.js.test/build.gradle` configuration:

``` gradle
sourceSets.test {
  java.srcDirs = ['harness-src']
  resources.srcDirs = [project(':com.ibm.wala.cast.js.test.data').file('examples-src')]
}
```

I could certainly look into ways of tweaking that Gradle configuration to avoid the absolute path name. I am confident it can be done if we decide that we prefer this approach.

The **advantage** of this approach is that it avoids making Maven builds depend on Gradle. The **disadvantage** of this approach is that it deviates from the pattern that `*/.{classpath,project}` are output files only, generated purely from the Gradle configuration. And of course, this all assumes that I can eliminate the problematic absolute path mentioned above.

# Commit Selected Eclipse Metadata to Git Under Alternate Names

Instead of storing `com.ibm.wala.cast.{java,js}.test.data/.{classpath,project}`, we could bring back these files (as needed by Maven) under alternate names, such as `com.ibm.wala.cast.{java,js}.test.data/maven-{classpath,project}`. This keeps them out of Buildship’s way. On the Maven side, we would need to arrange for these alternate `maven-classpath` and `maven-project` files to be used where needed instead of the default `.classpath` and `.project` files. Or perhaps part of a Maven build would involve copying `maven-classpath` and `maven-project` into `.classpath` and `.project`, respectively? I am not really sure what would be needed, since I don't understand what happens within Maven that leads to these Eclipse metadata files being used in the first place.

---

@juliandolby, @msridhar: any thoughts?

juliandolby:
I like the first option best, I think.  If maven can launch a small gradle script that creates the needed metadata, that seems best because it keeps the invariant eclipse metadata is generated.  And requiring gradle in addition to maven seems like a small price since WALA already sucks down a lot of software.

Perhaps we can even have a little gradle script that creates the metadata for the needed projects and also downloads stuff like android so that the maven build can run...

liblit:
OK, I will start working on the option to have Gradle create the `com.ibm.wala.cast.{java,js}.test.data/.{classpath,project}` files. The Gradle side of that should be easy, and will likely be triggered by a simple command like `./gradlew prepareMavenBuild`. I may need help figuring out how/where/when Maven should run that external command.

Regarding using Gradle to also download other supporting material, such as the Andoid SDK, I recommend that we defer that for now. Yes, our Gradle build system has [excellent knowledge](https://github.com/liblit/WALA/blob/gradle-and-buildship/README-Gradle.md#external-dependencies-patience-is-a-virtue) of what to download, where to get it from, and even the expected checksums so that we know that we are getting what we expected. It even knows how to unpack and install the required Android SDK components. We could certainly leverage that to ease setup before Maven builds. But that is an _enhancement_ for Maven builds; it is not strictly _needed_ in order for this pull request to be ready for merging without disrupting Maven builds as they have worked up until now.

juliandolby:
Deferring my wish list makes sense.  As for how to invoke the gradle script to prepare for maven, I think we could use use the maven exec plugin.

liblit:
More clues:

1. If I use Gradle to create `com.ibm.wala.cast.{java,js}.test.data/.{classpath,project}` files that are similar (but not identical) to those created by Buildship, then [Maven fails the same problematic tests in the same way as before](https://travis-ci.org/liblit/WALA/builds/366663098#L1331).

1. If I keep the exact `com.ibm.wala.cast.{java,js}.test.data/.{classpath,project}` files created by a real Buildship import, then [Maven fails the same problematic tests in the same way as before](https://travis-ci.org/liblit/WALA/builds/366663069#L1041).

1. If I keep _all_ of the Eclipse metadata files created by a real Buildship import, then [Maven passes the problematic tests](https://travis-ci.org/liblit/WALA/builds/366663125#L1041), but [fails some other test later by running out of memory](https://travis-ci.org/liblit/WALA/builds/366663125#L1265). So I guess that’s progress?

So it seems that _something_ created during Buildship import is important for getting the problematic Maven tests to pass, but that it isn’t (or isn’t only) `com.ibm.wala.cast.{java,js}.test.data/.{classpath,project}`. I can systematically delta-debug my way through the entire set of 63 Buildship-generated files, but that is going to be tedious. @juliandolby, do you have any well-informed guesses as to what subset of the [63 Buildship-generated files](https://github.com/liblit/WALA/commit/941def449588ceb978c7d395895cd8a3691490b6) are more or less likely to be important?

juliandolby:
I suspect that the out-of-memory that you are seeing is the same one that I ran into and seem to have been able to fix.  Is it in the com.ibm.wala.cast.js.nodejs.test project?

liblit:
Yes:

```
Running com.ibm.wala.cast.js.nodejs.test.NodejsRequireTargetSelectorResolveTest
Exception in thread "ThreadedStreamConsumer" java.lang.OutOfMemoryError: Java heap space
```

liblit:
@juliandolby, unfortunately your fix partially reverts https://github.com/liblit/WALA/commit/97dd5ac7456df85af96c7e8ef318680828d1d130, and therefore may cause regressions for concurrent testing. Would you please rework your fix to use `Files.createTempFile(...)` instead of creating invariant file names using the `java.io.tmpdir` system property? It should be easy enough to see how to use `Files.createTempFile(...)` here by looking at [the “before” version of that code](https://github.com/wala/WALA/pull/278/commits/ae68b2c95adca0bd13d0270d919266b6cb5a7d4f#diff-32e4af253df15bff6148d0437d587148L56) in the diff from your change. Thanks!

liblit:
Meanwhile, reducing the 96 Buildship-generated Eclipse metadata files down to the few that matter was a perfect use case for delta debugging. It turns out that what we really need are `com.ibm.wala.cast.java.test.data/.{classpath,project}`. If those two Buildship-generated files are present before running `mvn clean install`, then the problematic `com.ibm.wala.cast.java.test.JDTJava{,15}IRTests` tests pass.

I will need to examine this more deeply to fix it in a good way that works for Eclipse users, command-line users, Maven users, Gradle users, etc. But this is definitely a step in the right direction.

liblit:
Hmm. I reported [above](com.ibm.wala.cast.java.test.JDTJava{,15}IRTests) that delta debugging found that `com.ibm.wala.cast.java.test.data/.{classpath,project}` are needed in order for `com.ibm.wala.cast.java.test.JDTJava{,15}IRTests` to pass. But I had [already tried keeping those (case 3)](https://github.com/wala/WALA/pull/278#issuecomment-381375339) (along with two other files) and found that it did not work. So now I am even more confused. Sigh. I will keep investigating.

juliandolby:
@liblit I think using createTempFile is what may be causing the problems because it made analysis sometimes appear to think that repeated uses of the same core module are different since the class name is derived from the file name. 

To keep the same file name, I am now creating a temporary directory using Files.createTempDirectory and then using a single file name in that directory.  

The code still appears to run for me, and I think temporary directories ought to keep things safe in a concurrent setting.

liblit:
Ah, I can easily imagine [my temporary-file changes](https://github.com/liblit/WALA/commit/97dd5ac7456df85af96c7e8ef318680828d1d130) breaking anything that required a specific, fixed file name. @juliandolby, your use of a temporary _directory_ instead should work just fine, modulo the [unnecessary extra `deleteOnExit` call that I pointed out earlier](https://github.com/liblit/WALA/commit/118079b9be0480d9e454d4fb269c2ab544c715bb#commitcomment-28619643). Thank you!

liblit:
Oops, disregard my [claim](https://github.com/wala/WALA/pull/278#issuecomment-382058852) that https://github.com/wala/WALA/pull/278/commits/118079b9be0480d9e454d4fb269c2ab544c715bb introduces an unnecessary extra `deleteOnExit` call. Apparently `deleteOnExit` does _not_ recursively delete non-empty directories. So `deleteOnExit` calls are needed for both the temporary file and the temporary directory containing it: exactly what @juliandolby already did.

liblit:
Hooray! The problematic `com.ibm.wala.cast.java.test.JDTJava{,15}IRTests` are finally passing when run under Maven using the latest `gradle-and-buildship` revision. Big thanks to @juliandolby for the [insight](https://github.com/wala/WALA/pull/278#issuecomment-381333595) that broke this hard nut open.

We are nearly ready to merge this all back into WALA `master`. Two last things to consider doing:

1. This may warrant bumping the minor version number from `1.4.4-SNAPSHOT` to `1.5-SNAPSHOT`. That gives the WALA maintainers the option of doing future `1.4.5+` releases from of a pre-Gradle branch if these merged Gradle changes turn out to be more disruptive than expected. I can make this change if you agree that I should.

2. @msridhar [proposed rebasing these changes on the latest WALA `master`](https://github.com/wala/WALA/pull/278#issuecomment-374028801), which [I did](https://github.com/wala/WALA/pull/278#issuecomment-374063853). However, `master` has continued to evolve since then. I am happy to rebase again if we really are just about ready to merge. I prefer not to rebase over and over, though, as doing so (1) requires moderate effort and (2) creates some disruption due to invalidation of hashes in existing clones.

juliandolby:
Horray indeed!  And a (much) bigger thanks to @liblit for pushing this.  Let's bump the version to 1.5-SNAPSHOT, rebase off the latest master, and then merge this pull request and start using Gradle!

liblit:
- 3 contributors
- 5 months of effort
- 191 commits
- 259 changed files
- 1,548 deleted lines
- 2,513 inserted lines

And most importantly…

- **_[1 Travis CI build with all 3 jobs passing](https://travis-ci.org/wala/WALA/builds/367875387)_**

It’s time. Let’s merge this puppy.

msridhar:
Congratulations!!  🎉🎉🎉  Since he helped out I'll let @juliandolby have the satisfaction of clicking the merge button 😄

liblit:
Thanks for merging, @juliandolby. But it looks like you did a “squash and merge.” I thought [we agreed not to do that](https://github.com/wala/WALA/pull/278#issuecomment-374063853). Five months of development history and commit messages may have some value; I don’t like the idea of squashing and thereby throwing all of that history away.

As a specific, recent example, consider my commit message from https://github.com/wala/WALA/pull/278/commits/742cad58e97a028bf4e884048462e6ad76ebbcd1. That message includes useful information that would eventually help you implement [your wish](https://github.com/wala/WALA/pull/278#issuecomment-381368703) to “have a little gradle script that creates the metadata for the needed projects and also downloads stuff like android so that the maven build can run.” With a squash merge, that information is thrown away.

liblit:
An explanatory note for future reference. #278 was initially squashed and merged as https://github.com/wala/WALA/commit/8f41069b23bb996bde3877e9a212302dcb72275f. However, subsequent discussion led us to retroactively prefer merging without squashing. So #298 reverted the squashed merge in https://github.com/wala/WALA/commit/99c2493e373b62ff101f61453e6a12f8941ff490. Lastly, #299 re-merged the same commits without squashing, bringing us to https://github.com/wala/WALA/commit/a6096e6d7591168f080868c2953be1dbd60f1ec3.

PR #287, opened 2018-03-08
Protected visibility for JavaSourceLoaderImpl's internal methods and classes

pierlauro:
Since JavaSourceLoaderImpl is an abstract class, I would recommend to give visibility of internal fields at least to all the implementing classes in order to allow user modifications.


msridhar:
I'm fine with this change.  Any objections @juliandolby?

PR #299, opened 2018-04-18
Build WALA using Gradle instead of Maven

liblit:
This is a reissue of #278 so that we can re-merge it without squashing. See #278 for a more complete view of this effort’s history, discussions, etc.


liblit:
An explanatory note for future reference. #278 was initially squashed and merged as https://github.com/wala/WALA/commit/8f41069b23bb996bde3877e9a212302dcb72275f. However, subsequent discussion led us to retroactively prefer merging without squashing. So #298 reverted the squashed merge in https://github.com/wala/WALA/commit/99c2493e373b62ff101f61453e6a12f8941ff490. Lastly, #299 re-merged the same commits without squashing, bringing us to https://github.com/wala/WALA/commit/a6096e6d7591168f080868c2953be1dbd60f1ec3.

PR #311, opened 2018-05-23
Have Travis CI periodically try to build each Gradle subproject separately

liblit:
In theory, if Gradle dependencies are set up correctly, then it should be possible to build any subproject starting with a pristine tree. If subproject *X* depends on subproject *Y*, then Gradle should know that building *X* requires building *Y* first. The way to test this is to systematically build each subproject starting from a pristine WALA tree. Unfortunately, with 29 subprojects on 2 supported platforms, that can take a while. Travis CI runs up to five jobs at a time, and running this entire set of tests takes nearly 7 hours. So that’s not something we want to do on every push or on every pull request. However, this would be reasonable to do less frequently, such as once per week.

Fortunately, Travis CI has support for running periodic (e.g., weekly) tests: [cron jobs](https://docs.travis-ci.com/user/cron-jobs/). This pull request adds a 29 × 2 single-subproject build tests that will only be enabled when run from a Travis CI cron job. If you think this is a good idea, then you should **do two things:**

1. Accept this pull request.
2. Follow the [Travis CI cron job instructions](https://docs.travis-ci.com/user/cron-jobs/) to add a weekly cron job using the following configuration:
    * Branch: master
    * Interval: weekly
    * Options: Always run

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,3 +38,255 @@ matrix:
     - os: osx
       env: BUILD_SYSTEM=gradle
       language: java
+    - os: linux
+      env: BUILD_SYSTEM=gradle BUILD_ONLY_SUBMODULE=com.ibm.wala.cast
+      language: java
+      jdk: oraclejdk8
+      if: type = cron
+    - os: linux
+      env: BUILD_SYSTEM=gradle BUILD_ONLY_SUBMODULE=com.ibm.wala.cast.java
+      language: java
+      jdk: oraclejdk8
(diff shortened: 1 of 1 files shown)


liblit:
Oops, I forgot some pieces of this. Closing for now; I will reopen when ready.

liblit:
GitHub forbids reopening this request due to various git shenanigans that I performed on my branch since closing this request. So I have instead opened a new pull request, #315, which is now ready to go.

PR #315, opened 2018-05-24
Have Travis CI periodically try to build each subproject separately

liblit:
In theory, if Gradle dependencies are set up correctly, then it should be possible to build any subproject starting with a pristine tree. If subproject *X* depends on subproject *Y*, then Gradle should know that building *X* requires building *Y* first. The way to test this is to systematically build each subproject starting from a pristine WALA tree. Unfortunately, with 29 subprojects on 2 supported platforms, that can take a while. Travis CI runs up to five jobs at a time, and running this entire set of tests takes nearly 7 hours. So that’s not something we want to do on every push or on every pull request. However, this would be reasonable to do less frequently, such as once per week.

Fortunately, Travis CI has support for running periodic (e.g., weekly) tests: [cron jobs](https://docs.travis-ci.com/user/cron-jobs/). This pull request adds 29 × 2 single-subproject build tests that will only be enabled when run from a Travis CI cron job. If you think this is a good idea, then you should **do two things:**

1. Accept this pull request.
2. Follow the [Travis CI cron job instructions](https://docs.travis-ci.com/user/cron-jobs/) to add a weekly cron job using the following configuration:
    * Branch: master
    * Interval: weekly
    * Options: Always run

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,3 +38,255 @@ matrix:
     - os: osx
       env: BUILD_SYSTEM=gradle
       language: java
+    - os: linux
+      env: BUILD_SYSTEM=gradle BUILD_ONLY_SUBMODULE=com.ibm.wala.cast
+      language: java
+      jdk: oraclejdk8
+      if: type = cron
+    - os: linux
+      env: BUILD_SYSTEM=gradle BUILD_ONLY_SUBMODULE=com.ibm.wala.cast.java
+      language: java
+      jdk: oraclejdk8
+      if: type = cron
+    - os: linux
+      env: BUILD_SYSTEM=gradle BUILD_ONLY_SUBMODULE=com.ibm.wala.cast.java.ecj
+      language: java
+      jdk: oraclejdk8
+      if: type = cron
+    - os: linux
+      env: BUILD_SYSTEM=gradle BUILD_ONLY_SUBMODULE=com.ibm.wala.cast.java.test
+      language: java
+      jdk: oraclejdk8
+      if: type = cron
+    - os: linux
+      env: BUILD_SYSTEM=gradle BUILD_ONLY_SUBMODULE=com.ibm.wala.cast.java.test.data
+      language: java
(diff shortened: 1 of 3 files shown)


msridhar:
@liblit sorry just getting to look at this.  What is the motivation here?  Have you seen bugs like this before where building a particular sub-project doesn't work?  If so, what was the root cause?  I haven't seen this kind of problem before, but maybe we have more complex Gradle scripts here that can lead to problems.

liblit:
The root cause of the bugs this tries to prevent is failure to declare one or more dependencies, specifically when subproject _B_ depends on something built by subproject _A_. In the presence of such a bug, building all of WALA sequentially would succeed, assuming that _A_ is alphabetically before _B_. Building all of WALA in parallel would succeed or fail nondeterministically, depending on whether _A_ built the depended-upon component before or after _B_ tried to use it. Building just subproject _B_ would succeed if _A_ had already been built separately, or would fail if _A_ had not already been built. Changes to the depended-upon _A_ component would fail to cause transitive rebuilds (e.g., rerun of tests) in _B_, since Gradle had no idea that anything in _A_ affected anything in _B_.

During the long process of developing WALA's Gradle build scripts, several bugs of this kind did arise. The most common case was when _A_ built (e.g., downloaded) some data files that were needed as inputs to tests in _B_. After being bit by this several times, I finally decided to find and fix all such bugs. I created a simple shell script that iterated over each subproject _B_. For such subproject, my script checked out a completely fresh, clean WALA tree, then built and ran tests for _B_. This did indeed shake out several additional bugs, all due to missing cross-module dependencies.

Eventually I realized that Travis CI could do this for me, which led me to create this alternate Travis CI testing configuration. No new missing-cross-subproject-dependency bugs have arisen since then, but that's mainly because the Gradle scripts are fairly stable by now. Thus, I consider this form of testing overkill if done per-commit. But doing it once per week seems prudent.

msridhar:
Also scheduled the cron job

PR #317, opened 2018-06-04
Future Java compatibility fixes

liblit:
This began as an exploration of what it would take to get WALA running under Java 9 or 10. Neither of those is possible yet, but the attempt yielded several improvements. Highlights include:

- improved compatibility with future Java versions
  - replace constructors (0b9ac5a), methods (f06ddba), and command-line tools (5cf8b7e) that are deprecated or removed in Java 9 or 10
  - find native JVM shared library more reliably across different Java versions and operating systems (fbac524)
  - restructure a tricky expression to be compatible with Java 9+ generics type inference (989ef3e)
- improved testing
  - if Gradle is working offline, suppress a test that requires Internet access (32e105a)
  - add complete `smoke_main` native execution test, previously omitted from Gradle builds (6679262)


com.ibm.wala.util/src/com/ibm/wala/util/intset/IntSetUtil.java
```diff
@@ -31,7 +31,7 @@
       try {
         @SuppressWarnings("unchecked")
         Class<? extends MutableIntSetFactory<?>> intSetFactoryClass = (Class<? extends MutableIntSetFactory<?>>) Class.forName(System.getProperty(INT_SET_FACTORY_CONFIG_PROPERTY_NAME));
-        MutableIntSetFactory<?> intSetFactory = intSetFactoryClass.newInstance();
+        MutableIntSetFactory<?> intSetFactory = intSetFactoryClass.getDeclaredConstructor().newInstance();
```
msridhar: FYI I think I tried this change before but it caused problems for running WALA in a browser /cc @juliandolby

juliandolby: Manu is right.  That method broke TeaVM last time I tried it.

liblit: Indeed, @juliandolby did have to work around this TeaVM problem once before: 2c5d6ed15fa8313759f47a39a1e8a9bf43d0000c. I have reproduced this failure in the [TeaVM sandbox](http://teavm.org/sandbox/index.html) and [reported it to the TeaVM team](https://github.com/konsoletyper/teavm/issues/344) in my usual level of detail.

So hopefully TeaVM will fix this eventually. Until then, I will revise my pull request to exclude the `Class.getDeclaredConstructor()` changes. That is easily to do, as those changes were already encapsulated in their own commit.

msridhar: Is this an issue at execution time on TeaVM or classloading time?  If it's at execution time, we could just basically not use this feature on TeaVM and never execute the `getDeclaredConstructor()` call.  If it's a classloading-time issue then we're stuck.

liblit: I don’t know when TeaVM would trigger this; perhaps @juliandolby knows.

Either way, let’s keep in mind that the deprecated `Class.newInstance()` method is still available for use in [Java 9](https://docs.oracle.com/javase/9/docs/api/java/lang/Class.html#newInstance--) and even [Java 10](https://docs.oracle.com/javase/10/docs/api/java/lang/Class.html#newInstance--). And WALA is stuck at Java 1.8 for the time being anyway. So leaving these calls unchanged for now is perfectly reasonable. If `Class.newInstance()` goes away in Java 11 or later, the TeaVM maintainer (@konsoletyper) has offered [some suggested workarounds](https://github.com/konsoletyper/teavm/issues/344#issuecomment-394842552).

PR #332, opened 2018-07-19
Add install_name when build libcast on macos.

Leeleo3x:
Gradle won't pass absolute path when build libcast. We need to set install_name manually otherwise `dyld` would not able to find libcast at runtime.

This is only needed on macos since `ld` will look up all runtime search path automatically.


com.ibm.wala.cast/build.gradle
```diff
@@ -82,6 +82,9 @@ model {
 						.find { it.targetPlatform.name == targetPlatform.name }
 						.sharedLibraryLinkFile.parent
 					linker.args "-Wl,-rpath,$libDir"
+					if (targetPlatform.operatingSystem.name == "osx") {
```
liblit: It would be better to use a single-quoted string here (`'osx'`) instead of a double-quoted string (`"osx"`), since no `$` interpolation is needed.

liblit:
Thanks for that quoting fix! :+1:

msridhar:
Can we add a regression test for this?  We do have a Travis job that runs on Mac

Leeleo3x:
It won't change anything at the code level. It only tells `dyld` how to link this library. So I'm not sure if we can add tests for this.

Currently if you run `otool -L libcast.dylib` you would see:
```
./libcast.dylib:                                                                           
        @rpath/libcast.dylib (compatibility version 0.0.0, current version 0.0.0)          
        @rpath/libjvm.dylib (compatibility version 1.0.0, current version 1.0.0)           
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 400.9.0)     
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.50.4)
```

Before it was
```
./libcast.dylib:                                                                           
        libcast.dylib (compatibility version 0.0.0, current version 0.0.0)          
        @rpath/libjvm.dylib (compatibility version 1.0.0, current version 1.0.0)           
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 400.9.0)     
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.50.4)
```
and you have to copy paste libcast.dylib to the same folder as the executable so that the linker can find it.

And I also checked the maven build
```
./libcast.dylib:                                                                           
        /The/absolute/path/to//libcast.dylib (compatibility version 0.0.0, current version 0.0.0)          
        @rpath/libjvm.dylib (compatibility version 1.0.0, current version 1.0.0)           
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 400.9.0)     
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.50.4)
```

So it doesn't have the path issue as well.

liblit:
Surely something about this should be testable. If not, then why do we need the change? The WALA test suite, as run on macOS by Travis CI, currently passes all tests. So if this change is needed, then presumably WALA has some latent problem that the current macOS Travis CI tests do not cover. In other words, our current test suite is deficient, and therefore buggy. Let’s use this opportunity to fix that bug too.

There should be something we can try to do that will fail without this change, but will succeed with this change. What is that thing, and how to we arrange for the macOS Travis CI runs to try that thing automatically?

Leeleo3x:
Yes, that's true. There is one test that relates to this issue. And currently WALA solves the path issue by setting `DYLD_LIBRARY_PATH`

https://github.com/wala/WALA/blob/master/com.ibm.wala.cast.test/build.gradle#L45

Currently removing this environment variable will cause a runtime error

```
dyld: Library not loaded: libcast.dylib
```

After we adopt this patch, we don't need to set the environment variable anymore. Do you want me to remove https://github.com/wala/WALA/blob/master/com.ibm.wala.cast.test/build.gradle#L45 as well?

msridhar:
Yes, removing that line sounds good to me.

--Manu

liblit:
I added the logic to set `DYLD_LIBRARY_PATH`, because that fixed the macOS `dyld` failures I was seeing. However, I am absolutely not a macOS expert. If @Leeleo3x thinks that setting the embedded `install_name` is more appropriate, then I am happy to go along. Furthermore, it is generally better for a library to work properly using only the information embedded therein, rather than requiring that some obscure environment variable be set. So yes, all in all this seems like an improvement. Thanks, @Leeleo3x!

Leeleo3x:
You are welcome.

msridhar:
Thanks @Leeleo3x!  Merged.

PR #335, opened 2018-07-24
Avoid Exceptions with Phantom Superclasses

subarnob:
@msridhar /cc @lazaroclapp
Building the CFG fails with `NoSuperclassFoundException` when a callee method has a _phantom_ superclass.
Continue by returning `null` instead.

### DIFF
--- a/com.ibm.wala.core/src/com/ibm/wala/classLoader/BytecodeClass.java
+++ b/com.ibm.wala.core/src/com/ibm/wala/classLoader/BytecodeClass.java
@@ -296,7 +296,8 @@ public abstract class BytecodeClass<T extends IClassLoader> implements IClass {
       computeSuperclass();
     }
     if (superClass == null && !getReference().equals(TypeReference.JavaLangObject)) {
-      throw new NoSuperclassFoundException("No superclass found for " + this + " Superclass name " + superName);
+      //throw new NoSuperclassFoundException("No superclass found for " + this + " Superclass name " + superName);
+      return null;
     }
     return superClass;
   }
--- a/com.ibm.wala.core/src/com/ibm/wala/classLoader/PhantomClass.java
+++ b/com.ibm.wala.core/src/com/ibm/wala/classLoader/PhantomClass.java
@@ -53,7 +53,8 @@ public class PhantomClass extends SyntheticClass {
 
   @Override
   public IMethod getMethod(Selector selector) {
-    throw new UnsupportedOperationException();
+    //throw new UnsupportedOperationException();
+    return null;
   }
 
   @Override


com.ibm.wala.core/src/com/ibm/wala/classLoader/BytecodeClass.java
```diff
@@ -296,7 +296,8 @@ public IClass getSuperclass() {
       computeSuperclass();
     }
     if (superClass == null && !getReference().equals(TypeReference.JavaLangObject)) {
-      throw new NoSuperclassFoundException("No superclass found for " + this + " Superclass name " + superName);
+      //throw new NoSuperclassFoundException("No superclass found for " + this + " Superclass name " + superName);
```
msridhar: Why do we need this change?  What fails if we continue to throw the exception?  I'm hesitant about his change as this code has been throwing an exception for quite a while.

subarnob: It throws `NoSuperclassFoundException` while making CFG when the callee is in a `BytecodeClass` but has a `PhantomClass` superclass, as follows:

`com.ibm.wala.classLoader.NoSuperclassFoundException: No superclass found for <Application,Landroid/support/v4/view/AsyncLayoutInflater$BasicInflater> Superclass name Landroid/view/LayoutInflater
 at com.ibm.wala.classLoader.BytecodeClass.getSuperclass(BytecodeClass.java:299)
 at com.ibm.wala.classLoader.BytecodeClass.getMethod(BytecodeClass.java:464)
 at com.ibm.wala.classLoader.JavaLanguage.inferInvokeExceptions(JavaLanguage.java:585)
 at com.ibm.wala.cfg.ShrikeCFG$BasicBlock.addExceptionalEdges(ShrikeCFG.java:287)
 at com.ibm.wala.cfg.ShrikeCFG$BasicBlock.computeOutgoingEdges(ShrikeCFG.java:243)
 at com.ibm.wala.cfg.ShrikeCFG$BasicBlock.access$0(ShrikeCFG.java:232)
 at com.ibm.wala.cfg.ShrikeCFG.computeEdges(ShrikeCFG.java:133)
 at com.ibm.wala.cfg.ShrikeCFG.<init>(ShrikeCFG.java:76)
 at com.ibm.wala.cfg.ShrikeCFG.make(ShrikeCFG.java:63)
 at com.ibm.wala.classLoader.ShrikeIRFactory.makeCFG(ShrikeIRFactory.java:43)
 at com.ibm.wala.classLoader.ShrikeIRFactory.makeIR(ShrikeIRFactory.java:58)
 at com.ibm.wala.ssa.DefaultIRFactory.makeIR(DefaultIRFactory.java:69)`

msridhar: Ah ok; ugh.  Thanks for pointing this out.  I'd like to think a bit more about the right fix here.  It's possible I should just tweak the phantom class hacky implementation to work properly.

reddr:
During testing WALA master, I also stumpled across this issue.
It suffices to call _IClass.isSuperClass()_ on a class that has a PhantomClass as a super class.

This triggers _BytecodeClass.computeSuperClass()_ which executes this line in the else branch:
`this.superClass = this.loader.lookupClass(TypeName.findOrCreate(this.superName));`
_loader.lookupClass_ fails, returns null and subsequently triggers the NoSuperClassFoundException.

Without checking for possible side-effects, a solution could be that _ClassLoaderImpl.lookupClass_
returns a PhantomClass instead of null, when _ClassHierarchyFactory.makeWithPhantom_ was used.

msridhar:
@reddr I think my approach to the whole phantom superclass thing was kind of half-baked.  I suggest you not rely on it for now and try to get the relevant superclass into your `AnalysisScope`.  My thinking is when I have time I should revamp the whole way phantom superclasses are implemented to be more robust.

reddr:
@msridhar Unfortunately, this seems to be last puzzle piece for my tool to be adapted to WALA master. I'm automatically analyzing apps where it might happen that, in particular, dead code classes contain unresolvable superclasses (i.e., the superclass is part of a library that is not statically linked).
Still, I have to keep the class as I extract code fingerprints.

When I first stumpled across this issue, I patched it in my custom WALA version in a way that
unresolvable superclasses were simply replaced with the class hierarchy root. This worked flawlessly, but as a side-effect, you might loose some information (e.g., name of superclass(es)).

Such an option would be a quick fix to this problem, while if you want to keep references, the PhantomClass concept could be another option. Just let me know if such a dual approach could be interesting for you or whether you just want to have a single clean solution. In the meantime, I still have to use my custom version to resolve this issue.

msridhar:
@reddr I need to think more here.  Your use of class hierarchy root makes sense, and I thought about it at first but for some reason concluded it was risky.  Now I've forgotten why :-)  Let me think more and then we can figure out how to proceed.

msridhar:
Closing as out-of-date.  I think we have better phantom superclass handling on master now; see #377

PR #353, opened 2018-09-13
Added new DexFileModule constructor to support specific dex files inside apks.

sifislag:
We need this in order to support apks with multiple dex files inside, which appear quite often. To give you an example of our usage of it:

````java
private static void addAPKtoScope(ClassLoaderReference loader, AnalysisScope scope, String fileName){
        File apkFile = new File(fileName);
        MultiDexContainer<? extends DexBackedDexFile> multiDex = null;
        try {
            multiDex = DexFileFactory.loadDexContainer(apkFile, Opcodes.forApi(24));
        } catch (IOException e) {
            System.err.println("Failed to open " +fileName + " as multidex container.\n" + e);
        }

        try{
            for (String dexEntry : multiDex.getDexEntryNames()) {
                System.out.println("Adding dex file: " +dexEntry + " of file:" + fileName);
                scope.addToScope(loader, new DexFileModule(apkFile, dexEntry,24));
            }
        }catch (IOException e){
            System.err.println("Failed to open " +fileName + " as multidex container.\n" + e);
        }
    }
````


sifislag:
@msridhar I only added a new constructor without making any changes to the preexisting code of the class so I don't think it could affect any code using it.

msridhar:
> @msridhar I only added a new constructor without making any changes to the preexisting code of the class so I don't think it could affect any code using it.

Sure but how will we know if a future change breaks the support for APKs with multiple dex files?  A regression test will ensure no one breaks this in the future.

sifislag:
> Sure but how will we know if a future change breaks the support for APKs with multiple dex files? A regression test will ensure no one breaks this in the future.

I'll look into it in the following days. Do you have any tests in mind I could check/base it on?

sifislag:
I added a test. It does not check an apk with more than one .dex file because I could not find any in the dalvik.test project or in  DroidBench. I picked a random apk from DroidBench and create a scope and class hierarchy using my new constructor, and a scope and class hierarchy using the preexisting API and check that the 2 class hierarchies have the same number of Application classes. Will this do?

msridhar:
@sifislag that sounds good, but could we create a multi-dex apk by hand?  If not, this is good enough.

@juliandolby could you look over this code and merge?  I'm not fully familiar with the Android code.

sifislag:
I hand built a small(5 classes) apk with 2 .dex files and added it in /data/. Using this I made a new test that checks the new function call with this apk.

msridhar:
Awesome!  Merged.

sifislag:
Thanks!

Issue #356, opened 2018-09-18
Add implementation of IMethod/IClass isSynthetic


reddr:
I'm maintaining [LibScout](https://github.com/reddr/LibScout) a library detector that identifies Java libraries in Java/Android apps. For the detection I need to extract properties both from compiled java (class) and Android (dex) code. To account for subtle code generation differences between javac/dx/d8 compilers, I filter synthetic methods in a pre-processing step.

Current WALA master does not implement IClass.isSynthetic and returns false for IMethod.isSynthetic(). 
In the quite dated, custom WALA version that my project uses I implemented the missing functions. However, I found that a proper implementation has some side-effects as the function call  IMethod.isSynthetic() is already used at a number of different code locations but apparently has never been tested for a return value other than false. 

Examples (grep over source):
`
com/ibm/wala/classLoader/CodeScanner.java:    if (m.isSynthetic()) {
com/ibm/wala/ssa/DefaultIRFactory.java:    if (method.isSynthetic()) {
com/ibm/wala/analysis/reflection/FactoryContextSelector.java:    if (callee.isSynthetic()) {
com/ibm/wala/analysis/reflection/FactoryBypassInterpreter.java:    if (node.getMethod().isSynthetic()) {
com/ibm/wala/demandpa/util/SimpleMemoryAccessMap.java:    if (ALWAYS_BUILD_IR || n.getMethod().isSynthetic()) {
com/ibm/wala/ipa/callgraph/impl/ExplicitCallGraph.java:      if (getMethod().isSynthetic()) {
com/ibm/wala/ipa/callgraph/impl/ExplicitCallGraph.java:      if (getMethod().isSynthetic()) {
com/ibm/wala/ipa/callgraph/impl/BasicCallGraph.java:      if (method != null && !method.isSynthetic() && method.isAbstract()) {
`

Due to these side-effects and for time reasons, I decided to add IMethod.isMethodSynthetic that does the job and is not called from other locations. However, as I want to adopt WALA master, this is no longer an option. 

The implementation for isSynthetic is straight-forward (I'm going to add pull requests), however, it should probably be tested or re-checked on the code locations shown above.

juliandolby:
A pull request would be very helpful.  We would be able to see what you need to change, and also Travis CI would run all of our tests to make sure everything still works.

reddr:
Here it is: https://github.com/wala/WALA/pull/357

PR #357, opened 2018-09-18
Implementation of isSynthetic for IClass and IMethods

reddr:
This addresses https://github.com/wala/WALA/issues/356.
Somebody with knowledge in the cast implementation should double-check the respective parts.

### DIFF
--- a/com.ibm.wala.cast/source/java/com/ibm/wala/cast/loader/AstClass.java
+++ b/com.ibm.wala.cast/source/java/com/ibm/wala/cast/loader/AstClass.java
@@ -86,6 +86,11 @@ abstract public class AstClass implements IClass, ClassConstants {
   }
 
   @Override
+  public boolean isSynthetic() {
+    return false;
+  }
+
+  @Override
   public int getModifiers() {
     return modifiers;
   }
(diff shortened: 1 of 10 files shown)


juliandolby:
Looking over the CAst changes, they make sense to me.  It looks like AstMethod already checked modifiers for isSynthetic, and it makes that classes from the AST are never synthetic.

The other changes also look good to me.

I am happy with this change, assuming the tests all pass.

juliandolby:
Based on the tests, it looks like there are places in the code where isSynthetic is used as a proxy for a SyntheticMethod type check.  We'll need to understand what the right fix is.

msridhar:
Yes, the documentation is bad, but `isSynthetic()` in WALA is *not* referring to synthetic methods generated by javac.  It's about methods for which WALA uses a custom model and not bytecode.  As suggested in #356 I think a new method with a different name (and documentation!) is a better solution.

reddr:
@msridhar Ok, this makes sense. I see two options for the renaming:

1. The proper way would probably be keeping isSynthetic for compiler-generated methods (following isAbstract, isPublic, ..) and renaming the current isSynthetic to something like isWalaSynthetic. This would require slightly more effort (updating code references) and it might break existing clients relying on these functions (if any).

2. Keep isSynthetic for WALA-generated methods (adding documentation!) and using something like isMethodSynthetic to refer to compiler-generated code.

For consistency, I would tend to 1. but I'm open to suggestions.

msridhar:
I'm in favor of 1.  I don't think it should break too many existing clients.  If we really want to be careful, we can cut a release before landing this change and bump the version number appropriately (but we've been quite sloppy about releases / versioning thus far so I assume there are no great expectations around this).

PR #359, opened 2018-09-21
Impl of IMethod.is(Wala)Synthetic and IClass.isSynthetic

reddr:
So far IMethod.isSynthetic referred to WALA-generated helper functions
and there was no equivalent to check whether an IMethod is synthetic in
terms of compiler-generated.
- To make naming consistent this patch first renames the isSynthetic to
isWalaSynthetic to clearly indicate that a given IMethod was generated
by WALA. 

- Then, we re-introduce isSynthetic that from now on checks
whether an IMethod is synthetic/compiler-generated (referring to the
synthetic flag in bytecode).

- Complementary to IMethod.isSynthetic, IClass.isSynthetic checks whether
an IClass is compiler-generated.


com.ibm.wala.core/src/com/ibm/wala/classLoader/IMethod.java
```diff
@@ -43,7 +43,14 @@
   boolean isNative();
 
   /**
-   * Did someone synthesize this method? (As opposed to reading it from a class file)
+   * Is this method created by WALA?
```
msridhar: Can we rephrase this "Is the implementation of this method a model generated by WALA?"  I think in many cases the method exists in the original bytecode, we just hijack the implementation.  Saying the method is created by WALA is a bit ambiguous.

msridhar:
Thanks for the contribution @reddr!

PR #374, opened 2018-11-19
Miscellaneous build and IDE updates

liblit:
These changes are a grab-bag of minor fixes, updates, and improvements to the Gradle build system, IDE configurations, and integrations between the two. See individual commits and commit messages for more detail. Some of the more interesting commits include:

- 72f95230f86f2544c3ba46f0c80b3cfcabf63da3: Remove `testJar` tasks; publish existing `jarTest` tasks instead. The only important difference is that `testJar` created archives named `*-tests.jar` while the existing `jarTest` creates archives named `*-test.jar`. Other than that small change, the artifacts published by each seem to be identical.

- e37fd696ccef103b2a9af5d5e9d63c126cb7e901: Remove private methods that are never locally used.

- 310f36163a48c2ef130b9d3e6c96d71dbf3f88fe: Upgrade to Gradle 4.10. No changes to our build scripts as a result of this upgrade.  However, incremental Java compilation is now [enabled by default](https://docs.gradle.org/4.10/release-notes.html#incremental-java-compilation-by-default), which might be nice for some folks.

- 091d2fe3df012aa3d1e7f87efd25d0ea62a00d63: Stop using Travis CI containers (`sudo: false`), as [they are being deprecated](https://blog.travis-ci.com/2018-10-04-combining-linux-infrastructures) and [will soon stop working entirely](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).


msridhar:
Thanks!

PR #377, opened 2018-12-11
Introduce ClassHierarchyFactory.makeWithRoot

reddr:
Build ClassHierarchy and replace any missing superclass
with the hierarchy root. Adds necessary changes to co-exist
with makeWithPhantom.

### DIFF
--- a/com.ibm.wala.core.tests/src/com/ibm/wala/core/tests/cha/MissingSuperTest.java
+++ b/com.ibm.wala.core.tests/src/com/ibm/wala/core/tests/cha/MissingSuperTest.java
@@ -46,14 +46,24 @@ public class MissingSuperTest extends WalaTestCase {
     AnalysisScope scope = AnalysisScopeReader.readJavaScope(TestConstants.WALA_TESTDATA,
         (new FileProvider()).getFile("J2SEClassHierarchyExclusions.txt"),
         MissingSuperTest.class.getClassLoader());
+
     TypeReference ref = TypeReference.findOrCreate(ClassLoaderReference.Application,
         "Lmissingsuper/MissingSuper");
+
     // without phantom classes, won't be able to resolve
     ClassHierarchy cha = ClassHierarchyFactory.make(scope);
     Assert.assertNull("lookup should not work", cha.lookupClass(ref));
+
+    // with makeWithRoot lookup should succeed and
+    // unresolvable super class "Super" should be replaced by hierarchy root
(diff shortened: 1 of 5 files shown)


reddr:
Some additional note regarding the superclass issue in #335:
I stumbled across the same issue again with makeWithRoot. Apparently, ClassHierarchy.addClass
computes and checks superclasses but does *not* set the inferred superclass. This is deferred
until the first call of IClass.getSuperClass (for performance reasons?). 

There, I explicitly set the superclass to the hierarchy root, but this needs to be done for the PhantomClass mode as well. Probably, this already fixes the issues with makeWithPhantom.

msridhar:
Sorry for my slow response here.  This change looks great!  But I don't see any tests.  Can we enhance [this test](https://github.com/wala/WALA/blob/master/com.ibm.wala.core.tests/src/com/ibm/wala/core/tests/cha/MissingSuperTest.java) to cover the `ROOT` case as well?

msridhar:
> Hmm only the maven build fails for some (unrelated) reason. Any idea?

I kicked the build again, hopefully it'll pass.

> Any thoughts on my second comment, i.e. why addClass computes the superclass but does not set it?
> What is the rationale behind not setting it right away but deferring it to the first call of getSuperClass? Can't be lazy computation as it is already computed in addClass.
> 
> To me, this looks like a construct that may easily lead to new issues, e.g. the (incomplete) phantom patch seems to miss setting the superclass there.

I don't know immediately the rationale for this, but the `IClass` interface doesn't have a way to set a superclass; that is delegated to the implementations.  So setting it in `ClassHierarchy.addClass` is not straightforward.  We could try updating `IClass` with a `setSuperclass` method but that will affect languages beyond Java and may require some careful thought.

msridhar:
@reddr these changes look good to me but I'm unsure if this PR is still incomplete due to the `addClass()` issue.  Let me know.

reddr:
@msridhar I think this one is good to go. Was unsure about why computing and setting the superclass is located in different methods and whether there is an easy way to fix this. I still think that this is the problem with the current phantom patch, I left a comment in the code. But that will be a separate PR.

msridhar:
> @msridhar I think this one is good to go. Was unsure about why computing and setting the superclass is located in different methods and whether there is an easy way to fix this. I still think that this is the problem with the current phantom patch, I left a comment in the code. But that will be a separate PR.

Ok, sounds good.  If you have time could you open an issue regarding the specific problem with phantom classes?

PR #380, opened 2018-12-24
Minor container API improvements

liblit:
Miscellaneous improvements under the general theme of how WALA uses standard Java containers/collections. Each commit focuses on one or several instances of a single improvement, so code review is probably best done for one commit at a time.

### DIFF
--- a/com.ibm.wala.cast/source/java/com/ibm/wala/cast/loader/CAstAbstractLoader.java
+++ b/com.ibm.wala.cast/source/java/com/ibm/wala/cast/loader/CAstAbstractLoader.java
@@ -26,10 +26,7 @@ import com.ibm.wala.classLoader.IMethod;
 import com.ibm.wala.classLoader.ModuleEntry;
 import com.ibm.wala.ipa.cha.IClassHierarchy;
 import com.ibm.wala.types.TypeName;
-import com.ibm.wala.util.collections.FilterIterator;
 import com.ibm.wala.util.collections.HashMapFactory;
-import com.ibm.wala.util.collections.HashSetFactory;
-import com.ibm.wala.util.collections.MapIterator;
 import com.ibm.wala.util.strings.Atom;
 import com.ibm.wala.util.warnings.Warning;
 
@@ -62,31 +59,24 @@ public abstract class CAstAbstractLoader implements IClassLoader {
     this(cha, null);
   }
 
+  private Set<Warning> messagesFor(ModuleEntry module) {
+    return errors.computeIfAbsent(module, (key) -> new HashSet<>());
+  }
+
   public void addMessage(ModuleEntry module, Set<Warning> message) {
-    if (! errors.containsKey(module)) {
-      errors.put(module, new HashSet<Warning>());
-    }
-    
-    errors.get(module).addAll(message);
+    messagesFor(module).addAll(message);
   }
 
   public void addMessage(ModuleEntry module, Warning message) {
-    if (! errors.containsKey(module)) {
-      errors.put(module, new HashSet<Warning>());
-    }
-    
-    errors.get(module).add(message);
+    messagesFor(module).add(message);
   }
 
   private Iterator<ModuleEntry> getMessages(final byte severity) {
-    return new MapIterator<>(new FilterIterator<>(errors.entrySet().iterator(), o -> {
-       for(Warning w : o.getValue()) {
-         if (w.getLevel() == severity) {
-           return true;
-         }
-       }
-       return false;
-    }), Map.Entry::getKey);
+    return errors.entrySet().stream()
+            .filter(entry -> entry.getValue().stream()
+                    .anyMatch(w -> w.getLevel() == severity))
+            .map(Map.Entry::getKey)
+            .iterator();
   }
   
   public Iterator<ModuleEntry> getModulesWithParseErrors() {
@@ -133,12 +123,9 @@ public abstract class CAstAbstractLoader implements IClassLoader {
 
   @Override
   public int getNumberOfMethods() {
-    int i = 0;
-    for (IClass cls : types.values()) {
-      i += cls.getDeclaredMethods().size();
-    }
-
-    return i;
+    return types.values().stream()
+            .mapToInt(cls -> cls.getDeclaredMethods().size())
+            .sum();
   }
 
   @Override
@@ -184,17 +171,7 @@ public abstract class CAstAbstractLoader implements IClassLoader {
 
   @Override
   public void removeAll(Collection<IClass> toRemove) {
-    Set<TypeName> keys = HashSetFactory.make();
-
-    for (Map.Entry<TypeName, IClass> E : types.entrySet()) {
-      if (toRemove.contains(E.getValue())) {
-        keys.add(E.getKey());
-      }
-    }
-
-    for (TypeName key : keys) {
-      types.remove(key);
-    }
+    types.values().removeIf(toRemove::contains);
   }
 
 }
--- a/com.ibm.wala.cast/source/java/com/ibm/wala/cast/loader/CAstAbstractLoader.java
+++ b/com.ibm.wala.cast/source/java/com/ibm/wala/cast/loader/CAstAbstractLoader.java
@@ -63,7 +63,7 @@ public abstract class CAstAbstractLoader implements IClassLoader {
     return errors.computeIfAbsent(module, (key) -> new HashSet<>());
   }
 
-  public void addMessage(ModuleEntry module, Set<Warning> message) {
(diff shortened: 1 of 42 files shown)


msridhar:
@liblit are you making these changes guided by some tool?  Let me know when the PR is ready for review.

com.ibm.wala.cast/source/java/com/ibm/wala/cast/loader/CAstAbstractLoader.java
```diff
@@ -62,31 +59,24 @@ public CAstAbstractLoader(IClassHierarchy cha) {
     this(cha, null);
   }
 
+  private Set<Warning> messagesFor(ModuleEntry module) {
+    return errors.computeIfAbsent(module, (key) -> new HashSet<>());
+  }
+
   public void addMessage(ModuleEntry module, Set<Warning> message) {
```
msridhar: Rename to `addMessages`?  (Since we're modifying this code anyway)

com.ibm.wala.cast/source/java/com/ibm/wala/cast/loader/CAstAbstractLoader.java
```diff
@@ -62,31 +59,24 @@ public CAstAbstractLoader(IClassHierarchy cha) {
...
   private Iterator<ModuleEntry> getMessages(final byte severity) {
-    return new MapIterator<>(new FilterIterator<>(errors.entrySet().iterator(), o -> {
-       for(Warning w : o.getValue()) {
-         if (w.getLevel() == severity) {
-           return true;
-         }
-       }
-       return false;
-    }), Map.Entry::getKey);
+    return errors.entrySet().stream()
```
msridhar: Would it make more sense to just stream the keys and do a `Map.get()` operation in the filter, rather than streaming the entries and then mapping back to the key?

com.ibm.wala.core/src/com/ibm/wala/cfg/exc/inter/InterprocNullPointerAnalysis.java
```diff
@@ -11,11 +11,8 @@
 
 package com.ibm.wala.cfg.exc.inter;
 
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Map;
+import java.util.*;
```
msridhar: Not a fan of star imports 😄 If / when I get Error Prone working I'll make this fail at build time

com.ibm.wala.core/src/com/ibm/wala/classLoader/ClassLoaderImpl.java
```diff
@@ -687,19 +673,11 @@ public Reader getSource(IClass klass) {
...
-      throw new IllegalArgumentException("toRemove is null");
-    }
-    for (IClass klass : toRemove) {
-      if (DEBUG_LEVEL > 0) {
-        System.err.println("removing " + klass.getName());
-      }
-      loadedClasses.remove(klass.getName());
-      sourceMap.remove(klass.getName());
-    }
+    toRemove.stream().map(IClass::getName)
```
msridhar: Retain the null check on the argument

com.ibm.wala.core/src/com/ibm/wala/ssa/DefUse.java
```diff
@@ -187,4 +187,14 @@ public void remove() {
...
     return uses[v] == null ? 0 : uses[v].size();
   }
+
+  /**
+   * @param v a value number
+   * @return true if the variable with the given value number has no uses
+   */
+  public boolean isUnused(int v) {
+    final MutableIntSet usesSet = uses[v];
+    assert usesSet == null || !usesSet.isEmpty();
```
msridhar: Seems like an odd place for this assertion?  If we care we should assert at the point where we construct the `DefUse`, for all value numbers

com.ibm.wala.scandroid/source/org/scandroid/flow/InflowAnalysis.java
```diff
@@ -54,13 +54,7 @@
...
 //import static util.MyLogger.LogLevel.DEBUG;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
msridhar: Star import

com.ibm.wala.scandroid/source/org/scandroid/prefixtransfer/UriPrefixTransferGraph.java
```diff
@@ -48,15 +48,8 @@
...
 package org.scandroid.prefixtransfer;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
msridhar: Star import

com.ibm.wala.cast/source/java/com/ibm/wala/cast/loader/CAstAbstractLoader.java
liblit: OK. I will take care of this in a group of updates for this pull request.

com.ibm.wala.cast/source/java/com/ibm/wala/cast/loader/CAstAbstractLoader.java
liblit: For every key, we will eventually need the corresponding value for use during the `filter` stage of this code. If we only stream the keys, then we must use `Map.get()` to fetch each corresponding value. We end up doing an sweep through the map keys plus _k_ look-ups for a _k_-entry map.

In my proposed change, we are streaming the entries rather than the keys, so each item in the stream is already a (key, value) pair. What you describe as “mapping back to the key” is really just extracting the first component of such a pair. That’s going to be quite efficient, as it does not require an additional map lookup.

I am marking this as resolved, but please reopen and clarify if you still think that there is a good reason to stream keys instead of (key, value) pairs.

com.ibm.wala.core/src/com/ibm/wala/cfg/exc/inter/InterprocNullPointerAnalysis.java
liblit: I agree. My IDE was trying to be helpful behind my back. I have reconfigured my IDE and will correct this in a forthcoming group of updates for this pull request.

com.ibm.wala.scandroid/source/org/scandroid/flow/InflowAnalysis.java
liblit: I agree. My IDE was trying to be helpful behind my back. I have reconfigured my IDE and will correct this in a forthcoming group of updates for this pull request.

com.ibm.wala.scandroid/source/org/scandroid/prefixtransfer/UriPrefixTransferGraph.java
liblit: I agree. My IDE was trying to be helpful behind my back. I have reconfigured my IDE and will correct this in a forthcoming group of updates for this pull request.

com.ibm.wala.core/src/com/ibm/wala/ssa/DefUse.java
liblit: In a build with assertions off, the code in this method simplifies to `return uses[v] == null`. Quite efficient, but wrong if `uses[v]` were present but empty. I put this assertion here because here is where I rely on it being true.

I am happy to broaden the check to all value numbers and move it to the end of the `DefUse` constructor. I’m equally happy to leave the check here or to remove the check entirely. Please let me know which you like best and I will make it so.

com.ibm.wala.core/src/com/ibm/wala/classLoader/ClassLoaderImpl.java
liblit: Oops, good catch. I restore the `null` check this in a forthcoming group of updates for this pull request. Thank you for reviewing these proposed changes so attentively!

liblit:
I appreciate your concerns about both test coverage and stream performance. Perhaps some parts of this pull request are safer than others. If you like, I could split this pull request into smaller pieces, so that you can easily approve the ones you feel confident about while rejecting the rest.

I grouped these commits into a single pull request because they are thematically related, but the commits are _not_ interdependent, and this pull request need not be an all-or-nothing affair.

PR #392, opened 2019-01-16
Fixes #373  (Cast error)

### LINKED ISSUE #373: com.ibm.wala.dalvik.dex.instructions.UnaryOperation cannot be cast to org.jf.dexlib2.iface.instruction.TwoRegisterInstruction
In com.ibm.wala.dalvik.classLoader.DexIMethod.findOutArrayElementType(), WALA's instruction cannot be cast to org.jf.dexlib2.iface.instruction
image

reddr:
I recently faced the same issue. This PR fixes the problem.

### DIFF
--- a/com.ibm.wala.dalvik/src/com/ibm/wala/dalvik/classLoader/DexIMethod.java
+++ b/com.ibm.wala.dalvik/src/com/ibm/wala/dalvik/classLoader/DexIMethod.java
@@ -3194,12 +3194,11 @@ public class DexIMethod implements IBytecodeMethod<Instruction> {
 					return newArray.newSiteRef.getDeclaredType().getArrayElementType();
 				}
 			} else if (curInst.getOpcode() == Opcode.MOVE_OBJECT || curInst.getOpcode() == Opcode.MOVE_OBJECT_16 || curInst.getOpcode() == Opcode.MOVE_OBJECT_FROM16) {
-				TwoRegisterInstruction tri = (TwoRegisterInstruction) curInst;
-				int regA = tri.getRegisterA();
-				int regB = tri.getRegisterB();
-				if (regA == interestingRegister) {
-					interestingRegister = regB;
-				}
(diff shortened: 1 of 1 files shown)


msridhar:
Can we get a unit test?

reddr:
Adding a specific testcase wouldn't be an issue. But since testing for dalvik parsing is missing completely (as far as I see it), this would first require setting up some basic "infrastructure" for testing dalvik parsing in general, i.e. a dex file that, in best case, contains all different instructions or java code samples that are compiled to dex for testing (I think that's what exists for the java bytecode), or maybe even re-use the java examples for java bytecode or the droidbench apps for testing.

Setting this up properly seems to be some effort and this seems to be preferable over a single parsing testcase that might not be reusable for testing the other instructions. IMHO this should even be a separate issue.

msridhar:
I agree that separating out testing of the dex parsing functionality can be a separate issue.  I see two options:

1. Piggy-back on some other tests that exercise the dex parsing code.  E.g., add some kind of Android call graph test that would fail without this fix?

2. Punt on the test for now but open a follow-up issue and try to address it soon.

@reddr WDYT?

msridhar:
Sorry for being picky about this, I just want to understand the situation.  @reddr is the current situation that there is literally no unit test that exercises the Dex parsing code?  Or do some tests exercise that code but they also exercise call graph construction, etc.?  If it's the latter, can we just add one test that uses the existing mechanisms?  I just really want to avoid adding new code without tests (and I know existing code hasn't been held to this standard).

reddr:
As far as I see it the vast majority of dalvik tests are CG generation tests and checks for uncalled methods etc. These will probably test parts of the dex parsing code. But, rather by chance than systematic (this part of the code was most probably never tested, as it crashes immediately).

Given this situation, I still think that doing this systematically will require one or more (small) dex files that contain code that triggers these parsing functions. Maybe some of the droidbench apks already have it, but checking this would require some manual investigation.

PR #408, opened 2019-02-02
Expose AST node children as overt List rather than covert array

liblit:
Previously access to AST children was provided via `getChildCount()` and `getChild(int)` methods.  These effectively treated the children as some sort of array-like sequence, but with none of the API conveniences that a proper Java array or `Collection` offers.  Exposing a node’s children as a standard `List` makes it easy for us to write loops over children, create `Iterator`s over children, use bulk operations like `Collections.addAll`, etc.

Most such `List`s of children will actually still be arrays, wrapped up as `ArrayList`. One might fear that this extra level of encapsulation would harm performance. However, I found no consistent performance change across 20 trials of the `com.ibm.wala.core.tests.callGraph*` tests.


msridhar:
LGTM.  I wouldn't expect this change to affect performance too much, as most performance-critical stuff operates on IRs, not ASTs.

@juliandolby can you take a look?  Would be nice to get your stamp as you wrote most of this code.

liblit:
I am going to be bold and assume implicit approval from @juliandolby. If something breaks, we can always revert.

Issue #409, opened 2019-02-03
Proposal: Google Java Format all the things


msridhar:
I've found some recent code reviews to be hard to read due to inconsistent source code formatting in the repo.  I propose we use [Google Java Format](https://github.com/google/google-java-format) (GJF) to get rid of this problem for Java code.  The basic idea would be:

* Leverage the [GJF Gradle plugin](https://github.com/sherter/google-java-format-gradle-plugin) to run GJF and verify that code is formatted.
* Modify `gradlew` to install a Git pre-commit hook that formats code at commit time.  Once this is run once, auto-formatting should be automatic for developers.
* On CI, run `./gradlew verGJF` to ensure code is formatted correctly.
* Initially have one massive commit that reformats the whole repo.  (Anyone with outstanding changes will have to do a slightly tricky rebase, so this would have to be announced / coordinated properly.)

This is the setup we use with NullAway.  There are other setups like [pre-commit](https://pre-commit.com) that don't require hacking `gradlew` to install a pre-commit hook, but they require installation of a separate tool.

@liblit @juliandolby thoughts?  If we're on the same page I can put this together.

liblit:
Overall: :+1:.

I have only weak opinions about code formats, but a _strong_ opinion that whatever format is used must have good tool integration. GJF is available as a command-line tool, a Gradle plugin, an IntelliJ IDEA plugin, an Eclipse plugin, and more. That makes it a fine choice in my book.

> Modify `gradlew` to install a Git pre-commit hook that formats code at commit time. Once this is run once, auto-formatting should be automatic for developers.

I would expect that any modifications to `gradlew` will be overwritten the next time we upgrade Gradle. Remembering to restore our tweaks seems like a hassle. @msridhar, apparently you have found this to be reasonable in practice on other projects?

Why not install Git hooks under Gradle control as a task that can be run explicitly, or run implicitly as a dependency of other tasks? [This gist](https://gist.github.com/KenVanHoeylandt/c7a928426bce83ffab400ab1fd99054a) shows a nice way to do it by creating a Gradle task to do the work and having the standard `build` task depend on it. [This Gradle plugin](https://github.com/STAR-ZERO/gradle-githook) also looks helpful.

> On CI, run `./gradlew verGJF` to ensure code is formatted correctly.

Agreed. It would also be easy to hook that up so that it runs as part of `./gradlew build` along with all of our other tests. Literally just a one-line change to add `build.dependsOn verGJF` to the top-level `build.gradle`.

[...]

PR #410, opened 2019-02-03
Add scripting for Google Java Format

msridhar:
WIP to address #409, for discussion


build.gradle
```diff
@@ -10,8 +10,13 @@ plugins {
 	id 'nebula.lint' version '8.3.1'
 	id 'nebula.source-jar' version '7.0.1' apply false
 	id 'maven-publish'
+        id 'com.github.sherter.google-java-format' version '0.8'
```
liblit: Keep this plugin list sorted alphabetically.

build.gradle
```diff
@@ -10,8 +10,13 @@ plugins {
...
 }
 
+repositories {
+  // to get the google-java-format jar and dependencies
+  mavenCentral()
```
liblit: If adding `mavenCentral()` repository here, then no need to also add it later for just the Java subprojects. Look for and remove the now-redundant `mavenCentral()` that appears [later in this same file](#diff-c197962302397baf3a4cc36463dce5eaR78).

config/hooks/pre-commit
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+set -e
+
+REPO_ROOT_DIR="$(git rev-parse --show-toplevel)"
+
+files=$((git diff --cached --name-only --diff-filter=ACMR | grep -Ei "\.java$") || true)
```
liblit: Shells disambiguate `$((` differently or not at all. Add space, replacing `$((` with `$( (`.

Caught by `shellcheck` linter.

config/hooks/pre-commit
```diff
@@ -0,0 +1,12 @@
...
+if [ ! -z "${files}" ]; then
```
liblit: Simplify by using `-n` instead of `! -z`.

Caught by `shellcheck` linter.

liblit: Simplify by  using `"$files"` instead of `"${files}"`. No need for curly braces here.

config/hooks/pre-commit
```diff
@@ -0,0 +1,12 @@
...
+    comma_files=$(echo "$files" | paste -s -d "," -)
+    "${REPO_ROOT_DIR}/gradlew" goJF -DgoogleJavaFormat.include="$comma_files" &>/dev/null
```
liblit: POSIX sh does not mandate support for `&>`. Replace `&>/dev/null` with POSIX-compliant equivalent `>/dev/null 2>&1`.

Caught by `shellcheck` linter.

build.gradle
liblit: Use tabs for indentation in `build.gradle` files, as is already being done elsewhere in this file.

build.gradle
```diff
@@ -10,8 +10,13 @@ plugins {
 	id 'nebula.lint' version '8.3.1'
 	id 'nebula.source-jar' version '7.0.1' apply false
 	id 'maven-publish'
+        id 'com.github.sherter.google-java-format' version '0.8'
 }
 
+repositories {
+  // to get the google-java-format jar and dependencies
```
liblit: Use tabs for indentation in `build.gradle` files, as is already being done elsewhere in this file.

build.gradle
```diff
@@ -103,6 +108,10 @@ task afterEclipseBuildshipImport(type: Exec) {
 }
 
 
+googleJavaFormat {
+  toolVersion = '1.7'
```
liblit: Use tabs for indentation in `build.gradle` files, as is already being done elsewhere in this file.

build.gradle
```diff
@@ -103,6 +108,10 @@ task afterEclipseBuildshipImport(type: Exec) {
...
+}
+
```
liblit: Two blank lines between major sections of this file.

build.gradle
```diff
@@ -10,8 +10,13 @@ plugins {
 	id 'nebula.lint' version '8.3.1'
 	id 'nebula.source-jar' version '7.0.1' apply false
 	id 'maven-publish'
+        id 'com.github.sherter.google-java-format' version '0.8'
 }
 
+repositories {
+  // to get the google-java-format jar and dependencies
+  mavenCentral()
+}
```
liblit: Two blank lines between major sections of this file.

build.gradle
```diff
@@ -103,6 +108,10 @@ task afterEclipseBuildshipImport(type: Exec) {
 }
 
 
+googleJavaFormat {
```
liblit: One blank line before this block, since it is within the same major section of the file.

config/hooks/install-pre-commit
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -e
+
+if [ ! -f .git/hooks/pre-commit ]; then
+    echo "Installing pre-commit hook"
+    cp config/hooks/pre-commit-stub .git/hooks/pre-commit
+    chmod +x .git/hooks/pre-commit
+fi
+
```
liblit: Remove trailing blank line.

msridhar:
@liblit can you take another look here?  I got rid of the `gradlew` hack.  When I land this, I will initially disable making all Gradle tasks depend on `installGitHooks`, as we want to do that along with re-formatting the code base.

liblit:
I've pushed several additional tweaks. The most important of these is probably 3cfcd1bb54f74335306cd1f87ae051df7beb91bb, which avoids undoing the effort from previously invested in task configuration avoidance (#418).

I have two remaining questions/concerns that go deeper than mere linting.

----

First, what is the benefit of using a `pre-commit-stub` script to launch the real `pre-commit` script? A comment in the former states that `if script cannot be found, exit without error (to not block local commits)`, but that does not explain why the stub is needed at all. Instead of copying `config/hooks/pre-commit-stub` into `.git/hooks/pre-commit`, why not just copy `config/hooks/pre-commit` into `.git/hooks/pre-commit` directly?

----

Second, I am not thrilled with the idea of a pre-commit hook that modifies commit contents. I see the value of a pre-commit hook that can _veto_ a commit. But silently changing the commit to be different from what I thought I was committing makes me uncomfortable. As a concrete example, suppose I use `git add --patch` or related features from other Git front-ends. So I have selected only a subset of the modified lines to include in a commit. The proposed pre-commit hook script will override that and `git add` the entire file. That seems like a nasty surprise.

If we are adamant that incorrectly-formatted code must never appear in the official WALA repository or even any local clone thereof, then let’s have a pre-commit hook that uses `verGJF` to veto commits with bad formatting. I strongly prefer that over a hook that changes the commit behind my back.

For that matter, a formatting check seems like just another kind of automated test to me. Let us seriously consider eliminating the pre-commit hook entirely. Instead, we can add `verGJF` to the Travis CI tasks that run for any new push to GitHub, including for pull requests. We already would not accept a pull request that failed any tests run by `./gradlew build`. We will notice and reject pull requests that fail `verGJF` as well. We just need a one-line change to one Travis CI script. No Git pre-commit hooks needed. Temporarily pushing code with bad formatting will be just like pushing code that fails any other tests: fine to do in your own private clone or branch, but you best believe we won’t accept that in a pull request until you clean it up.

msridhar:
> First, what is the benefit of using a `pre-commit-stub` script to launch the real `pre-commit` script? A comment in the former states that `if script cannot be found, exit without error (to not block local commits)`, but that does not explain why the stub is needed at all. Instead of copying `config/hooks/pre-commit-stub` into `.git/hooks/pre-commit`, why not just copy `config/hooks/pre-commit` into `.git/hooks/pre-commit` directly?

This was to be able to update the pre-commit hook logic via version control without having to copy it over again into the `.git/hooks` directory.  Not relevant for how the installation is currently configured, but could be related to the other point.

I am still in favor of a pre-commit hook by default, for the following reasons:

* Unlike test failures, we know how to automatically fix formatting; just run GJF.  I'm in favor of removing friction from the development process, so making things work such that you don't need to remember to run GJF by default seems to be a win.
* It's actually slightly annoying to do a localized GJF run.  If you do `./gradlew goJF` it will run on all Java source files, which is kind of slow.  The hook script has the advantage of just running GJF on the modified files.  We could maybe write a different script to replicate this behavior, but that just makes things more complicated.

I think we could modify things to make it possible to opt out of the pre-commit hook.  The basic idea would be that if `.git/hooks/pre-commit` already exists, don't automatically overwrite it.  Then, @liblit, you could opt out by just running `touch .git/hooks/pre-commit` (or adding your own hook with whatever logic you like).  We'd still run `verGJF` on CI to make sure things are formatted, independent of whether individual devs run GJF automatically or manually.  In this scenario, having `pre-commit-stub` is still useful: we won't overwrite `.git/hooks/pre-commit` once we've copied the stub once, but we can still update the logic by adding changes to the `pre-commit` script.

@liblit thoughts?

liblit:
@msridhar, we are both in agreement that we should add `verGJF` checking to our Travis CI jobs. Excellent.

You note that bad formatting can be fixed automatically, unlike other test failures. I had not previously considered that, but it is an important distinction and I am glad you pointed it out.

You and I both want the mass-reformatting to happen, and we both want to _maintain_ uniform formatting once we’ve made that mass-cleanup effort. We just need to figure out a strategy for ongoing enforcement that is as developer-friendly as possible.

You sketched out a scheme whereby the pre-commit hook would be installed, but not replaced if already present, and so I would manually install some sort of no-op hook to block the default pre-commit hook. This seems doable, but brittle. It does something potentially unexpected (modifying the commit) by default, and requires explicit action to prevent the risky hook from being installed.

I suggest that the opposite default would be safer and more robust: do not install any pre-commit hook automatically. Provide a reformatting script suitable for use as a pre-commit hook. But leave it to each WALA developer to _choose_ to install this reformatting script if he or she so desires. It’s a simple `cp`, or we could provide a Gradle task to install the script. But that task would run only when requested, not automatically as a dependency of every other task.

We could even offer two such hook scripts: one that reformats, and one that merely checks for correct formatting and vetoes the commit if formatting is wrong. Each developer can choose to install one, the other, or neither. And I do believe that “neither” is a reasonable option here given that formatting plugins are available for both IntelliJ IDEA and for Eclipse. One can even configure these to run automatically whenever one saves a file: [IntelliJ IDEA instructions](https://stackoverflow.com/q/946993/91929); [Eclipse instructions](http://www.eclipseonetips.com/2009/12/13/automatically-format-and-cleanup-code-every-time-you-save/).

Think of this as the “first, do no harm” option. A pre-commit hook that changes what will be committed can lead to nasty surprises. So let’s default to doing no harm, and give developers the tools to automate reformatting on commit only if they choose to accept the risk along with the convenience.

@juliandolby, would you care to weigh in?

liblit:
I mentioned that one can configure IntelliJ IDEA or Eclipse to reformat on save. I just noticed that IntelliJ IDEA also offers this as [an option before committing](https://www.jetbrains.com/help/idea/commit-changes-dialog.html#before_commit). Clearly there are diverse opinions as to when reformatting should happen.

Let’s provide what help we can, but let each developer decide what is best for him or her. Provide a suitable reformatting script, but don’t install it as the pre-commit hook unless explicitly requested.

msridhar:
I am ok with trying out the opt-in pre-commit hook.  Based on my experience, here is what I think will happen:

1. The main committers (~3 people right now) will configure things to their liking without issue.  For this category I think opt-in or opt-out doesn't matter much.

2. Folks submitting one-off PRs will have some additional friction.  They will sometimes get a Travis failure due to `verGJF` and they'll have to go find the instructions for how to fix or we'll have to tell them.  (This has happened to me when trying to contribute to other projects)

My prediction is that doing opt-in rather than opt-out will lead to an overall slightly worse experience for folks in category 2, who are the most affected (unless we get lots more regular committers, which would be great!).  But, I'm happy to try opt-in and see what happens.

liblit:
Good points again, @msridhar. “Based on my experience” and “This has happened to me” carry a lot of weight. I am OK with installing the `pre-commit` hook by default: opt-out instead of opt-in.

Can I ask that you look into adapting ideas from [here](https://stackoverflow.com/q/22858850/91929) to make partial (`git add --patch`) commits safer?

msridhar:
@liblit I read over the `git add --patch` scenario and understood a bit better.  It does seem like a nasty surprise to get changes added that you didn't intend.

I can imagine other technical solutions to make opt-in easier to discover.  But for now I think we should just go with opt-in, add some docs, and see how often issues crop up while we think further.

liblit:
Sounds good to me, @msridar. I’m glad we spent the time to hash through this from multiple angles.

I still hope to hear from @juliandolby about #409 in general, but I have no remaining technical concerns about this pull request in particular.

msridhar:
@liblit oof I force pushed without having pulled all your most recent changes.  Really sorry.  I think I recovered most of your changes but I think I'm missing a change in `build.gradle` that got rid of `new File` calls.  Do you still have it in your local clone, and if so can you restore it on the `ms/gjf-pre-commit` branch?

msridhar:
@liblit if you just push the latest commit you had from `ms/gjf-pre-commit` as a new branch, I can figure out how to rebase the changes

liblit:
I’ve restored two lost commits:

- 011575ea6156d8fe8108b7783e0840818f4f0c85 comes back as 1579f2ac6247abd828e1ba6823dee78b72fa44ef: Simplify creating `File` instances relative to the root project
- 59f51ef0d62e55e31d9211a8183467886f27cd22 comes back as b0df18c6295dfdb34ebc6d5d3af8973d3ede665a: Two blank lines between major sections, as elsewhere in this file

The only other commit that was lost was 3cfcd1bb54f74335306cd1f87ae051df7beb91bb. That commit improved how we add `installGitHooks` as a dependency of all other tasks so as to maintain task configuration avoidance. It looks like @msridhar has removed the original version of this dependency-adding code following our discussions above, so there is no longer any such code to improve.

liblit:
> I force pushed without having pulled all your most recent changes.

I rarely force push at all, but when I do, I _only_ do it as `--force-with-lease` instead of `--force`.  [_Much_ safer](https://developer.atlassian.com/blog/2015/04/force-with-lease/), and presumably would have blocked your not-up-to-date force push in this instance.

msridhar:
> I rarely force push at all, but when I do, I _only_ do it as `--force-with-lease` instead of `--force`.  [_Much_ safer](https://developer.atlassian.com/blog/2015/04/force-with-lease/), and presumably would have blocked your not-up-to-date force push in this instance.

Dang, what a great tip!  Yes that would have prevented this mistake 100%.  Thanks!!

PR #416, opened 2019-02-11
Fix license header formatting

msridhar:
Our previous license header formatting looked like Javadoc (since it started with `/**`), which messed up syntax highlighting and also tools like Google Java Format.  This patches the formatting of the header to not have long lines of `*` characters.  Change done via regex search-and-replace.


liblit:
No alphanumeric characters were harmed in the making of this commit. :laughing:

I thought we might be able to enable IntelliJ IDEA’s "Dangling Javadoc comment” inspection after this change, but no dice. There are still 148 such issues. Many of them are simply "`/** ... */`" comments in contexts where Javadoc is not expected, such as in the middle of a block of code. We might want to clean those up eventually too, but it need not happen now. I mention them just as a heads-up to @msridhar in case the Google Java Format tools turn out to have problems with these comments too.

msridhar:
> I thought we might be able to enable IntelliJ IDEA’s "Dangling Javadoc comment” inspection after this change, but no dice. There are still 148 such issues. Many of them are simply "`/** ... */`" comments in contexts where Javadoc is not expected, such as in the middle of a block of code. We might want to clean those up eventually too, but it need not happen now. I mention them just as a heads-up to @msridhar in case the Google Java Format tools turn out to have problems with these comments too.

Thanks do you have an example file or two?  I can keep an eye out on those if / when we reformat everything with GJF to see how things look.

liblit:
> do you have an example file or two?

Here are some representative cases:

* [completely empty Javadoc comment before `package` line](https://github.com/wala/WALA/blob/d639ed05941808cf990afa0acef8ce37f522f7a3/com.ibm.wala.cast/source/java/com/ibm/wala/cast/ir/translator/AbstractClassEntity.java#L12)
* [Javadoc comment before commented-out field](https://github.com/wala/WALA/blob/d639ed05941808cf990afa0acef8ce37f522f7a3/com.ibm.wala.dalvik/src/com/ibm/wala/dalvik/ssa/AbstractIntRegisterMachine.java#L93)
* [Javadoc comment before commented-out method](https://github.com/wala/WALA/blob/d639ed05941808cf990afa0acef8ce37f522f7a3/com.ibm.wala.dalvik/src/com/ibm/wala/dalvik/ssa/AbstractIntRegisterMachine.java#L380)
* [Javadoc comment as extra copyright block](https://github.com/wala/WALA/blob/f50866c8455fde438f75786d9badbceb46675e01/com.ibm.wala.scandroid/source/org/scandroid/util/AndroidAnalysisContext.java#L10)
* [single-line Javadoc comment in the middle of a code block](https://github.com/wala/WALA/blob/b2a589f8418192e78113bcf3103fd9673b5d7df7/com.ibm.wala.dalvik/src/com/ibm/wala/dalvik/util/AndroidEntryPointLocator.java#L204)
* [multi-line Javadoc comment in the middle of a code block](https://github.com/wala/WALA/blob/0752bf321c9a62548ebe38f1dc09b882916c7130/com.ibm.wala.dalvik/src/com/ibm/wala/dalvik/util/AndroidManifestXMLReader.java#L584)
* [Javadoc comment after last `enum` constant](https://github.com/wala/WALA/blob/75b9feac543d9567da512e1447165b1b0f85b5a4/com.ibm.wala.dalvik/src/com/ibm/wala/dalvik/ipa/callgraph/androidModel/parameters/AndroidModelParameterManager.java#L84); note as well that each of the preceding Javadoc comments will be misinterpreted as applying to the next following `enum` constant on the line below
* [single-line `BEGIN` or `END` Javadoc comment](https://github.com/wala/WALA/blob/d639ed05941808cf990afa0acef8ce37f522f7a3/com.ibm.wala.core/src/com/ibm/wala/analysis/reflection/ClassFactoryContextInterpreter.java#L66); these appear in a variety of places, including both within and outside of code blocks

If you’re already using IntelliJ IDEA, then Analyze → Run Inspection by Name → Dangling Javadoc comment  → Custom scope: WALA-Controlled → OK will find them all.

PR #422, opened 2019-02-13
Fix leak in LambdaSummaryClass

### LINKED ISSUE #419: Memory leaks in LambdaSummaryClass
Hello, 
we have occured memory leaks probably produced by WALA.

We are using WALA 1.5.0 (via Maven) in JUnit tests. We are computing new ClassHierarchy and CallGraph in every unit test. 

In class LambdaSummaryClass, static field _summaries_ seems to collect and hold information about ClassHierarchy from each of previous runs (see attached pictures). 
lambdasummaryclass
memory_before_test
memory_after_test
This way, we have memory issues after running higher amount of JUnit tests. Same issue will probably arise in case, when several separate WALA analyses are done in sequence. 

Is there any fix available for solving our issue (expect of manually clearing summaries field via Reflection after each test)?

msridhar:
Fixes #419 

Move the cache from `LambdaSummaryClass` to `LambdaMethodTargetSelector` and make it non-static.  There should be only one instance of `LambdaMethodTargetSelector` per analysis so this should remove the leak.

Also remove use of `WeakHashMap` as in these cases the keys are reachable from the values via strong references.

### DIFF
--- a/com.ibm.wala.core/src/com/ibm/wala/ipa/summaries/LambdaMethodTargetSelector.java
+++ b/com.ibm.wala.core/src/com/ibm/wala/ipa/summaries/LambdaMethodTargetSelector.java
@@ -10,9 +10,6 @@
  */
 package com.ibm.wala.ipa.summaries;
 
-import java.util.Map;
-import java.util.WeakHashMap;
-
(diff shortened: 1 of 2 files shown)


msridhar:
Note that I will squash commits before merging.

PR #427, opened 2019-02-22
Add NodeManager stream API

liblit:
This allows expressing some graph operations more succinctly by using the Java stream APIs.  However, be cautious: performance implications of Java streams are nontrivial.

For some folks, Java streams have a reputation for poor performance. The reality is more nuanced.  Yes, they can encourage needless abstraction, adding layers of indirect objects, making simple tasks unnecessarily slow.  But they can also speed things up by encouraging use of higher-level bulk operations.  Processing stream elements in parallel can also either boost or harm performance, depending on how many items are being manipulated and much potential concurrency is available.  So … it’s complicated.

### DIFF
--- a/com.ibm.wala.core/src/com/ibm/wala/ipa/callgraph/pruned/PrunedCallGraph.java
+++ b/com.ibm.wala.core/src/com/ibm/wala/ipa/callgraph/pruned/PrunedCallGraph.java
@@ -18,6 +18,7 @@ import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Stream;
 
 import com.ibm.wala.classLoader.CallSiteReference;
 import com.ibm.wala.classLoader.IMethod;
@@ -64,6 +65,12 @@ public class PrunedCallGraph implements CallGraph {
 		}
 		
 		return col.iterator();
+	}
+
+	@Override
+	public Stream<CGNode> stream() {
+		assert keep.stream().allMatch(cg::containsNode);
+		return keep.stream();
 	}
 
 	@Override
--- a/com.ibm.wala.core/src/com/ibm/wala/ipa/cfg/AbstractInterproceduralCFG.java
+++ b/com.ibm.wala.core/src/com/ibm/wala/ipa/cfg/AbstractInterproceduralCFG.java
@@ -402,13 +402,7 @@ public abstract class AbstractInterproceduralCFG<T extends ISSABasicBlock> imple
    */
   @Override
   public Iterator<BasicBlockInContext<T>> iterator() {
-    if (WARN_ON_EAGER_CONSTRUCTION) {
-      System.err.println("WARNING: forcing full ICFG construction by calling iterator()");
-    }
-    if (FAIL_ON_EAGER_CONSTRUCTION) {
-      throw new UnimplementedError();
-    }
-    constructFullGraph();
+    constructFullGraph("iterator");
     return g.iterator();
   }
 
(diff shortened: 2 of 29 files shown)


com.ibm.wala.core/src/com/ibm/wala/ipa/callgraph/pruned/PrunedCallGraph.java
```diff
@@ -66,6 +67,12 @@ public void removeNodeAndEdges(CGNode n) throws UnsupportedOperationException {
 		return col.iterator();
 	}
 
+	@Override
+	public Stream<CGNode> stream() {
+		assert keep.stream().allMatch(cg::containsNode);
```
msridhar: While we're editing this file, can we patch up `iterator()` to do something like `Collections.unmodifiableSet(keep).iterator()`?  I don't see a good reason for it to do what it's doing now.  Unless iterating over `keep` is *really* slow?  But iterating over the whole call graph is probably slower.

com.ibm.wala.core/src/com/ibm/wala/ipa/cfg/AbstractInterproceduralCFG.java
```diff
@@ -401,34 +402,38 @@ public void removeNodeAndEdges(BasicBlockInContext<T> N) throws UnsupportedOpera
...
+    constructFullGraph("getNumberOfNodes");
+    return g.getNumberOfNodes();
+  }
+
+  private boolean constructedFullGraph = false;
+
+  private void constructFullGraph(String onBehalfOf) {
     if (WARN_ON_EAGER_CONSTRUCTION) {
-      System.err.println("WARNING: forcing full ICFG construction by calling getNumberOfNodes()");
+      System.err.format("WARNING: forcing full ICFG construction by calling %s()\n", onBehalfOf);
```
msridhar: 👍 Nice refactoring

liblit: Thanks! I am particularly pleased with my choice of `onBehalfOf` as the formal parameter name. :nerd_face:

com.ibm.wala.core/src/com/ibm/wala/ipa/callgraph/pruned/PrunedCallGraph.java
liblit: @msridhar, your suggested replacement is only equivalent if `keep` is a subset of the nodes in `cg`. However, I think that is not a safe assumption. The existing code strongly implies that it is OK for `keep` to contain nodes not in `cg`. For example, this must be why `PrunedCallGraph::containsNode()` checks both `cg` and `keep` instead of just checking `keep`.

I wrote the assertion `keep.stream().allMatch(cg::containsNode)` in `PrunedCallGraph::stream()`, but I now believe that this assertion is wrong in general. Rather, `PrunedCallGraph::stream()` should allow `keep` to include nodes not in `cg`, and should do the right thing in that case. The change we really need here is to make `PrunedCallGraph::stream()` look more like `PrunedCallGraph::iterator()`. That is, each of these methods should yield only nodes in the _intersection_ of `keep` and `cg`.

That being said, we can certainly avoid creating the temporary `col` container that holds this intersection. Instead, we just iterate through one of the two sets of nodes and only yield each item if it is also in the other set, thereby computing the intersection on the fly without ever reifying it.

@msridhar: your thoughts?

liblit: Then again, `PrunedCallGraph::getNumberOfNodes()` is clearly assuming that `keep` is a subset of `cg`, since it returns the size of the former rather than the size of the intersection. OK, at this point I really do not know whether the subset assumption is valid or invalid. Some of the code hints one way, some hints the other. Help!

msridhar: We should have `keep` be a subset of `cg`.  I would say any code in contradiction of that assumption is buggy, and that we should add runtime checks (not just assertions) to fail fast if `keep` contains a node not in the `cg`, assuming it's not too expensive.

But, if desired we can address the broader problem in a follow-up issue / PR.

liblit: Your proposed invariant sounds fine to me. But yes, I like the idea of dealing with that invariant separately. The logic in this file was already inconsistent as to the `keep` ⊆ `cg` question even before #427. We came across it due to examining #427 closely, but fixing it is a distinct matter and can sensibly be addressed separately.

PR #428, opened 2019-02-23
Allow the build cache to work for verified downloads

liblit:
The build cache would have worked here on its own. Unfortunately I was accidentally disabling it due to an ill-conceived `upToDateWhen` call. See [Gradle issue #8564](https://github.com/gradle/gradle/issues/8584) for more discussion of [why I thought that call was useful](https://github.com/gradle/gradle/issues/8584#issuecomment-466451327), but [why it turned out to be harmful instead](https://github.com/gradle/gradle/issues/8584#issuecomment-466299895).

This should give second and subsequent rebuilds a nice performance boost. It might even speed up our Travis CI builds and make them less flaky, since we ask Travis CI to cache (snapshot and restore) the Gradle build cache across jobs.

That being said, the original `upToDateWhen` call was added to work around a specific problem, whereby Gradle would become convinced that the correct outcome of some download task is to have nothing downloaded at all. Once Gradle gets into this state, there seems to be no way to convince it to reattempt the download short of removing the project’s `.gradle` subdirectory and rebuilding everything from scratch. (Using `./gradlew --rerun-tasks` might work too; I have not tried it yet.) I am still seeing this problem crop up occasionally, as noted [here](https://github.com/gradle/gradle/issues/8584#issuecomment-466458821) and [here](https://github.com/gradle/gradle/issues/8584#issuecomment-466469525). I haven’t been able to reproduce the problem on demand, though. Until I do, **it’s hard to say whether we are better off with or without this change. I am torn, to be honest.**

### DIFF
--- a/buildSrc/src/main/groovy/verified-download.groovy
+++ b/buildSrc/src/main/groovy/verified-download.groovy
@@ -28,12 +28,6 @@ class VerifiedDownload extends org.gradle.api.DefaultTask {
 		return project.file(dest)
 	}
 
-	VerifiedDownload() {
-		outputs.upToDateWhen {
-			getDest().exists()
-		}
-	}
-
 	@TaskAction
 	downloadAndVerify() {
 		def destFile = getDest()


liblit:
Unfortunately I do _not_ think I am on the cusp of a better understanding. I think it’s going to take more field experience with this change to see how it behaves and perhaps start to notice when/if it gets itself into the bad state. For that, it will be better to have this change merged into `master` so that I and others will be exercising it on a regular basis.

So let’s be bold and merge this in. If you find that your build tree has gotten itself stuck in a state where a download is missing but the download task is believed to be complete, please let me know, and try to recall what actions led you to that state. You can also try `./gradlew --rerun-tasks downloadTaskName` where `downloadTaskName` is the name of the Gradle build task that should be downloading something but isn’t. Those tasks generally being with `download`, such as `downloadJLex` or `downloadAndroidSdk`.

liblit:
[I initially suggested](#issue-255640411) that this change “might even speed up our Travis CI builds and make them less flaky.” That is incorrect. We run our Travis CI Gadle builds with the `--no-build-cache` flag, so any cached downloads will be ignored.

Removing this flag would likely make our Travis CI jobs faster and less flaky as originally teased. But there may be some benefit to making Travis CI tests start from scratch each time, unable to accidentally benefit from anything that happened to be carried forward in a cache from an earlier job.

PR #432, opened 2019-02-26
In PrunedCallGraph, require that kept nodes be subset of unpruned graph

liblit:
In `PrunedCallGraph`, require that `keep` (the set of nodes to be kept) be a subset of the nodes in the unpruned graph.  Check for this at construction time, and raise an exception if violated.

The caller could potentially change either the `keep` set or the unpruned graph after constructing the `PrunedCallGraph` instance.  If they do that, then PrunedCallGraph might break, and the caller gets to keep all the pieces.  We could be rechecking the “subset” relationship in each and every place it is assumed to hold, but that seems excessive.  So instead we just add a big warning in the constructor’s Javadoc documentation.

See <https://github.com/wala/WALA/pull/427#discussion_r259913855> for the concluding piece of the discussion that led us to enact this “subset” requirement.

### DIFF
--- a/com.ibm.wala.core/src/com/ibm/wala/ipa/callgraph/pruned/PrunedCallGraph.java
+++ b/com.ibm.wala.core/src/com/ibm/wala/ipa/callgraph/pruned/PrunedCallGraph.java
@@ -36,9 +36,24 @@ public class PrunedCallGraph implements CallGraph {
 	private Set<CGNode> keep;
 	private Map<CGNode,Set<CGNode>> remove = Collections.emptyMap();
 	
+	/**
+	 * Create a pruned (filtered) view of an existing call graph.
+	 *
+	 * <p>Note: the created instance retains references to {@code cg} and {@code keep} without making private copies
+	 * thereof.  Be cautious if subsequently modifying {@code cg} or {@code keep} outside of this class's control.  In
+	 * particular, {@code keep} must always be a subset of the nodes in {@code cg} or else unexpected behavior may
+	 * arise.</p>
+	 *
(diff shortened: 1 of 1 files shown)


com.ibm.wala.core/src/com/ibm/wala/ipa/callgraph/pruned/PrunedCallGraph.java
```diff
@@ -36,9 +36,24 @@
 	private Set<CGNode> keep;
 	private Map<CGNode,Set<CGNode>> remove = Collections.emptyMap();
 	
+	/**
+	 * Create a pruned (filtered) view of an existing call graph.
+	 *
+	 * <p>Note: the created instance retains references to {@code cg} and {@code keep} without making private copies
```
msridhar: Is there a good reason for not just making a private copy?  Would that break existing uses?

liblit: Making private copies would impose compute and memory demands, of course. So there is a safety/performance trade-off.

Beyond that, making private copies requires cloning the unpruned `cg` call graph deeply enough that the private copy is insulated from outside mutation. None of the `CallGraph` implementations override `Object.clone()`. For most `CallGraph`-implementing classes, I very much doubt that the default `Object.clone()` operates deeply enough to give us a truly private `cg` copy that nobody else can mutate.

msridhar:
Yes, true, call graphs are mutable. I was thinking more about copying
`keep` but probably it’s fine to not do that.

PR #436, opened 2019-03-12
Fix dangling Javadoc comments

liblit:
These commits fix various instances of Javadoc comments without an associated item to comment upon. Such comments arise for a variety of reasons; each commit fixes all instances of one broad category.

### DIFF
--- a/.idea/inspectionProfiles/Javadoc.xml
+++ b/.idea/inspectionProfiles/Javadoc.xml
@@ -25,7 +25,10 @@
     <inspection_tool class="BashDuplicateFunction" enabled="false" level="WARNING" enabled_by_default="false" />
     <inspection_tool class="BashEvaluateArithmeticExpression" enabled="false" level="WARNING" enabled_by_default="false" />
     <inspection_tool class="BashEvaluateExpression" enabled="false" level="WARNING" enabled_by_default="false" />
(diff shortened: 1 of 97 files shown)


liblit:
@msridhar wrote:
> It'd be great if we could enforce Javadoc clean-ness on CI, but not sure if that's possible

We get a certain amount of Javadoc checking from Travis CI. We [build the `javadoc` target](https://github.com/wala/WALA/blob/205a94d08fe407960e299bff3c9491c0276a34a3/travis/script-gradle#L21) as part of testing, and we do that with [all Javadoc linter options enabled except for complaining about missing documentation](https://github.com/wala/WALA/blob/205a94d08fe407960e299bff3c9491c0276a34a3/build.gradle#L187).

Of course, if that specific Javadoc implementation does not check for problem _x_, then we won’t have “disallow problem _x_” enforcement. The dangling Javadoc comments I found were detected by a specific IntelliJ IDEA inspection, but not by any other tools. Apparently it is possible to [run IntelliJ IDEA inspections without the full, interactive IDE](https://www.jetbrains.com/help/idea/running-inspections-offline.html). I have not tried this, but it should let us use the same IntelliJ IDEA inspection profile for both interactive (IDE-based) checking and Travis CI enforcement. It would require downloading and unpacking IntelliJ IDEA into each Travis CI job, though.

PR #440, opened 2019-03-17
Move responsibilities among Travis CI lifecycle scripts

liblit:
Per <https://docs.travis-ci.com/user/job-lifecycle/>, the `install` stage should "install any dependencies required", while the `script` stage should "run the build script".

For our purposes, it makes sense to treat WALA's many assorted extra download tasks as `install`-stage dependencies: `downloadAndroidSdk`, `downloadBcel`, `downloadJavaCup`, etc.  If any of these fails, I believe Travis CI will treat that job as "errored" rather than "failed".  That might make it easier for us to quickly recognize when the problem was a transient inability to download (which happens often).

Having moved pre-build downloads into the `install` stage, it makes sense for the `script` stage to encompass both building WALA and running all of its various checks/tests.  In particular, `verifyGoogleJavaFormat` is now done in the `script` stage rather than the `install` stage.

### DIFF
--- a/travis/script-gradle
+++ b/travis/script-gradle
@@ -19,4 +19,5 @@ esac
 $headless ./gradlew --continue --no-build-cache --stacktrace \
 	  "$SUBMODULE_PREFIX"build \
 	  ${documentation:+$SUBMODULE_PREFIX$documentation} \
-	  lintGradle
+	  lintGradle \
+	  verifyGoogleJavaFormat
--- a/travis/script-gradle
+++ b/travis/script-gradle
@@ -17,7 +17,7 @@ case "$SUBMODULE_PREFIX" in
 esac
 
 $headless ./gradlew --continue --no-build-cache --stacktrace \
+	  verifyGoogleJavaFormat \
 	  "$SUBMODULE_PREFIX"build \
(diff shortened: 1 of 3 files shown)


travis/script-gradle
```diff
@@ -19,4 +19,5 @@ esac
 $headless ./gradlew --continue --no-build-cache --stacktrace \
 	  "$SUBMODULE_PREFIX"build \
 	  ${documentation:+$SUBMODULE_PREFIX$documentation} \
-	  lintGradle
+	  lintGradle \
+	  verifyGoogleJavaFormat
```
msridhar: Can we list `verifyGoogleJavaFormat` first here?  It's a faster check than the `build` tasks, and this way you won't have to wait all the way through execution of regression tests to see that there is a formatting issue.  (I think on CI Gradle runs these tasks in the order they are listed after running dependent tasks.)

msridhar: You can see [here](https://travis-ci.org/wala/WALA/jobs/507474346) that with the current ordering `verifyGoogleJavaFormat` is run last

liblit: Good suggestion. Thanks! Enacted in 8f18698ec9d0a55aef457dae471c251de17947bb.

PR #447, opened 2019-04-18
Some support for JDK 11 in Java bytecode frontend

msridhar:
This is work-in-progress WALA support for JDK 11, based on #371.  Full WALA support for JDK 11 will require significantly more work, but I thought it would be best to commit the changes to master incrementally rather than trying to fix everything all at once.

The main functional change here is from #371: on JDK 11, we use the `.jmod` files packaged with OpenJDK as the standard library when otherwise unspecified, rather than relying on the bootclasspath, which has been removed.  In order to get this working on Linux, I had to remove some older logic from `WalaProperties` for guessing the stdlib directory, as it breaks things on JDK 11.

We also add enough support to get most of the WALA core tests passing on JDK 11.  Necessary changes included removing the Eclipse plugin config from some test projects and a couple other minor tweaks.

We change the Travis config to test on JDK 11 on both Mac and Linux.  For now, we only run the `com.ibm.wala.core.tests` tests, and we disable some tests that don't yet pass.  I will open follow-up PRs to fix these tests individually.


com.ibm.wala.util/src/com/ibm/wala/util/PlatformUtil.java
```diff
@@ -46,14 +50,29 @@ public static boolean onIKVM() {
...
+    String classpath = null;
+    String javaVersion = System.getProperty("java.specification.version");
+    if (javaVersion.equals("9") || javaVersion.equals("10") || javaVersion.equals("11")) {
+      // java11 support for jmod files
+      try {
+        classpath =
+            String.join(
+                File.pathSeparator,
+                Files.list(Paths.get(System.getenv("JAVA_HOME"), "jmods"))
+                    .map(x -> x.toString())
```
liblit: Prefer method reference instead of lambda: `.map(Path::toString)`

msridhar: Fixed

PR #448, opened 2019-04-18
cache the p2asmaven directory on Travis

### LINKED ISSUE #446: Cache p2AsMaven deps on Travis?
I am wondering if it would be possible to cache the p2AsMaven downloads / steps performed by the Gradle build on Travis.  These seem to not be cached and take quite a long time on each Travis job.  @liblit would this be possible, or is it dangerous?

msridhar:
Fixes #446 

Beyond caching this directory, update the cleanup script to remove a few other files inside the Gradle cache that seem to change needlessly between runs.  If we don't remove these files, Travis is forced to re-upload on each build, which takes about a minute.  I've confirmed by repeatedly running a build job that deleting these files does not put the cache in an inconsistent state that breaks future builds.  I also filed gradle/gradle#9175 as a start at trying to get some root cause issues fixed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ cache:
   - "$HOME/.m2"
   - "$HOME/apache-maven-3.5.0"
   - "$TRAVIS_BUILD_DIR/com.ibm.wala.core.testdata/ocaml/ocamljava-2.0-alpha1/lib"
+  - "$TRAVIS_BUILD_DIR/build/p2asmaven"
 env:
   global:
     - secure: KcugjQYnBqeZ7XenZD5QY7jhekVPO0QpQyjDUteLytaokhyRK2g6eNvr/pPerN2uWUvsPwO18P9F+oOupge1cpPZf4cEY8RzLIromyUoRWd6JA0SKciUYdN2kSqnC4uZSJGXeGLoroyEEL4Q2sqimpkbIGxgxYtVniWgJULOyR4=
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,8 +25,8 @@ cache:
   - "$HOME/.gradle/wrapper"
   - "$HOME/.m2"
   - "$HOME/apache-maven-3.5.0"
-  - "$TRAVIS_BUILD_DIR/com.ibm.wala.core.testdata/ocaml/ocamljava-2.0-alpha1/lib"
   - "$TRAVIS_BUILD_DIR/build/p2asmaven"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -26,6 +26,7 @@ cache:
   - "$HOME/.m2"
   - "$HOME/apache-maven-3.5.0"
   - "$TRAVIS_BUILD_DIR/com.ibm.wala.core.testdata/ocaml/ocamljava-2.0-alpha1/lib"
+  - "$TRAVIS_BUILD_DIR/build/p2asmaven"
```
liblit: Swap with preceding line to keep this list in alphabetical order.

travis/before-cache-gradle
```diff
@@ -5,3 +5,9 @@ caches="$HOME/.gradle/caches"
 rm -rf \
    "$caches/modules-2/modules-2.lock" \
```
liblit: Remove, since this is now redundant with respect to the `*.lock` deletion performed by one of the `find` commands below.

travis/before-cache-gradle
```diff
@@ -5,3 +5,9 @@ caches="$HOME/.gradle/caches"
...
    "$caches"/*/plugin-resolution
+
+find "$caches" -name '*.lock' -delete
+find "$caches" -name '*.bin' -delete
+find "$caches" -name 'cache.properties' -delete
```
liblit: Remove unnecessary single quotes around filename with no special characters.

travis/before-cache-gradle
```diff
@@ -5,3 +5,9 @@ caches="$HOME/.gradle/caches"
 rm -rf \
    "$caches/modules-2/modules-2.lock" \
    "$caches"/*/plugin-resolution
+
+find "$caches" -name '*.lock' -delete
```
liblit: Sort these multiple `find` commands alphabetically. Consider fusing into a single `find` command, but OK to skip that if it looks like it will be significantly less readable.

liblit:
Nice performance boost! I am glad you decided to try this out.

Kudos on the thorough investigation that led to gradle/gradle#9175. That is truly some professional-grade bug reporting!

.travis.yml
msridhar: Fixed

travis/before-cache-gradle
msridhar: Fixed

travis/before-cache-gradle
msridhar: I sorted them, putting `*` commands at the end

travis/before-cache-gradle
msridhar: Fixed

PR #451, opened 2019-04-27
List some required IntelliJ IDEA plugins

liblit:
"Required" here is interpreted broadly.  Strictly speaking, one doesn't _need_ to have the Google Java formatter, Git integration, or JUnit integration.  But these are strongly recommended, so let's try requiring them and see if anyone complains.

A more strict treatment of "required" would list only the Gradle plugin and the version of IntelliJ IDEA itself.  Regarding the latter, unfortunately I haven't actually been able to get IntelliJ IDEA to complain no matter what "min-version" I list for the IDE itself.  Not sure what I'm doing wrong here.

### DIFF
--- a/.idea/externalDependencies.xml
+++ b/.idea/externalDependencies.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ExternalDependencies">
+    <plugin id="com.intellij" min-version="191" />
+    <plugin id="Git4Idea" />
+    <plugin id="google-java-format" />
+    <plugin id="JUnit" />
+    <plugin id="org.jetbrains.plugins.gradle" />
+  </component>
+</project>
--- a/.idea/externalDependencies.xml
+++ b/.idea/externalDependencies.xml
@@ -2,7 +2,6 @@
 <project version="4">
   <component name="ExternalDependencies">
     <plugin id="com.intellij" min-version="191" />
-    <plugin id="Git4Idea" />
     <plugin id="google-java-format" />
     <plugin id="JUnit" />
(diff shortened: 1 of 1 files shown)


msridhar:
I don’t think I have these plugins installed 🙂 Maybe I’m missing out. Will I see errors in IntelliJ with this change?

liblit:
When you open WALA as an IntelliJ IDEA project, the IDE should warn you that you are missing some required plugins, and perhaps offer links into the plugin manager so that you can find and install them. You are probably free to ignore these warnings too. My intent, though, is to list plugins that a WALA developer would _want_ to have if he or she were aware of them. So if you find that there are listed plugins that you intend to avoid indefinitely, we should consider removing those from the "required" list entirely.

msridhar:
I think Git interface is often a personal preference so maybe we should not
make the Git plugin required. The others sound fine.

liblit:
@msridhar, the two Java 11 Travis CI jobs are consistently failing almost immediately. Something is going wrong with the command to install JDK 11:

```
$ ~/bin/install-jdk.sh --target "/home/travis/openjdk11" --workspace "/home/travis/.cache/install-jdk" --feature "11" --license "GPL" --cacerts
install-jdk.sh 2019-04-18
The command "~/bin/install-jdk.sh --target "/home/travis/openjdk11" --workspace "/home/travis/.cache/install-jdk" --feature "11" --license "GPL" --cacerts" failed and exited with 51 during .
```

The failures do not seem to involve any of the changes from this pull request. Can you take a look?

msridhar:
Jdk 11 failures look like a Travis issue. Maybe we can try again tomorrow? If you’d like to go ahead and land today I don’t object

liblit:
OK, thanks @msridhar.

PR #454, opened 2019-04-28
Use Java 5 language/library features for clarity and simplicity

liblit:
These changes put various Java 5 language and library features to use for improved code clarity and simplicity. All were found and applied using various IntelliJ IDEA inspections, followed by manual examination (and in a few cases. reversal) of the changes.


msridhar:
Regarding JDK 11 Travis failures, here are some explanations:

https://travis-ci.community/t/install-of-openjdk11-is-failing-again/3061/21
https://github.com/sormuras/bach/issues/51

I may try [this fix](https://github.com/sormuras/bach/issues/51#issuecomment-487385445) which seems more permanent?  In any case will try to fix soon so we can ensure no regressions there

com.ibm.wala.cast.java.ecj/src/com/ibm/wala/cast/java/translator/jdt/JDTJava2CAstTranslator.java
```diff
@@ -572,8 +572,8 @@ else if (met.getParameterTypes().length > 0)
       // this.
 
       List<CAstNode> bodyNodes = new ArrayList<>(staticInits.size());
-      for (int i = 0; i < staticInits.size(); i++)
-        bodyNodes.add(visitFieldInitNode(staticInits.get(i), newContext));
+      for (ASTNode staticInit : staticInits)
```
msridhar: Do we know what kind of bytecodes this generates?  I think the modified code will allocate and use an iterator, rather than using `get()` calls.  In performance-critical code this could cause a slight performance degradation.  But I'm not 100% sure.  For arrays I think the same code comes out so we should be fine.

liblit: Consensus opinion I am seeing online (e.g., [here](https://stackoverflow.com/q/9226483/91929) or [here](https://stackoverflow.com/q/256859/91929) is that the two variants are equally efficient, or that the for-each variant might be slightly faster by computing the upper bound just once.

msridhar: Ok.  I'm not 100% convinced, but we should have real benchmarks if we are worried about this stuff.  Probably in most perf-critical spots we are already using arrays anyway.

PR #456, opened 2019-04-28
Fix Linux OpenJDK 11 builds and disable Mac OpenJDK11

msridhar:
If we use the Xenial Ubuntu release we get JDK 11 by default.  Mac JDK 11 builds are still broken, so we disable them for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,11 +38,11 @@ matrix:
       env: BUILD_SYSTEM=gradle
       language: java
       jdk: openjdk11
-    - os: osx
-      osx_image: xcode8.3
-      env: BUILD_SYSTEM=gradle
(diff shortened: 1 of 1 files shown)


liblit:
These changes seem reasonable to me, but [Travis CI thinks otherwise](https://travis-ci.org/wala/WALA/builds/525698928). I’m going to withhold review approval until Travis CI is happy too.

msridhar:
Oof, yeah, our `script-gradle` relies on getting the Java version from `TRAVIS_JDK_VERSION` which isn't set unless you do `jdk: openjdk11`, even though it's unnecessary on Xenial.  Reverted a commit, hopefully it's fixed now

PR #457, opened 2019-04-28
Download `parser.cup` from Maven, not Georgia Tech

### LINKED ISSUE #455: Download parser.cup file from Maven Central
We're currently seeing Travis failures due to a failure to download http://www.cc.gatech.edu/gvu/people/faculty/hudson/java_cup/classes.v0.9e/java_cup/parser.cup.  It looks like the file is available in this jar from Maven Central, which should be a more reliable host:

https://repo1.maven.org/maven2/java_cup/java_cup/0.9e/java_cup-0.9e-sources.jar

@liblit how hard would it be to download this jar instead and extract the file we need from there?

liblit:
Fixes #455, wherein Georgia Tech downloads were failing. Even if those faillures were only temporary, Maven should be a more reliable long-term source for this file.

Furthermore, extracting this file from a Maven-tracked artifact means that we can use Gradle’s standard mechanisms for declaring dependencies and resolving those dependencies in online repositories like Maven Central. Nice! I wish more of WALA’s required downloads were available from Maven repositories as well.

The rebuilt `downloadSampleCup` task looks like it should be a `Copy` task, since all it does is extract a file from a Jar archive. However, a `Copy` task will always treat the entire destination directory as its task outputs. This specific task is only adding one file to an existing directory; the rest of the directory is not this task’s responsibility. We want more fine-grained control over what this task declares as its outputs. Therefore, we implement `downloadSampleCup` as a generic task that uses the `Project.copy` method rather than as a `Copy` task.

### DIFF
--- a/com.ibm.wala.dalvik.test/build.gradle
+++ b/com.ibm.wala.dalvik.test/build.gradle
@@ -116,10 +116,18 @@ tasks.register('copyAndroidJar', Sync) {
 	into temporaryDir
 }
 
-tasks.register('downloadSampleCup', VerifiedDownload) {
-	src 'http://www.cc.gatech.edu/gvu/people/faculty/hudson/java_cup/classes.v0.9e/java_cup/parser.cup'
-	dest 'data/sample.cup'
-	checksum '76b549e7c6e802b811a374248175ecf4'
+tasks.register('downloadSampleCup') {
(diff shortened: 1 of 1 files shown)


liblit:
Travis CI passes except for JDK 11 jobs: good enough until #456 is merged.

PR #463, opened 2019-05-04
Switch to maven Gradle plugin

msridhar:
We should use Gradle for creating Maven artifacts, to enable deprecation of WALA Maven builds.  I couldn't get proper artifacts built with the maven-publish plugin, so this PR switches to the maven plugin.  It makes use of the [gradle-mvn-push](https://github.com/chrisbanes/gradle-mvn-push/blob/master/gradle-mvn-push.gradle) script, with slight modifications.

I've tested locally and successfully installed artifacts to the local M2 repository.  Once this lands, I plan to enable uploading of snapshots on every successful CI build of master, so that will give us some testing on CI.

To test locally, run `./gradlew install` and then check the contents of `~/.m2/repository/com/ibm/wala` for the artifacts.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,6 @@ plugins {
 	id 'de.undercouch.download'
 	id 'nebula.lint' version '9.2.0'
 	id 'nebula.source-jar' version '9.5.0' apply false
-	id 'maven-publish'
 }
 
 repositories {
@@ -73,7 +72,6 @@ subprojects { subproject ->
 	}
 
 	apply plugin: 'java'
-	apply plugin: 'maven-publish'
 	apply plugin: 'nebula.source-jar'
 
 	version rootProject.version
@@ -86,12 +84,6 @@ subprojects { subproject ->
 	}
 
 	jar.manifest.from('META-INF/MANIFEST.MF')
-
-	publishing.publications {
-		mavenJava(MavenPublication) {
-			from components.java
-		}
-	}
 
 	tasks.register('afterEclipseBuildshipImport') {
 		dependsOn 'processTestResources'
--- a/com.ibm.wala.cast.java.ecj/build.gradle
+++ b/com.ibm.wala.cast.java.ecj/build.gradle
@@ -2,14 +2,14 @@ sourceSets.main.java.srcDirs = ['src']
 
 dependencies {
 	compile(
-		'eclipse-deps:org.eclipse.core.runtime:+',
-		'eclipse-deps:org.eclipse.equinox.common:+',
-		'eclipse-deps:org.eclipse.jdt.core:+',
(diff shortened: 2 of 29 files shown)


msridhar:
The version of the script here almost certainly started from [this version](https://github.com/uber/android-template/blob/master/gradle/gradle-mvn-push.gradle) in the Uber template repo for open-source projects.  That version has been battle-tested across multiple popular projects, so I trust it.  Tomorrow I'll take a closer look and clarify the differences from that one.  The main one was that I commented out the lines that turn off linting for Javadoc tasks, as I was pretty sure we want that to be on.

liblit:
If you prefer to base our copy on Uber’s version instead of the original author’s version, that’s fine with me. But we should at least note the origin somewhere, such as in a comment in the script itself and/or in the commit message. Otherwise, future maintainers may have to hunt for the point of origin with much less information than you have right now.

msridhar:
@liblit I've added a comment clarifying the source of the script and inline comments where I made a change.  Here is the diff from the upstream script.

```diff
--- gradle-mvn-push.gradle	2019-05-05 08:34:39.000000000 -0700
+++ /Users/msridhar/git-repos/WALA/gradle-mvn-push.gradle	2019-05-05 08:40:04.000000000 -0700
@@ -14,6 +14,9 @@
  * limitations under the License.
  */
 
+// This script is derived from the version here:
+// https://github.com/uber/android-template/blob/master/gradle/gradle-mvn-push.gradle
+
 apply plugin: 'maven'
 apply plugin: 'signing'
 
@@ -91,10 +94,17 @@
   }
 
   signing {
-    required { isReleaseBuild() && gradle.taskGraph.hasTask("uploadArchives") }
+    // Use external gpg cmd.  This makes it easy to use gpg-agent,
+    // to avoid being prompted for a password once per artifact
+    useGpgCmd()
     sign configurations.archives
   }
 
+  // only sign the archives if we are uploading a release build
+  tasks.withType(Sign) {
+    onlyIf { isReleaseBuild() && gradle.taskGraph.hasTask("uploadArchives") }
+  }
+
   if (project.getPlugins().hasPlugin('com.android.application') ||
           project.getPlugins().hasPlugin('com.android.library')) {
     task install(type: Upload, dependsOn: assemble) {
@@ -204,14 +214,6 @@
     }
   }
 
-  if (JavaVersion.current().isJava8Compatible()) {
-    allprojects {
-      tasks.withType(Javadoc) {
-        options.addStringOption('Xdoclint:none', '-quiet')
-      }
-    }
-  }
-
   artifacts {
     if (project.getPlugins().hasPlugin('com.android.application') ||
             project.getPlugins().hasPlugin('com.android.library')) {
```

PR #473, opened 2019-05-06
Upload snapshot builds on Travis

msridhar:
This will make it easier for users to experiment with cutting-edge WALA versions without building from source.  We upload from the master branch only, on successful builds.  You can see a successful upload here (expand the `after_success` section at the end):

https://travis-ci.org/wala/WALA/jobs/528819071


deploy_snapshot.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
liblit: Change to `#!/bin/bash -eux` as is used for the other `travis/*` scripts:

* `-e` to exit immediately if any command fails
* `-u` to treat uses of unset `$VARIABLES` as errors
* `-x` to verbosely show what commands are being run in the Travis CI job log

deploy_snapshot.sh
```diff
@@ -0,0 +1,28 @@
...
+#
+# Deploy a jar, source jar, and javadoc jar to Sonatype's snapshot repo.
+#
+# Adapted from https://coderwall.com/p/9b_lfq and
+# http://benlimmer.com/2013/12/26/automatically-publish-javadoc-to-gh-pages-with-travis-ci/
+
+SLUG="wala/WALA"
```
liblit: No need for quoting on this line or the three lines that follow.

deploy_snapshot.sh
```diff
@@ -0,0 +1,28 @@
...
+JDK="oraclejdk8"
+BRANCH="master"
+OSNAME="linux"
+set -e
```
liblit: Remove this line after adding `-eux` to first line of script.

gradle-mvn-push.gradle
```diff
@@ -45,11 +45,15 @@ def getRepositoryPassword() {
   return hasProperty('SONATYPE_NEXUS_PASSWORD') ? SONATYPE_NEXUS_PASSWORD : ""
 }
 
+def needsSigning() {
+  return isReleaseBuild() && gradle.taskGraph.hasTask("uploadArchives")
```
liblit: Single-quote `'uploadArchives'`, since we do not need `$` expansion in this string.

travis/after-success-gradle
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh -eux
+
+bash ./deploy_snapshot.sh
```
liblit: Remove `bash `, so that the `deploy_snapshot.sh` script can choose its own interpreter (and interpreter flags) via an initial `#!` line.

PR #475, opened 2019-05-06
Only trigger dependent project builds on WALA master branch

### LINKED ISSUE #474: only fire off builds on dependent projects once for each commit landed to master
`dependent-projects-trigger.sh` should only fire off dependent builds when a commit lands on master.  It should have logic like this:

https://github.com/wala/WALA/blob/0c20088388a5ca39ee00b1fd933b11ba8defb96d/upload-javascript.sh#L3-L6

Travis CI throttles the # of builds running simultaneously for a given project / org (see [here](https://stackoverflow.com/a/34809448/1126796)),  so this will reduce the lag for running useful WALA CI builds.

msridhar:
Fixes #474

### DIFF
--- a/dependent-projects-trigger.sh
+++ b/dependent-projects-trigger.sh
@@ -1,54 +1,57 @@
 #!/bin/bash
 
-body='{
-"request": {
-"branch":"master"
-}}'
+if [ "$TRAVIS_REPO_SLUG" == "wala/WALA" ] &&
+   [ "$TRAVIS_PULL_REQUEST" == "false" ] &&
+   [ "$TRAVIS_BRANCH" == "master" ]; then
 
-curl -s -X POST \
-   -H "Content-Type: application/json" \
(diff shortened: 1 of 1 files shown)


dependent-projects-trigger.sh
```diff
@@ -1,54 +1,57 @@
...
-   https://api.travis-ci.org/repo/SunghoLee%2FHybriDroid/requests
-
-curl -s -X POST \
-   -H "Content-Type: application/json" \
-   -H "Accept: application/json" \
-   -H "Travis-API-Version: 3" \
-   -H "Authorization: token $1" \
-   -d "$body" \
-   https://api.travis-ci.org/repo/wala%2FML/requests
+if [ "$TRAVIS_REPO_SLUG" == "wala/WALA" ] &&
```
liblit: On this line and the two following lines, Use `=` instead of `==` for POSIX conformance.

PR #477, opened 2019-05-07
Replace static Collections.sort with non-static List.sort

liblit:


liblit:
More concise. More object-oriented. Removes an extra function call (`Collections.sort` will presumably merely call `List.sort`). Avoids importing `Collections` if not used elsewhere in the same file.

PR #478, opened 2019-05-07
Use Arrays.setAll instead of explicit loop where possible

liblit:
The `Arrays.setAll` version is certainly concise, especially where the second argument is a method reference, as in `Arrays.setAll(uses, phi::getUse);`.  However, I could imagine the heavy use of method
references and lambdas being a performance concern.  Let’s consider this group of changes carefully.

We could potentially use `Arrays.parallelSetAll` in cases where the values computed for each slot are independent of those computed for other slots.  However, I don’t know which of the arrays are typically
large enough to make parallel execution a good idea.

### DIFF
--- a/com.ibm.wala.shrike/src/com/ibm/wala/shrikeBT/Decoder.java
+++ b/com.ibm.wala.shrike/src/com/ibm/wala/shrikeBT/Decoder.java
@@ -1108,7 +1108,7 @@ public abstract class Decoder implements Constants {
         }
       }
     } else {
-      Arrays.setAll(handlers, i -> noHandlers);
+      Arrays.fill(handlers, noHandlers);
     }
 
     decoded = null;
--- a/com.ibm.wala.shrike/src/com/ibm/wala/shrikeBT/Decoder.java
+++ b/com.ibm.wala.shrike/src/com/ibm/wala/shrikeBT/Decoder.java
@@ -1108,9 +1108,7 @@ public abstract class Decoder implements Constants {
         }
       }
     } else {
-      for (int i = 0; i < handlers.length; i++) {
-        handlers[i] = noHandlers;
-      }
+      Arrays.setAll(handlers, i -> noHandlers);
     }
 
     decoded = null;
--- a/com.ibm.wala.util/src/com/ibm/wala/util/heapTrace/HeapTracer.java
+++ b/com.ibm.wala.util/src/com/ibm/wala/util/heapTrace/HeapTracer.java
@@ -118,9 +118,7 @@ public class HeapTracer {
     }
     String[] result = new String[classFileNames.size()];
     Iterator<String> it = classFileNames.iterator();
-    for (int i = 0; i < result.length; i++) {
-      result[i] = it.next();
-    }
+    Arrays.setAll(result, i -> it.next());
     return result;
   }
 
@@ -460,9 +458,7 @@ public class HeapTracer {
       }
       Field[] result = new Field[s.size()];
       Object[] temp = s.toArray();
-      for (int i = 0; i < result.length; i++) {
-        result[i] = (Field) temp[i];
-      }
+      Arrays.setAll(result, i -> (Field) temp[i]);
       allReferenceFieldsCache.put(c, result);
       return result;
     }
--- a/com.ibm.wala.util/src/com/ibm/wala/util/heapTrace/HeapTracer.java
+++ b/com.ibm.wala.util/src/com/ibm/wala/util/heapTrace/HeapTracer.java
@@ -24,7 +24,6 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.IdentityHashMap;
-import java.util.Iterator;
 import java.util.Set;
 import java.util.Stack;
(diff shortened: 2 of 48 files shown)


com.ibm.wala.util/src/com/ibm/wala/util/heapTrace/HeapTracer.java
```diff
@@ -118,9 +119,7 @@ public static void main(String[] args) {
     }
     String[] result = new String[classFileNames.size()];
     Iterator<String> it = classFileNames.iterator();
-    for (int i = 0; i < result.length; i++) {
-      result[i] = it.next();
-    }
+    Arrays.setAll(result, i -> it.next());
```
msridhar: `toArray`

com.ibm.wala.util/src/com/ibm/wala/util/heapTrace/HeapTracer.java
```diff
@@ -460,9 +459,7 @@ private void traverseFieldOfScalar(
       }
       Field[] result = new Field[s.size()];
       Object[] temp = s.toArray();
-      for (int i = 0; i < result.length; i++) {
-        result[i] = (Field) temp[i];
-      }
+      Arrays.setAll(result, i -> (Field) temp[i]);
```
msridhar: `s.toArray(new Field[0])`

com.ibm.wala.util/src/com/ibm/wala/util/intset/MutableSparseLongSetFactory.java
```diff
@@ -43,7 +44,7 @@ public MutableLongSet make(long[] set) {
   public MutableLongSet parse(String string) throws NumberFormatException {
     int[] backingStore = SparseIntSet.parseIntArray(string);
     long[] bs = new long[backingStore.length];
-    for (int i = 0; i < bs.length; i++) bs[i] = backingStore[i];
+    Arrays.setAll(bs, i -> backingStore[i]);
```
msridhar: Could possibly use streams like: https://stackoverflow.com/a/37882699/1126796

com.ibm.wala.shrike/src/com/ibm/wala/shrikeBT/Decoder.java
```diff
@@ -1108,9 +1108,7 @@ public final void decode() throws InvalidBytecodeException {
         }
       }
     } else {
-      for (int i = 0; i < handlers.length; i++) {
-        handlers[i] = noHandlers;
-      }
+      Arrays.setAll(handlers, i -> noHandlers);
```
msridhar: `Arrays.fill`?

msridhar:
Sorry some comments got duplicated as I accidentally closed the tab in the middle of the review 😐

com.ibm.wala.util/src/com/ibm/wala/util/intset/IntSetUtil.java
```diff
@@ -56,9 +57,7 @@ public static MutableIntSet make(int[] initial) {
   public static IntSet make(Set<Integer> x) {
     int[] vals = new int[x.size()];
     Iterator<Integer> vs = x.iterator();
-    for (int i = 0; i < vals.length; i++) {
-      vals[i] = vs.next();
-    }
+    Arrays.setAll(vals, i -> vs.next());
```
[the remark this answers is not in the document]
liblit: Unfortunate indeed. Rewrote as `int[] vals = x.stream().mapToInt(Integer::intValue).toArray();` instead.

liblit:
All suggested API improvements applied. Thanks for those good suggestions, @msridhar!

`callGraph` performance test in progress.

liblit:
To assess performance impact, I performed 20 runs of each of the 58 `com.ibm.wala.core.tests.callGraph.*` tests with and without the proposed changes. I then used an two-sided, unpaired Student’s _t_-test to judge whether the mean test execution times changed. I applied this test separately to each of these 58 tests, meaning that we have 58 different decisions as to whether there was a performance degradation.

Of those 58 assessments for 58 tests, only 3 show mean execution time differences that satisfy the usual _p_ < 0.5 threshold:

* the proposed changes appear to slow down `com.ibm.wala.core.tests.callGraph.CallGraphTest.testRecursion`, growing its mean execution time from 375.85 milliseconds to 392.65 milliseconds.

* the proposed changes appear to speed up `com.ibm.wala.core.tests.callGraph.AcyclicCallGraphTest.testNList`, shrinking its mean execution time from 115.25 milliseconds to 111.9 milliseconds.

* the proposed changes appear to slow down `com.ibm.wala.core.tests.callGraph.CPATest.cpaTest2`, growing its mean execution time from 112.8 milliseconds to 116.1 milliseconds.

Given that we performed 58 separate assessments, it is to be expected that a few might appear to show statistically significant changes simply by random chance: this is the standard risk of performing many assessments. Note as well that the three tests listed above are all quite speedy, meaning that even a small shift may appear statistically significant. Informally, looking across all of the tests, changes in mean execution times appear to be small and as likely to be speedups as slowdowns.

My conclusion is that the proposed changes do not significantly harm overall WALA performance.

PR #486, opened 2019-05-09
#485 make reading manifest optional in AnalysisScopeReader

### LINKED ISSUE #485: Make reading manifest file optional in AnalysisScopeReader
`AnalysisScopeReader` will read the manifest file of jar file and add the `Class-Path` entries into scope, by default. However, the class-path specification is for use when executing Java applications via the "java -jar" command. It doesn't apply to webapps (WARs). Actually, FileNotfoundException will throw up when I use WALA to analyze some large-scale WARS.

And I am working on to solve this problem by adding some extra method into `AnalysisScopeReader`. When it is ready, I will create a PR.

coder-chenzhi:
To resolve #485 , I have introduced two extra method for back-compatible. Please review it.


msridhar:
You need to run `./gradlew goJF` locally to format the code correctly with Google Java Format.  You can also run `./gradlew installGitHooks` to have GJF run locally on every commit.

msridhar:
Can we add a regression test for this new code?  Like package some classes into a `.war` file during testing and then try to load it?

coder-chenzhi:
I have ran `./gradlew goJF` to  format the code and pushed the update to current branch.

We can use some existing simple war projects as test cases, such as [sample1](https://tomcat.apache.org/tomcat-7.0-doc/appdev/sample/) and [sample2](https://github.com/efsavage/hello-world-war). However, it seems that WALA does not support to load war files directly. In fact, I manually extract classes and libs under `WEB-INF` from war files and add them to `AnalysisScope` when performing the analysis. Is it necessary to add some helper classes to ease the analysis of war files?

msridhar:
@coder-chenzhi what is your usage scenario?  We used to have more J2EE support in WALA many years back but it was unmaintained and got removed.  Beyond handling WAR files, usually you need to do things like model J2EE entrypoints, callbacks, etc.  For your use case, is just including WAR files in the AnalysisScope good enough?  If so, maybe we should expand this PR to allow that directly

liblit:
There has been no new discussion about this PR since October 7, 2020. @coder-chenzhi and @msridhar, do you still expect to reach consensus that this PR is good and ready to merge? Or has this PR been abandoned, in which case we should just close it?

PR #487, opened 2019-05-10
Couple of fixes for build scripts

msridhar:
* Update `change-version.py` to also update `gradle.properties`
* Fix `needsSigning()` in `gradle-mvn-push.gradle` to match on `name` rather than using `hasTask` (which requires the fully-qualified task name)


msridhar:
Discovered when attempting to cut a release

PR #488, opened 2019-05-10
Replace anonymous types with lambdas where possible

liblit:


liblit:
@msridhar: no particular need for this to be merged before the release you're preparing.

PR #491, opened 2019-05-11
Do javadoc upload at the same time we upload snapshot builds

msridhar:
Moving another part of the build from Maven to Gradle.  This also adds an `aggregatedJavadocs` task (from [here](https://discuss.gradle.org/t/best-approach-gradle-multi-module-project-generate-just-one-global-javadoc/18657/21)) to create the aggregate Javadocs.

Confirmed that things are working on this build:

https://travis-ci.org/wala/WALA/jobs/531261607


deploy_snapshot_and_javadoc.sh
```diff
@@ -24,4 +25,26 @@ else
...
   ./gradlew clean uploadArchives
   echo "Snapshot deployed!"
+
+  echo "Uploading javadoc..."
+
+  ./gradlew aggregatedJavadocs
+
+  {
+    set +x
+    cd $HOME
```
liblit: This script tries to operate on the `~/javadoc` directory of whoever runs it, even if that directory has nothing to do with WALA. If that directory already exists, then hopefully the script will fail and stop at the `git clone` step, but this all makes me very nervous. Even if this script should only be run within an Travis CI job, the mere possibility of running `git rm -rf --quiet *` on a directory that might be outside of the WALA tree seems like an unnecessary risk.

Instead of working under `$HOME`, let’s change the script to work entirely within the `build` subdirectory of the current working directory when the script was launched. That will be the top-level `build` subdirectory of the WALA tree itself, which is where Gradle naturally puts all built artifacts. For example, we could check out the `wala/javadoc` project into `build/javadoc-for-gh-pages` and avoid any risk of damaging user files outside of the WALA tree.

Another option would be to simply use the `build/docs/javadoc` directory that the `aggregatedJavadocs` target will already have created. That directory is already exactly what we want to commit as the new `wala/javadoc`. The only thing that `build/docs/javadoc` is missing is Git metadata to treat it as a repository. So let’s change the script to add that, as in:

``` sh
# disclaimer: I have not actually tested this proposed sequence of commands!

./gradlew aggregatedJavadocs

cd build/docs/javadoc
git init
git commit -m "Latest javadoc on successful travis build $TRAVIS_BUILD_NUMBER auto-pushed to gh-pages" .  # don't overlook the important "." at the end of this line!

set +x
git remote add origin https://${GH_TOKEN}@github.com/wala/javadoc
set -x
git push --force
```

This avoids doing a `git clone` that we will simply blow away. On the other hand, `--force` is never to be taken lightly.

deploy_snapshot_and_javadoc.sh
```diff
@@ -24,4 +25,26 @@ else
...
+    git config --global user.email "travis@travis-ci.org"
+    git config --global user.name "travis-ci"
+    git clone --quiet https://${GH_TOKEN}@github.com/wala/javadoc > /dev/null
+    set -x
+  }
+  cd javadoc
+  git rm -rf --quiet *
```
liblit: Prefer `./*` instead of `*` so that filenames starting with `-` will not be treated as options. We currently have no files with names like that, but it is wise to be paranoid when recursively, forceably removing files.

deploy_snapshot_and_javadoc.sh
```diff
@@ -24,4 +25,26 @@ else
...
+  echo "Uploading javadoc..."
+
+  ./gradlew aggregatedJavadocs
+
+  {
+    set +x
+    cd $HOME
+    git config --global user.email "travis@travis-ci.org"
+    git config --global user.name "travis-ci"
+    git clone --quiet https://${GH_TOKEN}@github.com/wala/javadoc > /dev/null
```
liblit: Enclose either the entire URL or just `$GH_TOKEN` in double quotes in case of oddball characters in `$GH_TOKEN`. Special characters seem unlikely here, but we should be extra careful with anything coming into a script from its environment.

deploy_snapshot_and_javadoc.sh
```diff
@@ -24,4 +25,26 @@ else
...
+  }
+  cd javadoc
+  git rm -rf --quiet *
+  cp -Rf $HOME/build/wala/WALA/build/docs/javadoc/* .
+  git add -f .
+  git commit -m "Latest javadoc on successful travis build $TRAVIS_BUILD_NUMBER auto-pushed to gh-pages"
+  git push > /dev/null
+
+  echo -e "Published Javadoc to gh-pages.\n"
+
```
liblit: Remove unnecessary trailing spaces on this line.

msridhar:
Good feedback, @liblit.  The WALA javadoc repo has gotten quite unwieldy:

```
$ git clone https://github.com/wala/javadoc
Cloning into 'javadoc'...
remote: Enumerating objects: 1767470, done.
remote: Total 1767470 (delta 0), reused 0 (delta 0), pack-reused 1767470
Receiving objects: 100% (1767470/1767470), 786.99 MiB | 6.81 MiB/s, done.
Resolving deltas: 100% (1617648/1617648), done.
Checking out files: 100% (5756/5756), done.
```

We should not care about storing the history of generated javadoc, as we can re-generate it from the previous source code if needed.  So, I'm going to try your suggested force-push strategy.

deploy_snapshot_and_javadoc.sh
```diff
@@ -8,7 +8,7 @@
 
 SLUG=wala/WALA
 JDK=oraclejdk8
-BRANCH=master
+BRANCH=ms/agg-javadoc-gradle
```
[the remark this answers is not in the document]
msridhar: Yes this was just for testing, and I will change it again to test.  We should be sure this is `master` before landing.

build.gradle
```diff
@@ -322,3 +322,20 @@ tasks.register('installGitHooks', Copy) {
 tasks.register('downloads') {
 	dependsOn allprojects*.tasks*.withType(VerifiedDownload)
 }
+
+task aggregatedJavadocs(type: Javadoc, description: 'Generate javadocs from all child projects as if it was a single project', group: 'Documentation') {
```
msridhar: @liblit do i need to do some kind of lazy configuration thing here?

msridhar:
@liblit this is ready for review again

build.gradle
```diff
@@ -322,3 +322,20 @@ tasks.register('installGitHooks', Copy) {
...
+
+	subprojects.each { proj ->
+		proj.tasks.withType(Javadoc).each { javadocTask ->
+			source += javadocTask.source
+			classpath += javadocTask.classpath
+			excludes += javadocTask.excludes
+			includes += javadocTask.includes
+		}
+	}
+}
```
liblit: Add trailing newline.

deploy_snapshot_and_javadoc.sh
```diff
@@ -24,4 +25,29 @@ else
...
+    git remote add origin "https://${GH_TOKEN}@github.com/wala/javadoc"
+    set -x
+  }
+
+  # we can force push here since we don't care about maintaining javadoc history for every commit
+  git push origin master --force
+
+  echo -e "Published Javadoc to gh-pages.\n"
+
+  cd $TRAVIS_BUILD_DIR
```
liblit: Not sure what purpose this serves given that no further commands will be run.

Are you thinking that this script might be used in a `source` or `.` command rather than being run in its own shell? If so, then a more robust approach would be to enclose everything from the `cd build/docs/javadoc` line down to the `echo` in parentheses so that it runs in a subshell.

build.gradle
liblit: Not absolutely required, but yes, using the [task configuration avoidance APIs](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html) would be preferred. Replace the `task` line with the following:

``` gradle
tasks.register('aggregatedJavadocs', Javadoc) {
	description = 'Generate javadocs from all child projects as if it was a single project'
	group = 'Documentation'
```

deploy_snapshot_and_javadoc.sh
liblit: Not yet addressed as of 1c2601b7a259985932d8f0bc4a080cf8292c7a86.

liblit:
Forgot to include in my review:

Consider renaming `deploy_snapshot_and_javadoc.sh` to `deploy-snapshot-and-javadoc.sh`. Nearly all other files in this directory with multi-word names use hyphens instead of underscores:

* `change-version.py`
* `dependent-projects-trigger.sh`
* `gradle-mvn-push.gradle`
* `README-Gradle.md`
* `revert-launchers.sh`

The only exception is `After Importing WALA Into Eclipse.launch`, which is reasonable because its name is used directly in Eclipse’s launcher list UI.

PR #495, opened 2019-05-13
Relocate some jar files and simplify "clean*" tasks

liblit:
Hoist some jar files out of `com.ibm.wala.core.testdata/build`. The `com.ibm.wala.core.testdata/build.properties` file that Eclipse uses expects to find these jar files directly in the `com.ibm.wala.core.testdata` directory, not in the `com.ibm.wala.core.testdata/build` subdirectory.

Let Gradle generate some `clean*` tasks automatically. Previously we had to (or thought we had to) create these ourselves, but Gradle actually creates them correctly on its own.

### DIFF
--- a/com.ibm.wala.core.testdata/build.gradle
+++ b/com.ibm.wala.core.testdata/build.gradle
@@ -176,10 +176,6 @@ tasks.register('downloadJavaCup', VerifiedDownload) {
 	checksum '2bda8c40abd0cbc295d3038643d6e4ec'
 }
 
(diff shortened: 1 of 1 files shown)


liblit:
@msridhar asked:
> Do we need to update `.gitignore` to ignore new files due to this change?

Attentive question! No such updates are needed. The new jar files’ locations are exactly where Maven builds have been putting them for a long time. These are [already ignored](https://github.com/wala/WALA/blob/b490c792880803db42e634eab10a80435c812fac/.gitignore#L33).

PR #498, opened 2019-05-13
More cleanup of lambda-related code

msridhar:
Some additional code cleanup and documentation


msridhar:
> The code is clean, but I don’t know WALA well enough to speak to its correctness. The requested additional review from @juliandolby is a good idea.

This one should just be a refactor so I'm not too concerned about correctness.  The next one I'll put up is trickier 😄

PR #501, opened 2019-05-17
Port change-version script to Bash

liblit:
I was feeling an urge to validate our Python code somehow, such as via flake8 or pychecker.  However, we only had one Python script in the entire project, and mostly what it did was run other subcommands. The easier path to validation, then, was to port this script from Python to Bash so that it can be checked by our existing `shellCheck` task.  The Bash version ends up being a few lines shorter, too.

Also, remove redundant version string in `build.gradle` We can pick this up from `gradle.properties`, so there’s no reason to record it twice.


change-version.sh
```diff
@@ -0,0 +1,30 @@
...
+  exit 1
+fi
+
+replace() {
+  pattern=$1
+  replacement=$2
+  shift 2
+
+  find . \( "$@" \) -print0 |
+    xargs --null --max-args=1 sed --in-place --expression="s/$pattern/$replacement/g"
```
msridhar: So I used `perl` before because `sed` on Mac and Linux have some subtle differences (GNU vs. BSD I think).  It would be good to test this command on both platforms.  Let me know if you need help testing on Mac

liblit: Thanks for bringing this up, @msridhar. Yes, certain GNUisms will need to be changed. I will push a revised variant when ready, at which time I will greatly appreciate some macOS testing help from you. (Not yet, thought!)

By the way, the easy part of the changes will be to replace `--long-options` with single-character flags. The harder part is using `sed` for in-place rewriting, since the Linux and macOS flags for this are mutually incompatible. For that I will likely revert to using `perl` as you were doing before.

liblit:
OK, I believe that the script is now Linux/macOS portable. However, I have no macOS box to test this on. @msridhar, your help with that would be much appreciated.

msridhar:
I ran on my Mac and the script exited successfully and seemed to do the right thing.  For further testing I suggest the following.

* Run `./change-version.sh 1.5.4-SNAPSHOT 1.5.4`, push the result to Travis, and see that it passes.
* Run `./change-version.sh 1.5.4-SNAPSHOT 1.5.5-SNAPSHOT`, push the result to Travis, and see that it passes.

That will give us confidence that everything is working on Linux and Mac.  It may seem finicky, but I've run into a lot of flaky problems in the past when these version numbers became inconsistent.

liblit:
@msridhar requested:
> * Run `./change-version.sh 1.5.4-SNAPSHOT 1.5.4`, push the result to Travis, and see that it passes.

Done. [It passes.](https://travis-ci.org/liblit/WALA/builds/534349240)

> * Run `./change-version.sh 1.5.4-SNAPSHOT 1.5.5-SNAPSHOT`, push the result to Travis, and see that it passes.

Done. [It passes.](https://travis-ci.org/liblit/WALA/builds/534354930)

PR #506, opened 2019-05-20
Convert to java-library Gradle plugin

liblit:
The `java-library` plugin distinguishes dependencies needed only for a library’s internal implementation from dependencies that form part of a library’s external API. A library’s implementation dependencies are not transitively made available to other components that depend on that library. This keeps classpaths shorter and helps reduce recompilation. See [the `java-library` plugin documentation](https://docs.gradle.org/current/userguide/java_library_plugin.html) for more information.

I took an empirical approach to determining which dependencies are implementation dependencies and which are API dependencies. In general, one should prefer the more restrictive implementation categorization wherever possible. What I’ve done here is to use implementation dependencies everywhere I could, and only switch to an API dependency where required for compilation to succeed. For example, if class `A` in library `Alpha` is a subclass of class `B` in class `Beta`, then code that tries to create a new `A` will fail to compile unless `Beta` was also brought along as an API dependency of `Alpha`.

In all cases where an API dependency was required, I have included an explanation identifying at least one specific reason why, such as `public class A extends class B`. Such explanations are optional, but seem like a good idea and are even directly supported by Gradle’s dependency-declaration API.

It is definitely possible that some current implementation dependencies should really be API dependencies. This would most likely happen if third-party code uses WALA APIs in ways that WALA’s own tests do not. Returning to the `A`/`Alpha`/`B`/`Beta` example above, if no WALA test actually uses class `A`, then we would we unaware of the need for `Alpha` to propagate an API dependency on `Beta`. A third-party application or library code can work around such a misclassification by explicitly adding its own dependency on Beta, so this is not a show-stopping error. It just makes WALA a bit less convenient to use by requiring more explicit listing of dependencies that previously would have been picked up implicitly. We can certainly reclassify implementation dependencies as API dependencies wherever appropriate in the future.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,7 +13,7 @@ plugins {
 	id 'com.github.sherter.google-java-format' version '0.8'
 	id 'de.set.ecj' version '1.4.1' apply false
 	id 'de.undercouch.download'
-	id 'nebula.lint' version '9.2.0'
+	id 'nebula.lint' version '11.5.0'
 	id 'nebula.source-jar' version '9.5.0' apply false
 }
 
@@ -192,10 +192,10 @@ allprojects {
 	gradleLint {
 		rules = ['all-dependency']
 		excludedRules = [
-			'duplicate-dependency-class',
-			'transitive-duplicate-dependency-class',
-			'unused-dependency',
-			]
+				// <https://github.com/nebula-plugins/gradle-lint-plugin/issues/203>
+				'undeclared-dependency',
+				'unused-dependency',
+		]
 	}
 }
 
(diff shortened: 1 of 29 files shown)


liblit:
Advice for reviewers: these commits include many indentation changes. I recommend ignoring whitespace-only changes while reviewing this request. In GitHub’s web UI, click “Diff settings ▾” near the top right to reveal a “☑ Hide whitespace changes” toggle.

msridhar:
My main concern here is that Maven artifacts still get cut correctly.  @liblit can you run `./gradlew install` before and after this change, and see if the generated `pom.xml` files for the different projects still look similar?  @juliandolby you may want to check that whatever projects you have that depend on WALA Maven artifacts installed locally still work correctly

liblit:
@msridhar requested:
> @liblit can you run `./gradlew install` before and after this change, and see if the generated `pom.xml` files for the different projects still look similar?

Done, but I think the proper things to compare are the `com.ibm.wala.*-1.5.4-SNAPSHOT.pom` files installed under `~/.m2`. Right?

We have ten such files to compare. Ignore changes in dependency _order_, which I think is irrelevant. The only other changes are that many dependency scopes change from `<scope>compile</scope>` to `<scope>runtime</scope>`. Using the `java` plugin, we have a total of 40 dependencies, all `compile`-scoped. Using the `java-library` plugin, we have just 4 `compile`-scoped dependencies and 36 `runtime`-scoped dependencies.

An `implementation` Gradle dependency becomes a `runtime`-scoped Maven dependency: something that should be included in the classpath when _running_ code that uses a given library, but that need not be included in the classpath when _compiling_ code that uses the same library. Conversely, an `api` Gradle dependency becomes a `compile`-scoped Maven dependency: something that should be included in the classpath both when compiling _and_ when running code that uses a given library.

[More information on Maven dependency scopes](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope).

PR #511, opened 2019-05-26
Simpler null-safe equality check using Objects.equals

liblit:


com.ibm.wala.core.tests/src/com/ibm/wala/core/tests/util/WalaTestCase.java
```diff
@@ -110,6 +111,6 @@ protected static String getClasspathEntry(String elt) {
   }
 
   protected <T> void assertEquals(T x, T y) {
-    assert x == null ? y == null : x.equals(y) : "Expecting " + x + ", but got " + y;
+    assert Objects.equals(x, y) : "Expecting " + x + ", but got " + y;
```
msridhar: Actually, since we're touching this code... 😄 This method should be deleted and replaced with calls to JUnit's [`assertEquals` method](https://junit.org/junit4/javadoc/latest/org/junit/Assert.html#assertEquals(java.lang.Object,%20java.lang.Object)).  Can we do that?

liblit: Done.

PR #531, opened 2019-07-20
Suppress suggestion to use Map.computeIfAbsent

liblit:
`Map.computeIfAbsent` can do the same thing we need here, but might only perform one `Map` lookup instead of two (if implemented efficiently).


msridhar:
I think @juliandolby just switched away from this API for the browser scenario. See https://github.com/wala/WALA/commit/e13c1a8dc236926406bf1b6f3f0c073d3d98c995

liblit:
Oops! Thanks for catching this, @msridhar. I have revised my commit to suppress the suggested change rather than reapply what @juliandolby intentionally removed.

PR #535, opened 2019-08-13
Use Ubuntu 14.04 (Trusty) for most Travis CI Linux tests

liblit:
Travis CI made Ubuntu 16.04 (Xenial) the default in late April: <https://changelog.travis-ci.com/xenial-as-the-default-build-environment-99476>, <https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment>. However, we are only testing under Xenial in a few limited configurations.  For the most part, WALA still requires Java 8, and on Travis CI that requires Ubuntu 14.04 (Trusty).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,10 +47,12 @@ matrix:
       language: java
       jdk: openjdk11
     - os: linux
+      dist: trusty
       env: BUILD_SYSTEM=gradle
       language: java
       jdk: oraclejdk8
     - os: linux
+      dist: trusty
       env: BUILD_SYSTEM=maven
       language: android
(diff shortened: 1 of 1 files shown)


msridhar:
It looks like `openjdk8` is the default on Xenial:

https://docs.travis-ci.com/user/reference/xenial/#jvm-clojure-groovy-java-scala-support

Is that good enough?  Or does WALA for some reason actually require `oraclejdk8`?  I suggest we actually switch to `openjdk8` everywhere and get that to work.

liblit:
OpenJDK sounds good to me, and passes all tests. Implemented in new pull request #536.

PR #539, opened 2019-08-17
Miscellaneous cleanups for IntelliJ IDEA 2019.2

liblit:
Clean up various warnings revealed by inspections under IntelliJ IDEA 2019.2.

Includes suppression of some lambda suggestions that break Android builds, to avoid reversion of b51d9c38192a0d6dcf1b3f94805b79bfa92459a7, as [suggested by @msridhar](https://github.com/wala/WALA/commit/b51d9c38192a0d6dcf1b3f94805b79bfa92459a7#commitcomment-34450011).

### DIFF
--- a/com.ibm.wala.cast.java.test.data/src/javaonepointfive/MoreOverriddenGenerics.java
+++ b/com.ibm.wala.cast.java.test.data/src/javaonepointfive/MoreOverriddenGenerics.java
@@ -38,6 +38,7 @@
 package javaonepointfive;
 
 public class MoreOverriddenGenerics {
+	@SuppressWarnings("InnerClassMayBeStatic")
 	class Super<T> {
 		T x;
 
--- a/com.ibm.wala.core.tests/src/com/ibm/wala/core/tests/ir/AnnotationTest.java
+++ b/com.ibm.wala.core.tests/src/com/ibm/wala/core/tests/ir/AnnotationTest.java
@@ -37,7 +37,7 @@ import java.util.Collection;
 import java.util.Set;
 import org.junit.Test;
 
-public class AnnotationTest extends WalaTestCase {
+public abstract class AnnotationTest extends WalaTestCase {
 
   private final IClassHierarchy cha;
 
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -2,6 +2,7 @@
 /compiler.xml
 /libraries/*.xml
 /modules.xml
+/shelf/
 /tasks.xml
 /usage.statistics.xml
 /workspace.xml
--- a/com.ibm.wala.cast.java.ecj/src/com/ibm/wala/cast/java/translator/jdt/JDTJava2CAstTranslator.java
+++ b/com.ibm.wala.cast.java.ecj/src/com/ibm/wala/cast/java/translator/jdt/JDTJava2CAstTranslator.java
@@ -4023,7 +4023,7 @@ public abstract class JDTJava2CAstTranslator<T extends Position> {
     }
   }
 
-  private class AssignmentContext extends DelegatingContext {
+  private static class AssignmentContext extends DelegatingContext {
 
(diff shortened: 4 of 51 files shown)


com.ibm.wala.cast.java.test.data/src/javaonepointfive/MoreOverriddenGenerics.java
```diff
@@ -38,15 +38,15 @@
 package javaonepointfive;
 
 public class MoreOverriddenGenerics {
-	class Super<T> {
+	static class Super<T> {
```
msridhar: Do we really want to be applying these checks and fixes to input test data?

com.ibm.wala.core.tests/src/com/ibm/wala/core/tests/ir/AnnotationTest.java
```diff
@@ -37,7 +37,7 @@
 import java.util.Set;
 import org.junit.Test;
 
-public class AnnotationTest extends WalaTestCase {
+public abstract class AnnotationTest extends WalaTestCase {
```
msridhar: Why make this (and other test classes) abstract?

com.ibm.wala.cast.java.test.data/src/javaonepointfive/MoreOverriddenGenerics.java
liblit: I mistakenly thought that this code was test infrastructure, not test input. WALA is Java code for analyzing Java code, so sometimes one gets disoriented. :blush:

I will revert this and suppress the IDE's suggestion to change it.

com.ibm.wala.core.tests/src/com/ibm/wala/core/tests/ir/AnnotationTest.java
liblit: These specific test classes cannot be instantiated by typical JUnit test runners. These are apparently intended as base classes for real tests. IntelliJ IDEA 2019.2 adds an inspection for this, called "Unconstructable JUnit TestCase":

> Reports non-abstract JUnit test cases which do not expose a public no-arg constructor or a public constructor which takes a single string as an argument. Such test cases will be unrunnable by most JUnit test runners, including IDEA's.

Declaring such classes `abstract` resolves the inspection warnings and IMHO accurately reflects the intent that these classes only be used as base classes, never instantiated directly.

com.ibm.wala.cast.java.test.data/src/javaonepointfive/MoreOverriddenGenerics.java
```diff
@@ -38,6 +38,7 @@
 package javaonepointfive;
 
 public class MoreOverriddenGenerics {
+	@SuppressWarnings("InnerClassMayBeStatic")
```
msridhar: Any way to just disable diagnostics on the whole module, rather than adding suppressions?  Not a huge deal, but since it's a source front end in principle even annotations could affect the behavior of analyses

com.ibm.wala.core.tests/src/com/ibm/wala/core/tests/ir/AnnotationTest.java
msridhar: 👍

com.ibm.wala.cast.java.test.data/src/javaonepointfive/MoreOverriddenGenerics.java
liblit: We can define a new IntelliJ IDEA file scope that contains just files used as test inputs. Then we can reduce the severity level of individual inspections to "No highlighting, only fix" for files in that "test inputs" scope. I think that will have the desired effect.

If it's OK with you, I prefer to take that on as a separate task after merging this pull request.

msridhar:
Sure separate PR is fine

PR #542, opened 2019-08-26
Change a questionable integer division to double

liblit:
Near as I can tell, `size` here is never larger than `MAX_SIZE`, which means that this integer division would almost always be 0.  Presumably it was intended to be a floating-point ratio measuring how close `size` is to `MAX_SIZE`.


com.ibm.wala.util/src/com/ibm/wala/util/intset/MultiModalIntVector.java
```diff
@@ -81,8 +81,8 @@ float getGrowthFactor() {
    */
   float getGrowthFactor(int size) {
     if (CURRENT_GROWTH_RATE >= MINIMUM_GROWTH_FACTOR) {
```
msridhar: Since we're editing this code, can we rename `CURRENT_GROWTH_RATE` to `currentGrowthRate`, to make it clearer it's an instance variable?

liblit: Renamed.

PR #587, opened 2019-11-26
Stop using Eclipse P2 repositories

liblit:
Many core Eclipse libraries are now cross-published directly to Maven Central. The WST JSDT libraries are not cross-published there, but I've found a source for them in a different public Maven repository. So now we can get everything we need from Maven repositories rather than downloading and converting P2 repositories.

Getting Eclipse libraries from proper Maven repositories should make the first build in a clean WALA tree faster and friendlier. We still need to download the Eclipse libraries, of course. But those downloads can now be concurrent with other Gradle build activity, such as compiling and running tests for non-Eclipse-based pieces of WALA. We also get better download progress reporting, proper Gradle caching, etc.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,7 @@ buildscript {
 }
 
 plugins {
-	id 'com.diffplug.gradle.p2.asmaven' version '3.18.1'
+	id 'com.diffplug.gradle.eclipse.mavencentral' version '3.18.1' apply false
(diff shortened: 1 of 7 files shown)


liblit:
Hat tip to @nedtwigg for [making me aware that this was now an option] (https://github.com/diffplug/goomph/issues/106#issuecomment-558002572).

PR #596, opened 2019-12-06
Explore GitHub Actions for continuous integration

liblit:
GtiHub Actions could potentially replace Travis CI as our continuous integration engine. So far I’m not actually sure which I prefer. Neither is dramatically faster than the other, for example. However, if GitHub Actions had fewer download timeouts then Travis CI, that would be a clear win. I propose that we use both for a while, and see if a winner emerges.

This initial configuration of GitHub Actions for continuous integration comprises three jobs:

1. Full Gradle build and test using JDK 1.8 on recent Ubuntu
2. Full Gradle build and test using JDK 1.8 on recent macOS
3. Java compilation only using ECJ on recent Ubuntu

There is certainly more than we could do, such as adding older OS releases, adding Windows, building using Maven, etc. But this is a sensible start.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,0 +1,55 @@
+name: Continuous integration
+on:
+  - pull_request
+  - push
+jobs:
+  build:
+    name: Build and test
+    strategy:
+      matrix:
+        os:
+          - ubuntu-latest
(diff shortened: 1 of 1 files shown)


liblit:
Taking @juliandolby’s silence as tacit approval.

PR #598, opened 2019-12-07
Fix com.ibm.wala.cast.java.ecj Eclipse deps

msridhar:
The Eclipse dependencies of `com.ibm.wala.cast.java.ecj` were inconsistent, leading to the following inscrutable error when running the `SourceDirCallGraph` driver:
```
Exception in thread "main" com.ibm.wala.ipa.cha.ClassHierarchyException: factory.getLoader failed java.lang.SecurityException: class "org.eclipse.core.runtime.ICoreRunnable"'s signer information does not match signer information of other classes in the same package
        at com.ibm.wala.ipa.cha.ClassHierarchy.<init>(ClassHierarchy.java:291)
        at com.ibm.wala.ipa.cha.ClassHierarchy.<init>(ClassHierarchy.java:203)
        at com.ibm.wala.ipa.cha.ClassHierarchyFactory.make(ClassHierarchyFactory.java:81)
        at com.ibm.wala.ipa.cha.ClassHierarchyFactory.make(ClassHierarchyFactory.java:67)
        at com.ibm.wala.cast.java.ecj.util.SourceDirCallGraph.main(SourceDirCallGraph.java:63)
```
This update the `org.eclipse.core.runtime` dependence to make things consistent.  We also make the `SourceDirCallGraph` driver run as part of running ECJ tests.

@liblit I'm not sure this is the right fix.  Should we be using the `eclipseMavenCentral` stuff in `com.ibm.wala.cast.java.ecj/build.gradle` instead?


liblit:
Obscure indeed! Nice fix.

But yes, we should prefer to use `eclipseMavenCentral` for grabbing all Eclipse libraries if we want them to all be at a mutually-consistent version.

msridhar:
Switched to using `eclipseMavenCentral` now

com.ibm.wala.cast.java.ecj/build.gradle
```diff
@@ -1,9 +1,21 @@
...
+}
+
 sourceSets.main.java.srcDirs = ['src']
 
+eclipseMavenCentral {
+	release eclipseVersion, {
+		[
+				'org.eclipse.core.runtime',
+				'org.eclipse.jdt.core',
+		].each { dep 'implementation', it }
```
liblit: Prefer method reference as found in a few other `build.gradle` files:

``` gradle
].each it.&implementation
```

msridhar:
@cyw3 what is the exact error you are getting?  And are you running the same SourceDirCallGraph I was using?

https://github.com/wala/WALA/blob/10ed693e0ca75ea8612034a48488ffc092f77d96/com.ibm.wala.cast.java.ecj/src/main/java/com/ibm/wala/cast/java/ecj/util/SourceDirCallGraph.java

PR #599, opened 2019-12-07
IntelliJ IDEA 2019.3 inspection updates

liblit:
These changes collectively restore us to the desired state of having zero inspection diagnostics when applying the “No Back-Sliding” inspection profile to the “Aggressively Linted” inspection scope under IntelliJ IDEA 2019.3.


config/hooks/pre-commit
```diff
@@ -1,10 +1,17 @@
...
+#!/bin/bash -ex
 
 REPO_ROOT_DIR="$(git rev-parse --show-toplevel)"
 
-files=$( (git diff --cached --name-only --diff-filter=ACMR | grep -Ei '\.java$') || true)
-if [ -n "$files" ]; then
-    comma_files=$(echo "$files" | paste -s -d , -)
-    "${REPO_ROOT_DIR}/gradlew" goJF -DgoogleJavaFormat.include="$comma_files" >/dev/null 2>&1
-    git add $(echo "$files" | paste -s -d ' ' -)
+mapfile -t files < <(git diff --cached --name-only --diff-filter=ACMR | grep -Ei '\.java$')
```
msridhar: Add a comment on what this command does?  I've seen many bash scripts but I didn't know about `mapfile`

liblit: Added:
https://github.com/wala/WALA/blob/31eef06302f59940f743858a5891302c1ade585b/config/hooks/pre-commit#L5-L6

msridhar: We still run with `-ex` but we don't have an `|| true` with the new `mapfile` command.  Will the script still succeed even if the `grep` command fails?

msridhar: Also, on Mac:
```shell
$ /bin/bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin18)
Copyright (C) 2007 Free Software Foundation, Inc.
```
So I'm not sure the `mapfile` command will work.

liblit: Even if `pipefail` were turned on (which it is not), failure in the process-substitution subcommand would not propagate into an error status in the main flow of the script, and therefore would not terminate the script under `set -e`.

But yes, the archaic Bash version on macOS is going to be a problem. I will need to use an alternate approach. Drat.

msridhar: If a workaround is a huge pain we can require a more recent Bash version (which can easily be installed using `homebrew`).  But hopefully we can find a way that works on the old bash version.

liblit: Believed fixed for macOS in 202e34367561483dbd2a13c23708ff5971a4b6f3, but untested by me.

msridhar: Ok, I will test it out later just to make sure

msridhar:
@liblit Have you tested the pre commit script? I’d like it to be tested before we land as I rely on it. I can test later today probably. If you’ve tested on Linux that’s good enough

liblit:
I have not tested the latest version of `pre-commit`. Feel free to wait until you have tested it to your satisfaction before approving.

config/hooks/pre-commit
```diff
@@ -1,10 +1,22 @@
-#!/bin/sh -e
+#!/bin/bash -ex
```
msridhar: We should just do `-e`, not `-ex`, as we don't want the individual commands printed during each commit (at least I don't).

config/hooks/pre-commit-stub
```diff
@@ -9,5 +9,6 @@ REPO_ROOT_DIR="$(git rev-parse --show-toplevel)"
 PRE_COMMIT_SCRIPT="${REPO_ROOT_DIR}/config/hooks/pre-commit"
 
 if [ -f "$PRE_COMMIT_SCRIPT" ]; then
-    . "$PRE_COMMIT_SCRIPT"
+  # shellcheck source=pre-commit
+  . "$PRE_COMMIT_SCRIPT"
```
msridhar: Unfortunately, `. "$PRE_COMMIT_SCRIPT"` does not work any more, as that tries to `source` the script in `sh` (hence `bash` never runs).  We need just `"$PRE_COMMIT_SCRIPT"` instead, to respect the sha-bang there.

liblit: Oops, good that you tested! Fixed in 4624c4d70050defcd2a2ecf59d2610562e1d0837.

config/hooks/pre-commit
liblit: Enacted in 72d981a09679c859b7b059d2ca2ae7bea75e4303.

liblit:
Yeah, I see what you are saying about the confusing need to reinstall the script. Do you want me to cut the script changes out of this pull request entirely? I can do that if you wish.

msridhar:
Is there any way inside `pre-commit` to check if the shell is not `bash`, and if so, re-spawn the script in `bash` and exit? That would give us backward compatibility though it’s kind of gross

liblit:
> Is there any way inside `pre-commit` to check if the shell is not `bash`, and if so, re-spawn the script in `bash` and exit?

Two tricky things there: (1) distinguishing `bash` running as `bash` from `bash` running as `sh`, and (2) determining the name of the script being sourced as distinguished from `$0` (the script originally run). I don't know how to do either offhand; it would take some experimentation.

We have a bit of a mess here. Presumably the intent of `pre-commit-stub` was to let WALA maintainers modify `pre-commit` in the future without developers having to re-copy these modifications into `.git/hooks`. So instead we are in a situation where WALA maintainers can no longer modify `pre-commit` at all, at least not in the way we would like.

I am inclined to punt on this whole thing: revise this pull request to exclude the `pre-commit` work entirely, and configure IntelliJ IDEA's inspections to silently ignore the `pre-commit` problems it thinks it found. Those problems would only manifest if the WALA tree were in a directory with unusual characters in it anyway; this is more of a concern in theory than in practice.

msridhar:
Is there any way to fix the lint warnings without switching to `bash`, sticking with `sh` instead? Otherwise, yes, probably backing out the changes and suppressing warnings is our best option here.

liblit:
I have pushed five commits trying to make `pre-commit` robust and proper. I am done with this script. If there is a way for this script to be safe and robust using only `sh`, I am not presently inclined to look for it.

liblit:
Bash porting attempt for `pre-commit` reverted in f497ecc50b722586f7529df27522ec3f85ff4f66. shellcheck warning suppressed in b4a7ba4bad9a2654fdd5cac1066a271e29dc6172

msridhar:
@liblit if you can, keep your branch around, so we can recover the script changes later

liblit:
Branch [`IntelliJ-IDEA-2019.3-inspection-updates`](https://github.com/liblit/WALA/tree/IntelliJ-IDEA-2019.3-inspection-updates) from my staging repository at 9d9856711c0d7b97f677937aaf437659f726acf4 is now clean, with neither the Bash porting work nor its reversion.

Branch [`git-hooks-pre-commit-as-Bash-script`](https://github.com/wala/WALA/tree/git-hooks-pre-commit-as-Bash-script) from this official WALA repository contains just the unreverted Bash porting work in case it turns out to be useful to someone later.

PR #606, opened 2019-12-09
Convert pre-commit script to bash

msridhar:
Restore @liblit's aborted improvements to the pre-commit script from #599.  Those changes were aborted as they would cause a hard-to-understand error in the existing `pre-commit-stub` script.  Here we detect the bad condition and emit a more understandable error asking the user to update their script.


msridhar:
Tested locally and the new script works for formatting as well as detecting the old `pre-commit-stub`

PR #607, opened 2019-12-12
Windows build and test improvements

liblit:
These changes fix various failures in building or testing WALA on Windows. More work remains to be done before WALA on Windows passes all of its own tests, but these are constructive steps in that direction.


build.gradle
```diff
@@ -167,26 +169,30 @@ allprojects {
...
-		if (!isAvailable) executable 'true'
-
-		args inputs.files
-
-		final def consoleOutput = System.out
-		final def fileOutput = new FileOutputStream(outputs.files.singleFile)
-		final def bothOutput = new org.apache.tools.ant.util.TeeOutputStream(consoleOutput, fileOutput)
-		standardOutput = errorOutput = bothOutput
+if (System.getProperty('os.name').startsWith('Windows')) {
+	tasks.register('shellCheck')
```
msridhar: Why does it need to be registered here?  Maybe add a comment?

liblit: Clarified in 6a72360f106e1ca7b408749af2753ff718f19ed1.

liblit:
I have added continuous integration testing on Windows to our existing GitHub Actions. This required [disabling four tests that currently fail on Windows](https://github.com/wala/WALA/pull/607/commits/b85fa0f99b3ac9ee7dc4bdbe9f9d7645954b8764). All four are clearly identified by comments that refer to [a tracking issue for Windows test failures](https://github.com/wala/WALA/issues/608) so we don’t forget that these still need work.

@msridhar, I will need some help with one regression introduced by [that test-disabling commit](https://github.com/wala/WALA/pull/607/commits/b85fa0f99b3ac9ee7dc4bdbe9f9d7645954b8764). Travis CI shows that [the Maven build now fails](https://travis-ci.org/wala/WALA/jobs/624431350#L908). I suspect that the problem is that I am using newer Hamcrest 2.2 matchers. We [upgraded Hamcrest from 1.3 to 2.2 in the Gradle configuration](https://github.com/wala/WALA/pull/570), but that did not affect Maven builds. Presumably this line is relevant:

https://github.com/wala/WALA/blob/6f7e4288493257e7b056081abc5e6bc2654a45c2/com.ibm.wala.core.tests/META-INF/MANIFEST.MF#L13

But I cannot figure out what to change this to (or what to change elsewhere) to make Maven builds use Hamcrest 2.2 like the Gradle builds already do. @msridhar, can you help?

msridhar:
Oof I am stumped on Hamcrest 2.2 with Maven / Tycho as well.  Somehow the usual tricks of adding it to the Eclipse plugin classpath don't seem to work, maybe because an older version is packaged with the Eclipse version we are building against.  @liblit any chance you could avoid using those particular matchers until we can remove Maven build support?  Not sure how easy this will be to fix for the moment.

msridhar:
Update: I seem to be making progress on the Hamcrest issue.  Will push if it works.

msridhar:
Hopefully 508a8bd will fix it.  Apparently [upgrading from Hamcrest 1.x is non-trivial](http://hamcrest.org/JavaHamcrest/distributables#upgrading-from-hamcrest-1x).  I think the Gradle update worked via some kind of dependency resolution magic, but getting the Eclipse stuff working required more hackery.

.github/workflows/continuous-integration.yml
```diff
@@ -10,13 +10,16 @@ jobs:
...
+          - windows-latest
         java-compiler:
           - default
           - ecj
         exclude:
           # testing ECJ compilation on any one OS is sufficient
           - os: macos-latest
             java-compiler: ecj
+          - os: windows-latest
+            java-compiler: ecj
```
msridhar: Why ECJ on Windows?

msridhar: Actually, why also ECJ on Mac?

liblit:
Great work with the Hamcrest version wrangling, @msridhar! Thanks for the assist.

.github/workflows/continuous-integration.yml
liblit: Take a closer look at the comment and `exclude:` lines just above. Testing ECJ compilation on any one OS is sufficient, so we let ECJ on Ubuntu through and exclude the others.

PR #612, opened 2019-12-14
Convert cast.java.test to use the main source set

msridhar:
There are actually no tests run in `com.ibm.wala.cast.java.test`.  It just has abstract test classes that get extended in other test modules.  So, use the main source set instead of the test source set, which will enable us to more easily put an artifact on Maven Central (to let us move IDE projects out of the repo, to enable retiring Maven).


msridhar:
I'll create a follow-up issue for test fixtures.  My only concern there is ensuring that our current way of publishing artifacts works for getting a test fixture artifact.  Since our artifact publication is slightly non-standard that might take some work.  But once we remove Maven support maybe things will get easier.

msridhar:
Issue #613 created

Issue #462, opened 2019-05-04
Proposal: move com.ibm.wala.ide.* projects to a separate Git repository


msridhar:
I propose we move the `com.ibm.wala.ide.*` projects out of the wala/WALA repo and into a sibling repo like [wala/ML](https://github.com/wala/ML).  Pros:

* Will speed up Gradle builds of wala/WALA, as we'll no longer need the `p2AsMaven` integration.  (Caching makes this mostly a one-time cost, but there are various scenarios where the download needs to be repeated.)
* We can deprecate building this repo with Maven, which will speed up CI significantly (the Maven Travis jobs take about twice as long as the Gradle ones).  We will also get away from having a duplicated build structure that needs to be maintained.

Cons:
* We can asynchronously detect if/when a core WALA change breaks the IDE functionality, but we won't be blocked from landing such changes.
* Various ECJ-based checks only run on the Maven builds now.  Without an alternative, we would no longer block the build on those detected issues (e.g., unused imports).

I would argue that few WALA users rely on the IDE functionality, particularly since there is an ECJ-based Java source frontend now.  So I think this is the right tradeoff.  For the second con, I would argue we should move to Error Prone, which integrates through standard Java compiler interfaces and also has many of the checks built into ECJ.  Plus it integrates easily with Gradle.

@liblit @juliandolby thoughts?

liblit:
I am generally in favor of anything that lets us retire the Maven build system. I would also be happy to see the slow `p2AsMaven` downloads go away. I think that those two pros outweigh the two cons, so overall I support this proposal.

liblit:
I see several Gradle plugins that switch Gradle to use the ECJ compiler for batch compilation. If we can make the choice of compilers selectable on the `./gradlew` command line, then we can have standard-compiler Travis CI jobs and _also_ ECJ Travis CI jobs.

If we can make that work cleanly, I think that would be a good idea even if we do get Error Prone wired up. Error Prone will never be identical to what Eclipse will find. If we want to keep the code ECJ-clean, then we should include ECJ builds in our testing regime.

juliandolby:
I am mostly in favor of this as well, but I wonder about a few things:

1) We definitely need to keep the ECJ-based Java source front end working with the latest WALA, as that is being actively used in current projects.  I think we need to block changes that break it.

2) I think currently all tests of the ECJ-based source front end are actually in the com.ibm.wala.ide.jdt.test project, so we'd likely need a new test project in the core for that code.

[...]

PR #618, opened 2019-12-16
Reduce the scope of Maven builds

msridhar:
Maven builds now only do close to the minimum necessary to run the `com.ibm.wala.ide.*` Eclipse plug-in tests.  Unfortunately we can't easily port these test runs to Gradle as they rely on [Tycho](https://www.eclipse.org/tycho/), which is Maven-only.

As part of this change we do the following:
1. Remove most sub-modules from the top-level `pom.xml` so Maven no longer builds them.
2. Change the `com.ibm.wala.ide.*` projects to rely on the other WALA projects as Maven artifacts.  This means that `./gradlew install` must be run first for a Maven build to succeed or to see updated versions of other WALA code.  (Maybe we should run this as part of the Maven build?)
3. Related to 2, update the relevant Eclipse plug-in metadata in `com.ibm.wala.ide.*`.  In particular, we need exactly one plug-in to manage each core WALA Maven artifact jar and re-export its relevant packages.  Probably more packages are exported than needed right now; we can trim it down later if needed (I doubt it matters).
4. Update relevant Travis CI scripts in light of 2.

With these changes, a Maven build on Travis CI with a warm cache runs in under 5 minutes.

Note that we still pull in test data for `com.ibm.wala.ide.jdt.test` in a manner outside the WALA Maven artifacts, due to the particular structure required.  (`wala.cast.java.test.data` must be zipped in the structure of an Eclipse project and that zip needs to be copied over.)  We can further clean this up in the future.

This change is an intermediate step toward #462 and retiring Maven, and reduces the scope of Maven usage, so it is worth landing on its own.  A follow-up PR will remove various Maven, Ant, and Eclipse build configuration files that are no longer needed.


com.ibm.wala.cast.java.test.data/build.xml
```diff
@@ -78,8 +68,8 @@
 	   excludes="**/Activator.java"/>
   </target>
   
-  <target name="jar" depends="compile,testdatadir">
-    <jar destfile="${basedir}/../com.ibm.wala.ide.jdt.test/testdata/test_project.zip" manifest="${basedir}/META-INF/MANIFEST.MF" basedir="${basedir}"/>
+  <target name="jar" depends="compile">
```
msridhar: I suspect this was the main reason we had to run `mvn verify` twice to get the tests to pass before.  This script was writing a zip file directly to `com.ibm.wala.ide.jdt.test` but that project declared no dependence on this one.  I've fix things so the zip file gets written locally and then copied over by `com.ibm.wala.ide.jdt.test`, and that a dependence exists in the build.

travis/before-install-maven
```diff
@@ -1,8 +1,6 @@
 #!/bin/sh -eux
 
-./gradlew prepareMavenBuild
-
-git clone --depth=1 https://github.com/secure-software-engineering/DroidBench.git /tmp/DroidBench
+./gradlew prepareMavenBuild install -x javadoc
```
liblit: Ah, nice idea to exclude `javadoc` here!

msridhar: Yes, I also exclude javadoc when running the main Maven build.  We generate javadoc for everything during the Gradle build, so no need to do it here.  And it's surprising how expensive javadoc generation is!  It takes longer than compilation.

msridhar:
My understanding of the Eclipse build system is also not deep (though it's deeper than it was a few days ago...).  I was highly reliant on keeping the regression tests passing for sanity-checking my changes.  Since these changes succeed in doing that, and we have few to no users of the IDE code (as far as we know), this is good enough I think.

PR #623, opened 2019-12-22
Remove unnecessary build files

msridhar:
Remove various `pom.xml`, `build.xml`, and Eclipse build files that are no longer needed


build.gradle
```diff
@@ -59,8 +59,6 @@ subprojects { subproject ->
 		mavenCentral()
 	}
 
-	jar.manifest.from('META-INF/MANIFEST.MF')
```
msridhar: @liblit the CI build passes without this, so I assumed it was ok to remove

PR #630, opened 2019-12-25
Refactoring and validation of Git ignore patterns

liblit:
When a pattern applies only to one subdirectory, replace it with an equivalent pattern sunk down into a `.gitignore` in that subdirectory. This refactoring will simplify some broad path refactorings I have planned in the near future.

Remove obsolete Git ignore patterns for three subdirectories that no longer exist. Presumably they did exist at some time, but have since been removed.

Validate Git ignore patterns after regular Travis CI tests. All files created during testing should already be ignored, and no Git-tracked file should be modified during testing.  In other words, `git status` should still report a clean tree after testing is done.


travis/script-gradle
```diff
@@ -34,3 +34,11 @@ else
     ${documentation:+$SUBMODULE_PREFIX$documentation}
   run_gradle "$SUBMODULE_PREFIX"build
 fi
+
+# Validate ".gitignore" patterns.  All files created during testing should already be ignored, and
+# no git-tracked file should be modified during testing.  In other words, "git status" should still
+# report a clean tree after testing is done.
+if [ -n "$(git status --porcelain)" ]; then
+  git status
```
msridhar: Maybe `echo` a more explicit message indicating why the job is failing here?

liblit: Good call. Added in f45926499f7129eb48ecb97ce80072773dfa81b2.

liblit:
Requesting re-review after adding Git cleanliness check to GitHub CI builds as well. Previously I only thought to add this for Travis CI builds.

.github/workflows/continuous-integration.yml
```diff
@@ -55,5 +55,14 @@ jobs:
       - name: Build and test using Gradle and default JDK compiler
         run: ./gradlew --continue --no-build-cache --no-daemon linters javadoc build
         if: matrix.java-compiler != 'ecj'
+      - name: Check for Git cleanliness after build and test
+        run: |
+          if [ -n "$(git status --porcelain)" ]; then
```
msridhar: No way to avoid the copy-paste?  Not at all a huge deal, just curious

liblit: Best avoided by extracting these few lines into their own shells script. I was being lazy, but yeah, I guess that's worth doing unless we expect to retire Travis CI entirely soon.

liblit: Previously cloned code extracted into shell script in 1bdda400e939292b29b9a017f46a41e1bc2d9ccb.

PR #635, opened 2019-12-27
Standardize Gradle source and resource paths

liblit:
Gradle is an “opinionated” build system:  it has default assumptions about many things, including paths to source and resource files. These defaults can be overridden, but it can be more pleasant to conform to Gradle’s expectations. These expectations may be shared by new WALA developers, so conformance eases onboarding. Conformance also allows removing many little fragments of Gradle build configuration logic that were previously needed to override the defaults.


liblit:
I’m glad you like and approve this change, @msridhar. Before I merge it, we should think a bit about a second suite of changes that I have been considering but not yet started: merging most `foo.bar.test` subprojects into their corresponding `foo.bar` subprojects.

This second suite of changes would also involve mass directory restructuring: not as many as this one, but still many. I wonder if it would be better to delay merging this request until I am ready to apply that second suite of changes at the same time. That should reduce disruption for anyone else hacking on WALA: they get hit with one tricky merge rather than two.

The disadvantage of waiting is that there is no guarantee that the second set of changes will actually work and be desirable. I don’t like having a big change like this one sitting around, approved but unmerged, any longer than necessary.

@msridhar, any thoughts?

msridhar:
I'm very much in favor of merging the `foo.bar.test` subprojects into `foo.bar` wherever possible, assuming things still work afterward.  And if we're doing that change soon, we should do both changes together.

@liblit any way you can do a quick test as to the feasibility / difficulty of merging the test projects, and we can decide based on that?

liblit:
@msridhar wrote
> @liblit any way you can do a quick test as to the feasibility / difficulty of merging the test projects, and we can decide based on that?

I can and I have. The conversation continues in #636.

PR #650, opened 2020-01-08
#649 [Shrike] Float and Double instrumentation

### LINKED ISSUE #649: [Shrike] Float- and DoubleConstants instrumented as int's
Hello, I am using WALA for quite a while and was wondering why adding new double/float constants do not have the expected value in the compiled bytecode.

I created a simple method which pushes a double value as a first instruction and print its value. The first instruction is pushing the constant on the stack. As expected, the value is printed on the console. For example:
```java
public class TestClass {
		double value = Math.PI;
		System.out.println("Value starts with: " + value);
	}
}
```
Bytecode of the main-method:
```
0: Constant(D,3.141592653589793)
1: LocalStore(D,1)
2: Get(Ljava/io/PrintStream;,STATIC,Ljava/lang/System;,out)
3: LocalLoad(D,1)
4: InvokeDynamic [java/lang/invoke/StringConcatFactory:makeConcatWithConstants(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;Ljava/lang/String;[Ljava/lang/Object;)Ljava/lang/invoke/CallSite;] makeConcatWithConstants(D)Ljava/lang/String;
5: Invoke(VIRTUAL,Ljava/io/PrintStream;,println,(Ljava/lang/String;)V)
6: Return(V)
```

After using an OfflineInstrumenter and replacing (only) the first instruction (index 0 above), the compiled method does only push the int value of the new double constant, which results in the following output (example for double value 9.55555):
```java
	...
	me.replaceWith(0, new Patch() {
		@Override
		public void emitTo(Output w) {
			w.emit(ConstantInstruction.make(9.55555d));
		}
	});
	me.applyPatches();
	...
```
`Value starts with: 9.0`

The conversion issue is located in com.ibm.wala.shrike/src/com/ibm/wala/shrikeCT/ClassWriter.java in method:
```java
private void emitConstantPool() {
	...
	} else if (o instanceof Double) {
		int offset = reserveBuf(9);
		buf[offset] = CONSTANT_Double;
		setDouble(buf, offset + 1, ((Double) o).intValue());
	}
	...
}
```
Why is `intValue()` called on a new double constant? The same applies to new float constants a few lines above. If needed, I can offer a minimal example implementation.

RalfHerzog:
This PR fixes #649, where floating number constants are no longer passed as their integer value.
Verified by an example application.


RalfHerzog:
@msridhar Should I create another java gradle project at root level with the name `com.ibm.wala.shrike.test` for testing or is their already a project for shrike tests?

msridhar:
Thanks for the PR!

> @msridhar Should I create another java gradle project at root level with the name `com.ibm.wala.shrike.test` for testing or is there already a project for shrike tests?

I think any tests related to shrike would currently be in `com.ibm.wala.core.tests`.  Can you just stick the tests there for now, but leave a comment that they should be moved to a Shrike test project in the future?  @liblit and I are working on switching to a more standard directory layout for WALA projects, after which we will be able to move the tests to a `src/test/java` folder inside `com.ibm.wala.shrike`.

For `com.ibm.wala.core.tests`, you should stick the test input inside `com.ibm.wala.core.testdata`.  Hopefully the other tests serve as decent examples.  Let me know if you have any troubles.

msridhar:
@RalfHerzog let us know if you need more help here

RalfHerzog:
Here is an example class for testing double and float instrumentation, as well as a complete shrike test for it.
For better understanding, I want to describe how the test is structured:

1. Create a temp file for the instrumented jar
2. Create an OfflineInstrumenter which reads the example class (shrike/FloatingPoints) with a double and float method
3. Get the first instruction which is either a double- or float-constant-instruction. The value of this will be altered by a patch
4. Modify the previously found value of the constant instruction (by adding 2.5 currently)
5. Write the modified method as a jar to the temp file created in step 1.
6. Create an OfflineInstrumenter again but read the modified jar from step 5.
7. Find the correct method (either double or float example method) and constant instruction from step 3. and validate, that the value was written incremented by 2.5 (step 4.)

msridhar:
Thanks for the contribution @RalfHerzog!

PR #654, opened 2020-01-12
Add Gradle versions plugin

msridhar:
See [the docs](https://github.com/ben-manes/gradle-versions-plugin/blob/master/README.md).  Now we can run `./gradlew dependencyUpdates` to see which dependencies are out of date.  This is a complement to dependabot, which doesn't work perfectly.  (E.g., there are still a bunch of references to JUnit 4.12, even though we already updated to 4.13 based on the dependabot PR.)  The plugin is customizable so eventually we could ignore certain deps, certain types of releases, etc.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,6 +15,7 @@ plugins {
 	id 'de.undercouch.download'
 	id 'nebula.lint' version '14.2.5'
 	id 'nebula.source-jar' version '14.1.1' apply false
+	id 'com.github.ben-manes.versions' version '0.27.0'
 }
 
 repositories {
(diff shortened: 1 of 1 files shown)


msridhar:
Correction: my comment about the incomplete JUnit update was incorrect; I was looking at an out-of-date master.  Still, adding this plugin is useful I think.

build.gradle
```diff
@@ -15,6 +15,7 @@ plugins {
 	id 'de.undercouch.download'
 	id 'nebula.lint' version '14.2.5'
 	id 'nebula.source-jar' version '14.1.1' apply false
+	id 'com.github.ben-manes.versions' version '0.27.0'
```
liblit: Please keep this list of plugins in sorted order. I prefer sorting for things that are not order-dependent; I do the same for Java `import`s and C/C++ `#include`s.

PR #657, opened 2020-01-18
Delay creating an empty List until actually needed

liblit:


liblit:
Fair point; `Collections.emptyList()` is not a factory. But avoiding the unnecessary call is still sensible.

PR #658, opened 2020-01-18
Improvements for Java after 1.8

liblit:
These are assorted fixes to building, running, and testing WALA under Java 9 and later. It's not all working; there's still more work to be done. But these are steps in the right direction. Any future effort to bring WALA up-to-date with modern Java would likely have to rediscover each of these glitches and more.


com.ibm.wala.core.tests/build.gradle
```diff
@@ -44,17 +44,13 @@ tasks.named('test') {
...
-		exclude '**/callGraph/Java7CallGraphTest.class'
-		exclude '**/callGraph/CloneTest.class'
-		exclude '**/callGraph/DebuggingBitsetCallGraphTest.class'
-		exclude '**/callGraph/KawaCallGraphTest.class'
-		exclude '**/exceptionpruning/ExceptionAnalysisTest.class'
-		exclude '**/exceptionpruning/ExceptionAnalysis2EdgeFilterTest.class'
-		exclude '**/cha/LibraryVersionTest.class'
-		exclude '**/ir/TypeAnnotationTest.class'
+	// temporarily turn off some tests on JDK 11+
+	switch (JavaVersion.current()) {
```
msridhar: Why use a `switch` here as opposed to just an `if`?

liblit: Over-engineered remnant from when I thought I would need more cases. Replaced with equivalent `if` in 43052fb065087b2510b584552afcaff124e47756. Thank you for your attentive reviewing!

PR #661, opened 2020-01-19
Don't save a CallGraph we never use in the future

liblit:
We still build the call graph, which "shouldn't crash" per the existing comment.  But there's no need to save it in an otherwise-unused local variable.


liblit:
If the Java compiler were going to optimize away an entire `CallGraphTestUtil.buildZeroCFA` call, then I doubt that saving the result of that call in a local variable (or not) would make any difference.

That being said, I'm happy to revise this to print some result here if you can suggest what exactly should be printed.

msridhar:
Rather than printing, we can add a simple assertion that `getNumberOfNodes() > 0` on the result of the call. It’s a dummy assertion but should prevent optimization

liblit:
> Rather than printing, we can add a simple assertion that `getNumberOfNodes() > 0` on the result of the call. It’s a dummy assertion but should prevent optimization

Added in 3d67f726b8d43df88c8a291a72e059fbe7c7cc9d.

PR #669, opened 2020-01-30
Add JUnit4 entrypoint support

khatchad:
Enhancement to add JUnit4 entrypoint support. AFAIK, WALA previously only supported JUnit3.

I did not include regression tests in this PR as I am not very familiar with entrypoint testing in WALA. I am happy to add some if you think that they can be useful.

### DIFF
--- a/com.ibm.wala.core/src/main/java/com/ibm/wala/util/scope/JUnitEntryPoints.java
+++ b/com.ibm.wala.core/src/main/java/com/ibm/wala/util/scope/JUnitEntryPoints.java
@@ -84,7 +84,7 @@ public class JUnitEntryPoints {
           Set<IMethod> setUpTearDowns;
           try {
             setUpTearDowns = getSetUpTearDownMethods(klass);
-          } catch (ClassHierarchyException e) {
+          } catch (Exception e) {
             throw new IllegalArgumentException("Can't find test method entry points using class hierarchy: " + cha, e);
           }
           for (IMethod m : setUpTearDowns) {
--- a/com.ibm.wala.core/src/main/java/com/ibm/wala/util/scope/JUnitEntryPoints.java
+++ b/com.ibm.wala.core/src/main/java/com/ibm/wala/util/scope/JUnitEntryPoints.java
@@ -10,12 +10,6 @@
  */
 package com.ibm.wala.util.scope;
 
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.Set;
-import java.util.logging.Logger;
-
 import com.ibm.wala.classLoader.IClass;
 import com.ibm.wala.classLoader.IClassLoader;
 import com.ibm.wala.classLoader.IMethod;
@@ -29,6 +23,11 @@ import com.ibm.wala.types.TypeName;
 import com.ibm.wala.types.annotations.Annotation;
 import com.ibm.wala.util.collections.HashSetFactory;
 import com.ibm.wala.util.strings.Atom;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.logging.Logger;
 
 /**
  * This class represents entry points ({@link Entrypoint})s of JUnit test methods. JUnit test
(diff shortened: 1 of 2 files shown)


msridhar:
@khatchad you can run `./gradlew -PjavaCompiler=ecj compileJava` to reproduce the CI failure locally.  Looks like there are unused imports.

khatchad:
Thanks, @msridhar. I am still getting used to this new compilation process. I'll have a look.

khatchad:
`./gradlew build` on my local machine succeeded.

msridhar:
> `./gradlew build` on my local machine succeeded.

Yeah, unfortunately not all the ECJ checks run as part of `./gradlew build` by default.  You need to run `./gradlew -PjavaCompiler=ecj compileJava`.  Alternately, if you run `./gradlew goJF` that will run Google Java Format, which formats the code and deletes any unused imports.

com.ibm.wala.core/src/main/java/com/ibm/wala/util/scope/JUnitEntryPoints.java
```diff
@@ -44,33 +67,85 @@
...
     }
     return result::iterator;
   }
 
+  private static boolean isTestEntryPoint(TypeName typeName) {
+    String javaName = walaTypeNameToJavaName(typeName);
+    return TEST_ENTRY_POINT_ANNOTATION_NAMES.contains(javaName);
+  }
+
+  private static String walaTypeNameToJavaName(TypeName typeName) {
```
msridhar: Can we replace this method with just using `com.ibm.wala.util.strings.StringStuff.jvmToReadableType()` instead?

com.ibm.wala.core/src/main/java/com/ibm/wala/util/scope/JUnitEntryPoints.java
```diff
@@ -87,24 +162,20 @@
         TypeName.findOrCreateClass(targetPackageAtom, targetSimpleClassAtom);
     final Atom targetMethodAtom = Atom.findOrCreateAsciiAtom(targetMethodName);
 
-    if (DEBUG) {
-      System.err.println(("finding entrypoint " + targetMethodAtom + " in " + targetType));
-    }
+    logger.finer("finding entrypoint " + targetMethodAtom + " in " + targetType);
```
msridhar: In other places you pass a lambda to the logger API.  Why not here?

com.ibm.wala.core/src/main/java/com/ibm/wala/util/scope/JUnitEntryPoints.java
```diff
@@ -87,24 +162,20 @@
...
 
     final Set<Entrypoint> entryPts = HashSetFactory.make();
 
     for (IClass klass : cha) {
       TypeName klassType = klass.getName();
       if (klassType.equals(targetType) && isJUnitTestCase(klass)) {
-        if (DEBUG) {
-          System.err.println("found test class");
-        }
+        logger.finer("found test class");
```
msridhar: Same question regarding lack of lambda

com.ibm.wala.core/src/main/java/com/ibm/wala/util/scope/JUnitEntryPoints.java
khatchad: Sounds plausible to me.

com.ibm.wala.core/src/main/java/com/ibm/wala/util/scope/JUnitEntryPoints.java
khatchad: Ths one has a seemingly simple concatenation. Creating a lambda has a slight overhead (in fact, we have been studying this recently). Ideally, you would want to use the lambda version where the concentration outweighs the overhead.

In contrast, the statement at https://github.com/wala/WALA/pull/669/files#diff-6fe8230dbe8437b11645de9ccd666b8eR86 calls a `toString()` method, and I don't know how involved that method is. To be on the safe side, I used a lambda to avoid possibly unnecessary and possibly costly string concatenation.

com.ibm.wala.core/src/main/java/com/ibm/wala/util/scope/JUnitEntryPoints.java
khatchad: This one is even more obvious not to use a lambda as there's no string concatenation at all.

PR #672, opened 2020-01-31
More verbose output for wala.core tests

msridhar:
This should fix the timeouts we have been seeing on Travis. Not sure why they started, but I don’t see a fundamental underlying performance problem that arose recently


liblit:
I assume you will squash those reverted changes when merging? If so, please also add a descriptive commit message. Otherwise future maintainers May have no idea why this one subproject is logging more verbosely than the others.

PR #676, opened 2020-02-01
Prefer keeping downloaded/generated files in build outputs subdirectories

liblit:
Historically WALA has stored downloaded/generated files in a variety of places, and used many `.gitignore` files to exclude them from source control. I prefer that we use Gradle's standard `*/build` subdirectories where possible, just to keep things tidy.

These commits certainly don't move *every* downloaded or generated file into `*/build`. But they move a few, and clean up `.gitignore` files accordingly. They also improve a few test failure messages in cases where an expected test resource could not be found.


msridhar:
I am slightly concerned about losing downloaded files after running `./gradlew clean`.  Is the idea that the files will still be present in the build cache, so they won't have to be re-downloaded?

liblit:
Gradle dependency management works well enough that `./gradlew clean` should rarely be needed. But if it is, then yes, downloads can be repopulated from the build cache. I just tried that myself to be sure:

1. `./gradlew build -PexcludeSlowTests`
2. `./gradlew clean`
3. `./gradlew build -PexcludeSlowTests`

In step 3, every `download*` task reported `FROM-CACHE` except for `:com.ibm.wala.dalvik:downloadSampleCup`. I would need to look into that last task more closely, but it should unrelated to the changes proposed here.

liblit:
For the record, `:com.ibm.wala.dalvik:downloadSampleCup` is not actually a download task at all, and the fact that it is not cached is benign. I will change its name in an upcoming pull request.

PR #688, opened 2020-02-09
Reimplement publishing using `maven-publish` plugin

liblit:
# Description of Changes

## Jar Archives for Test Fixtures

Previously we published three jars for each subproject: sources, Javadoc, and class files. We still publish those, and now also publish sources, Javadoc, and class files for each subproject’s test fixtures. We also publish various checksum files. All jar archives and the POM files should be signed (via command-line GnuPG) when publishing a release.

## Task Name Changes

Names of tasks used for publishing have changed. Run `./gradlew tasks --group=publishing` for a list of all publishing-related tasks. Tasks likely to be of most interest are:

* `publishAllPublicationsToProjectBuildRepository`: populates a Maven repository in `build/maven-repository`. This is a good way to see what would be published without actually affecting the world outside the build tree.

* `publishToMavenLocal`: populates a Maven repository in `~/.m2`. This is the typical way to install WALA locally for use by other dependent projects.

* `publishAllPublicationsToMavenRepository`: uploads a WALA release or development snapshot to an external server. Releases are signed, but snapshots are not.

## Continuous Integration

We now publish to a local Maven repository as part of CI testing. This seems useful to help us notice if a change breaks publishing. Even better would be to *validate* the published repository contents, which we don’t currently do.

## Additional Metadata

Published POM files now contain more metadata than before. POM metadata previously included SCM links, just one developer/maintainer, and some license info which was actually inconsistent with the `LICENSE` file. Now we provide SCM links, *all* official WALA maintainers, consistent license information, project inception year, issue-management info, mailing-list info, and continuous-integration info.

# Reviewing Advice

## General Guidance

I have tested this to the best of my abilities, but it would still benefit from careful review by others. Consider checking out this pull request’s Git branch and examining the published artifacts for yourself. Use `./gradlew publishToMavenLocal` to populate a Maven repository in `~/.m2`. Or use `./gradlew publishAllPublicationsToProjectBuildRepository` to populate a Maven repository in `build/maven-repository`: a good way to see what would be published without actually affecting the world outside the build tree.

## Specific Areas of Concern

The parts I am least confident about involve uploading snapshots and releases to remote Maven repositories, and also signing of (non-snapshot) releases. I am not sure whether I even would know how to do these things myself using the old implementation.

Please take a particularly careful look at this logic for selecting the remote Maven repository URL:

https://github.com/wala/WALA/blob/054ab8ced73880664fc1c6abaa4ad50c71996c1d/gradle-mvn-push.gradle#L116-L118

Please also check this logic for fetching the username and password to upload to those remote repositories:

https://github.com/wala/WALA/blob/054ab8ced73880664fc1c6abaa4ad50c71996c1d/gradle-mvn-push.gradle#L120-L123

I based these on what was here before, but mistakes can happen, and none of this uploading logic has been tested in this new implementation.

### DIFF
--- a/gradle-mvn-push.gradle
+++ b/gradle-mvn-push.gradle
@@ -138,7 +138,7 @@ signing {
 	required = true
 }
 
-// Only sign the archives if we are uploading a snapshot or release.
+// Only sign release uploads; snapshots and local installations are unsigned.
 tasks.withType(Sign) {
 	onlyIf {
 		!isSnapshot &&
--- a/gradle-mvn-push.gradle
+++ b/gradle-mvn-push.gradle
@@ -122,11 +122,6 @@ publishing {
 				password = properties.get('SONATYPE_NEXUS_PASSWORD')
 			}
 		}
-
-		maven {
-			name = 'projectBuild'
-			url = "file://$rootProject.buildDir/maven-repository"
-		}
 	}
 }
 
--- a/gradle-mvn-push.gradle
+++ b/gradle-mvn-push.gradle
@@ -125,12 +125,12 @@ publishing {
 	repositories {
 		maven {
 			url = (isSnapshot
-					? properties.getOrDefault('SNAPSHOT_REPOSITORY_URL', 'https://oss.sonatype.org/content/repositories/snapshots/')
-					: properties.getOrDefault('RELEASE_REPOSITORY_URL', 'https://oss.sonatype.org/service/local/staging/deploy/maven2/')
+					? project.properties.getOrDefault('SNAPSHOT_REPOSITORY_URL', 'https://oss.sonatype.org/content/repositories/snapshots/')
+					: project.properties.getOrDefault('RELEASE_REPOSITORY_URL', 'https://oss.sonatype.org/service/local/staging/deploy/maven2/')
 			)
 			credentials {
-				username = properties.get('SONATYPE_NEXUS_USERNAME')
-				password = properties.get('SONATYPE_NEXUS_PASSWORD')
+				username = project.properties.get('SONATYPE_NEXUS_USERNAME')
+				password = project.properties.get('SONATYPE_NEXUS_PASSWORD')
 			}
 		}
 
--- a/gradle-mvn-push.gradle
+++ b/gradle-mvn-push.gradle
@@ -1,7 +1,7 @@
 apply plugin: 'maven-publish'
 apply plugin: 'signing'
 
-final def isSnapshot = version.contains('SNAPSHOT')
+final isSnapshot = version.contains('SNAPSHOT')
 
 publishing {
 	publications {
--- a/gradle-mvn-push.gradle
+++ b/gradle-mvn-push.gradle
@@ -133,6 +133,11 @@ publishing {
 				password = properties.get('SONATYPE_NEXUS_PASSWORD')
 			}
 		}
+
+		maven {
+			name = 'fakeRemote'
+			url = "file://$rootProject.buildDir/maven-fake-remote-repository"
+		}
 	}
 }
 
--- a/gradle-mvn-push.gradle
+++ b/gradle-mvn-push.gradle
@@ -3,10 +3,6 @@ apply plugin: 'signing'
 
 final isSnapshot = version.contains('SNAPSHOT')
 
-final isRemoteRepository(taskGraph) {
-	return taskGraph.hasTask("$project.path:publishRemotePublicationToMavenRepository")
-}
-
 final skipTestFixtures() {
 	components.java.withVariantsFromConfiguration(configurations.testFixturesApiElements) {
 		skip()
--- a/gradle-mvn-push.gradle
+++ b/gradle-mvn-push.gradle
@@ -3,24 +3,101 @@ apply plugin: 'signing'
 
 final isSnapshot = version.contains('SNAPSHOT')
 
+final isRemoteRepository(taskGraph) {
+	return taskGraph.hasTask("$project.path:publishRemotePublicationToMavenRepository")
+}
+
+final skipTestFixtures() {
+	components.java.withVariantsFromConfiguration(configurations.testFixturesApiElements) {
+		skip()
+	}
+	components.java.withVariantsFromConfiguration(configurations.testFixturesRuntimeElements) {
+		skip()
+	}
+}
+
+final sharedPublicationConfiguration(publication) {
+	publication.from components.java
+	publication.pom {
+		name = POM_NAME
+		description = 'T. J. Watson Libraries for Analysis'
+		inceptionYear = '2006'
+		url = 'https://github.com/wala/WALA'
+
+		ciManagement {
+			system = 'Travis CI'
+			url = 'https://travis-ci.org/wala/WALA'
+		}
+
+		developers {
+			[
+					// Current WALA maintainers, alphabetical by ID
+					juliandolby: 'Julian Dolby',
+					liblit     : 'Ben Liblit',
+					msridhar   : 'Manu Sridharan',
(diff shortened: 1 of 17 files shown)


msridhar:
Sounds awesome.  It may take me a couple of days (or more) to get to reviewing this.  As you noted it should be done a bit carefully

liblit:
Thanks, @msridhar. Take as much time as you need to do it right.

gradle-mvn-push.gradle
```diff
@@ -1,228 +1,155 @@
...
+	// to avoid being prompted for a password once per artifact
+	useGpgCmd()
+	sign publishing.publications.wala
+	required = true
 }
 
-def getSnapshotRepositoryUrl() {
-  return hasProperty('SNAPSHOT_REPOSITORY_URL') ? SNAPSHOT_REPOSITORY_URL
-          : "https://oss.sonatype.org/content/repositories/snapshots/"
+// Only sign the archives if we are uploading a snapshot or release.
```
msridhar: We don’t want to sign snapshots. I think just the comment is incorrect here

travis/before-install-maven
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh -eux
 
-./gradlew prepareMavenBuild install -x javadoc
+./gradlew prepareMavenBuild publishToMavenLocal -x javadoc
```
msridhar: Still disable javadoc?

msridhar:
Also test fixture artifacts should be published only to the local Maven repository, not to external repositories. Is that still what happens here?

gradle-mvn-push.gradle
liblit: Good catch. Comment corrected in beed95b863238b967100014d46e517f8eb2cf1fa, then reworked again in de3911d0060b283195a7fab91b7f59d1d43ecdde along with other changes.

travis/before-install-maven
liblit: Yes, that still makes sense here. This Gradle command is intended to populate `~/.m2` for use by a subsequent Maven build. The Maven build doesn’t care about documentation, so we may as well skip it.

Excluding the `javadoc` tasks here does the same thing it did before: `javadoc` jars will be installed under `~/.m2`, but these jars will be empty except for the manifest files. The important thing is that no time will be spent creating Javadoc HTML pages.

liblit:
@msridhar wrote:
> Also test fixture artifacts should be published only to the local Maven repository, not to external repositories. Is that still what happens here?

Implemented in de3911d0060b283195a7fab91b7f59d1d43ecdde. We now define two “publications”: collections of items to publish. The `remote` publication excludes test-fixtures artifacts, and will only be used with remote repositories. The `local` publication includes test-fixtures artifacts, and will only be used with local installations.

msridhar:
@liblit is there a way to do a local test to see which artifacts would be pushed to a remote repository?  I tried `./gradlew publishRemotePublicationToMavenLocal` but that didn't seem to put anything under `~/.m2`.

liblit:
I’m not in a position to do this right now, but I should be able to define a new local repository under `build` that accepts the `remote` publication. Same jars, same signing (or not, if snapshot) as what the remote repositories would get. Seems like a useful way to preflight things. I’ll put that together the next time I’m in front of a suitable machine.

msridhar:
I tried `./gradlew publishRemotePublicationToMavenRepository` and it failed as follows:
```
Execution failed for task ':com.ibm.wala.shrike:publishRemotePublicationToMavenRepository'.
> Failed to publish publication 'remote' to repository 'maven'
   > Username must not be null!
```
I confirmed my `~/.gradle/gradle.properties` file still sets `SONATYPE_NEXUS_USERNAME` and `SONATYPE_NEXUS_PASSWORD`.  The code now uses `properties.get('SONATYPE_NEXUS_USERNAME')`; is this still doing the right thing?  @liblit you can test by sticking dummy values in your `~/.gradle/gradle.properties`.  When it's working you should see some kind of login failure.

travis-after-success.sh
```diff
@@ -27,7 +27,7 @@ elif [ "$TRAVIS_EVENT_TYPE" == "cron" ]; then
   echo "Skipping snapshot deployment; running cron job."
 else
   echo "Deploying snapshot..."
-  ./gradlew clean uploadArchives
+  ./gradlew clean publishAllPublicationsToMavenRepository
```
msridhar: Should this be `publishRemotePublicationToMavenRepository`?

msridhar:
Also once this lands, we should update these wiki docs as needed:

https://github.com/wala/WALA/wiki/WALA-Releases-and-Snapshots

travis/script-gradle
```diff
@@ -36,7 +36,7 @@ else
       compileTestJava \
       ${documentation:+$SUBMODULE_PREFIX$documentation}
   fi
-  run_gradle "$SUBMODULE_PREFIX"build
+  run_gradle "$SUBMODULE_PREFIX"build publishToMavenLocal
```
msridhar: If we run `publishToMavenLocal` here will that result in all modules getting compiled, even on the cron jobs?  Just want to make sure this won't defeat the purpose of the cron

travis-after-success.sh
liblit: Those are equivalent: we explicitly skip the `local` publication for remote repository uploads, so `All` here is effectively just `Remote`.

(Conversely, we also skip the `remote` publication for local repository installs.)

travis/script-gradle
liblit: @msridhar asked:
> is there a way to do a local test to see which artifacts would be pushed to a remote repository?

Added in af2329aad5dda5c32e7ac1cddd74e6fed518e3ff. The `publishAllPublicationsToFakeRemoteRepository` task will populate a local repository under `build/maven-fake-remote-repository` that contains exactly what we would upload to a remote repository.

liblit:
@msridhar reported:
```
> Failed to publish publication 'remote' to repository 'maven'
   > Username must not be null!
```

Corrected in 9bd513996c3ad38794c522ebd3dc3b9d1c9bda7f.

liblit:
@msridhar asked:
> is there a way to do a local test to see which artifacts would be pushed to a remote repository?

Added in af2329aad5dda5c32e7ac1cddd74e6fed518e3ff. The `publishAllPublicationsToFakeRemoteRepository` task will populate a local repository under `build/maven-fake-remote-repository` that contains exactly what we would upload to a remote repository.

liblit:
Oops, @msridhar, I accidentally deleted a question from you regarding single-submodule builds, which adding `publishToMavenLocal` would probably foul. Good point; that was corrected in 9235671e5fefd014ca999843419115a976391d0b.

liblit:
@msridhar wrote:
> Also once this lands, we should update these wiki docs as needed:
> 
> https://github.com/wala/WALA/wiki/WALA-Releases-and-Snapshots

Agreed! GitHub lets me clone that wiki as a Git project in its own right, but unfortunately there seems to be no way to queue up pull requests for such a project. I have reviewed the document and I believe that the following are the only required changes:

1. Under "Installing local snapshots", change the `install` Gradle task to `publishToMavenLocal`.

2. Under "Tagging and uploading a release", change the `uploadArchives` Gradle task to ``publishAllPublicationsToMavenRepository`.

msridhar:
In doing further testing here I ran into #691.  Once we fix that one, @liblit could you merge in the fix?  I want to test that the command to upload actual releases still works.

liblit:
Merged in fix for #691.

liblit:
Thanks for the thorough vetting, @msridar.

liblit:
Wiki docs updated to match.

PR #697, opened 2020-02-16
Fixes to allow thin slicer to be run with Java source frontend

### LINKED ISSUE #696: Thin slicer throws java.lang.ClassCastException: com.ibm.wala.ipa.modref.ModRef$ModVisitor cannot be cast to com.ibm.wala.cast.java.ssa.AstJavaInstructionVisitor
Thin Slicer throws ClassCastException when I try to run it for source code. I am sharing the stack trace below, driver and input code is posted [here](https://gist.github.com/ufarooq/6c9937fbdc20706a749c2b556cf52482).  
Here are a few notes:
* It works just fine when I am analyzing byte code. 
* Also, the context-sensitive traditional slicer is working fine for both Java and byte code. 

Stack Trace: 
``` 
Exception in thread "main" java.lang.ClassCastException: com.ibm.wala.ipa.modref.ModRef$ModVisitor cannot be cast to com.ibm.wala.cast.java.ssa.AstJavaInstructionVisitor
	at com.ibm.wala.cast.java.ssa.AstJavaInvokeInstruction.visit(AstJavaInvokeInstruction.java:49)
	at com.ibm.wala.ipa.modref.ModRef.scanNodeForMod(ModRef.java:140)
	at com.ibm.wala.ipa.modref.ModRef.lambda$scanForMod$0(ModRef.java:106)
	at com.ibm.wala.ipa.callgraph.CallGraphTransitiveClosure.collectNodeResults(CallGraphTransitiveClosure.java:61)
	at com.ibm.wala.ipa.modref.ModRef.scanForMod(ModRef.java:105)
	at com.ibm.wala.ipa.modref.ModRef.computeMod(ModRef.java:63)
	at com.ibm.wala.ipa.slicer.SDG.<init>(SDG.java:137)
	at com.ibm.wala.ipa.slicer.SDG.<init>(SDG.java:119)
	at com.ibm.wala.ipa.slicer.Slicer.computeBackwardSlice(Slicer.java:176)
	at android.wala.test.AnalyzeJava.main(AnalyzeJava.java:94)
```

msridhar:
The two main changes are:

1. `AstJavaSlicer` no longer is a subclass of `Slicer`.  This causes confusion, as many static methods in `Slicer` will not work with the AST frontend (since the wrong type of `ModRef` object is constructed; the AST slicer needs `AstModRef` objects).  This is technically a breaking API change (since static methods can be invoked on subclasses), but one should be very easy to adapt to and should make code clearer.
2. Some fixes inside `CISlicer` to ensure the right type of `ModRef` object is used.

Fixes #696


com.ibm.wala.cast.java/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -96,7 +96,7 @@ org.eclipse.jdt.core.compiler.problem.syntheticAccessEmulation=ignore
 org.eclipse.jdt.core.compiler.problem.terminalDeprecation=error
 org.eclipse.jdt.core.compiler.problem.typeParameterHiding=error
 org.eclipse.jdt.core.compiler.problem.unavoidableGenericTypeProblems=disabled
-org.eclipse.jdt.core.compiler.problem.uncheckedTypeOperation=error
+org.eclipse.jdt.core.compiler.problem.uncheckedTypeOperation=ignore
```
msridhar: @liblit I needed this since I couldn't get a suppression to work; see below

com.ibm.wala.cast.java/src/testFixtures/java/com/ibm/wala/cast/java/test/JavaIRTests.java
```diff
@@ -721,6 +727,38 @@ public void testMiniaturSliceBug() throws IllegalArgumentException, CancelExcept
...
   }
 
+  @Test
+  public void testThinSlice() throws CancelException, IOException {
+    List<String> sources =
+        Collections.singletonList(getTestSrcPath() + File.separator + "MiniaturSliceBug.java");
+    Pair<CallGraph, PointerAnalysis<? extends InstanceKey>> x =
+        runTest(sources, rtJar, simpleTestEntryPoint(), emptyList, true, null);
+
+    @SuppressWarnings("unchecked")
```
msridhar: @liblit somehow ECJ doesn't pay attention to this suppression and reports an unchecked cast error on the next line anyway, which is why I changed the ECJ config.  I also don't know how to write this code to avoid an unchecked cast, without potentially changing a lot of other code (and it didn't seem worth the effort).  Help welcome on making the suppression work :-)  Maybe it's because it's in text fixture code?

com.ibm.wala.core/src/main/java/com/ibm/wala/ipa/slicer/SlicerUtil.java
```diff
@@ -38,10 +37,11 @@ private SlicerUtil() {}
   public static Statement findCallTo(CGNode n, String methodName) {
     IR ir = n.getIR();
     for (SSAInstruction s : Iterator2Iterable.make(ir.iterateAllInstructions())) {
-      if (s instanceof SSAInvokeInstruction) {
-        SSAInvokeInstruction call = (SSAInvokeInstruction) s;
+      if (s instanceof SSAAbstractInvokeInstruction) {
```
msridhar: This is a related change to avoid an unnecessary dependence on `SSAInvokeInstruction`, which does not work for the AST frontend

msridhar:
Ok I will merge this as is. I will leave the suppression so maybe we will figure out later why it doesn’t work

PR #701, opened 2020-02-23
Build more files in "build" subdirectories

liblit:
Avoid cluttering subprojects' directories with generated files. Instead, prefer to generate files in various `build` subdirectories. This makes it easier to recognize what's part of the source tree and what's built.

Clean up Git ignore patterns accordingly.

This doesn't address the non-`build` generated-files problem entirely; several such files remain. But it's an incremental improvement.


com.ibm.wala.cast.js/src/testFixtures/java/com/ibm/wala/cast/js/test/ExtractingToPredictableFileNames.java
```diff
@@ -0,0 +1,26 @@
+package com.ibm.wala.cast.js.test;
+
+import com.ibm.wala.cast.js.html.DomLessSourceExtractor;
+import com.ibm.wala.cast.js.html.JSSourceExtractor;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class ExtractingToPredictableFileNames implements AutoCloseable {
```
msridhar: I think I get it, but can we add some Javadoc indicating the purpose of this class?

liblit: We can, we should, and I will. Thanks for not letting me be lazy about documentation, @msridhar.

PR #714, opened 2020-03-08
Travis CI job housekeeping

liblit:
Remove single-module Travis CI `cron` jobs. I can’t remember the last time these found a cross-module dependency defect, which is the whole reason we were running them.

Remove Travis CI Android configuration. We download the Android SDK ourselves as part of our Gradle build rules.

Improve Travis CI macOS base image selection so that we get the Java versions we want. Unfortunately this isn’t as easy as just using `jdk:` as on Linux builds.

Improve detection of current Java version for Travis CI jobs.


.travis.yml
```diff
@@ -39,167 +33,13 @@ env:
...
-      if: type = cron
+      jdk: openjdk11
     - os: osx
-      env: BUILD_ONLY_SUBMODULE=com.ibm.wala.shrike
-      if: type = cron
+      osx_image: xcode9.3  # Java 1.8
     - os: osx
-      env: BUILD_ONLY_SUBMODULE=com.ibm.wala.util
-      if: type = cron
+      osx_image: xcode9.4  # Java 13
```
msridhar: So is there no way to get JDK 11 on a Travis image?  It's good that this passes, but I don't know that we want to support anything WALA related on JDK 13 right now

msridhar: If this is the best we can do we can just stick with it

liblit: Something changed on Travis CI’s side that now makes JDK 11 unavailable in any macOS image. [This Travis CI job suite](https://travis-ci.org/liblit/WALA/builds/659932950) tried macOS with each of the [17 available macOS base images](https://docs.travis-ci.com/user/reference/osx/#macos-version) (`xcode{6.4,7.3,8{,.3},9{,.{1,2,3,4}},10{,.{1,2,3}},11{,.{1,2,3}}}`) and with explicit `jdk:` requests for either `openjdk11` or `oraclejdk11`. That’s 34 possible job configurations. Travis CI silently ignored exactly half of them. Of the 17 that ran, `xcode9.3` and earlier always ended up with Java 1.8, and `xcode9.4` and later always ended up with Java 13.

liblit: Others are [posting about this same problem](https://travis-ci.community/t/build-randomly-using-jdk-13-despite-osx-image-xcode10-1/6085/3).

PR #725, opened 2020-03-23
Switch from dx to d8 for converting jar to dex

msridhar:
`dx` is deprecated so this modernizes our test code.  Also, using `d8` means we don't need to pull in the Android SDK as a `testCompile` dependence in `com.ibm.wala.dalvik`, since `d8` can be pulled from a Maven repository.  We still require the Android SDK as a `testRuntime` dependence.

Apparently `d8` strips out all bytecode annotations that are not runtime retention (undocumented, but mentioned [here](https://github.com/DexPatcher/dexpatcher-tool/releases/tag/v1.8.0-alpha4)).  So, we change our test code for Dalvik annotations to only look for annotations with runtime retention.  Cleaned up some related code in the process.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -58,6 +58,8 @@ subprojects { subproject ->
 
 	repositories {
 		mavenCentral()
+		// to get r8
+		google()
 	}
 
 	eclipse {
--- a/build.gradle
+++ b/build.gradle
@@ -57,9 +57,8 @@ subprojects { subproject ->
 	version rootProject.version
 
 	repositories {
+		google() // to get r8
 		mavenCentral()
-		// to get r8
-		google()
 	}
 
 	eclipse {
--- a/com.ibm.wala.dalvik/build.gradle
+++ b/com.ibm.wala.dalvik/build.gradle
@@ -135,10 +135,10 @@ dependencies {
 	)
 	sampleCup 'java_cup:java_cup:0.9e:sources'
 	testImplementation(
+			'com.android.tools:r8:1.6.84',
 			'junit:junit:4.13',
 			'org.osgi:org.osgi.core:6.0.0',
 			'org.smali:dexlib2:2.4.0',
-			'com.android.tools:r8:1.6.84',
 			project(':com.ibm.wala.core'),
 			project(':com.ibm.wala.dalvik'),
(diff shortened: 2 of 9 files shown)


msridhar:
Getting weird Travis failures:
```
* What went wrong:
278Execution failed for task ':com.ibm.wala.cast.js.nodejs:downloadNodeJS'.
279> rate limit exceeded (HTTP status code: 403, URL: https://api.github.com/repos/nodejs/node/zipball/0a604e92e258c5ee2752d763e50721e35053f135)
```

Hopefully this is transitory.

msridhar:
Travis has passed now.  I don't know what is up with GitHub CI, but I don't think that should block us.

com.ibm.wala.dalvik/build.gradle
```diff
@@ -138,7 +138,7 @@ dependencies {
 			'junit:junit:4.13',
 			'org.osgi:org.osgi.core:6.0.0',
 			'org.smali:dexlib2:2.4.0',
-			files(installAndroidSdk.map { "${it.component('build-tools')}/lib/dx.jar" }),
+			'com.android.tools:r8:1.6.84',
```
liblit: Please move “`'com.android.tools:r8:1.6.84',`” to just before the “`'junit:junit:4.13',`” line. We keep dependencies in sorted order, in part to emphasize that their order does not matter, and in part to make duplicates easier to spot.

build.gradle
```diff
@@ -58,6 +58,8 @@ subprojects { subproject ->
 
 	repositories {
 		mavenCentral()
+		// to get r8
+		google()
```
liblit: Please order these repositories alphabetically. We keep repositories in sorted order, in part to emphasize that their order does not matter, and in part to make duplicates easier to spot.

```suggestion
		google() // to get r8
		mavenCentral()
```

msridhar:
@liblit I don't have time to do those simplifications right now, but they're a really good idea.  I created #726 to track that work, and assigned it to you for now.

msridhar:
Will rebase on #727 once it lands to fix CI

msridhar:
The Travis build actually passed, so I am going to merge

PR #733, opened 2020-03-27
Fix several Eclipse-detected resource leaks

liblit:
Plug several resource leaks detected by Eclipse. Suppress one such detection that is actually a false positive.

Other Eclipse leak warnings remain, but this is an initial step on the ones that I could fix in the obvious way.

### DIFF
--- a/travis/script-gradle
+++ b/travis/script-gradle
@@ -8,7 +8,7 @@ case "$(javac -version 2>&1)" in
   (javac\ 1.8.*)
      # linters only need to run on one operating system
      if [ "${TRAVIS_OS_NAME:-}" = linux ]; then
-       run_gradle -PjavaCompiler=ecj linters compileJava compileTestJava javadoc
+       run_gradle -PjavaCompiler=ecj linters compileJava compileTestJava compileTestFixturesJava compileTestSubjectsJava javadoc
      fi
      run_gradle build publishToMavenLocal
      ;;
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -50,7 +50,7 @@ jobs:
         with:
           java-version: 1.8
       - name: Compile Java using Gradle and ECJ
-        run: ./gradlew --continue --no-build-cache --no-daemon -PjavaCompiler=${{ matrix.java-compiler }} compileJava compileTestJava
+        run: ./gradlew --continue --no-build-cache --no-daemon -PjavaCompiler=${{ matrix.java-compiler }} compileJava compileTestJava compileTestFixturesJava compileTestSubjectsJava
         if: matrix.java-compiler == 'ecj'
       - name: Build and test using Gradle and default JDK compiler
         run: ./gradlew --continue --no-build-cache --no-daemon linters javadoc build
--- a/.idea/runConfigurations/Compile_using_ECJ.xml
+++ b/.idea/runConfigurations/Compile_using_ECJ.xml
@@ -12,6 +12,8 @@
         <list>
           <option value="compileJava" />
           <option value="compileTestJava" />
+	  <option value="compileTestFixturesJava" />
+	  <option value="compileTestSubjectsJava" />
         </list>
       </option>
       <option name="vmOptions" value="" />
--- a/.idea/runConfigurations/Compile_using_ECJ.xml
+++ b/.idea/runConfigurations/Compile_using_ECJ.xml
@@ -1,0 +1,22 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="Compile using ECJ" type="GradleRunConfiguration" factoryName="Gradle">
+    <ExternalSystemSettings>
+      <option name="executionName" />
+      <option name="externalProjectPath" value="$PROJECT_DIR$" />
+      <option name="externalSystemIdString" value="GRADLE" />
+      <option name="scriptParameters" value="-PjavaCompiler=ecj" />
+      <option name="taskDescriptions">
+        <list />
+      </option>
+      <option name="taskNames">
+        <list>
+          <option value="compileJava" />
+          <option value="compileTestJava" />
+        </list>
+      </option>
+      <option name="vmOptions" value="" />
+    </ExternalSystemSettings>
+    <GradleScriptDebugEnabled>true</GradleScriptDebugEnabled>
+    <method v="2" />
+  </configuration>
+</component>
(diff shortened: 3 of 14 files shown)


msridhar:
How could this regress? Don’t we compile with ECJ on CI now?

liblit:
Good point, @msridhar. We do compile with ECJ, both [using Travis CI](https://github.com/wala/WALA/blob/master/travis/script-gradle#L11) and [using GitHub actions](https://github.com/wala/WALA/blob/master/.github/workflows/continuous-integration.yml#L52). But where is that ECJ coming from? Perhaps the ECJ we are using during CI is from a different (older) Eclipse release than the latest Eclipse 2020-03 that I am using on my desktop. I know [how we request ECJ at the Gradle level](https://github.com/wala/WALA/blob/master/build.gradle#L112). But I haven't looked closely at what happens below that level. I'd be happy to look into this further, either as a separate task or as a prerequisite before accepting this pull request.

msridhar:
I think I figured out, based on [this page](https://github.com/TwoStone/gradle-eclipse-compiler-plugin).  I pushed a `build.gradle` change that, if I add to master, causes the lint job to fail with a detected resource leak.  Let's see what it does here.

liblit:
Nicely done, @msridhar. How did you know to pick version `3.21.0` specifically?

msridhar:
Haha, just pushed a comment regarding the version number 😄

liblit:
FYI, if this ECJ behaves the same as Eclipse 2020-03 on my desktop, then the ECJ continuous-integration tests should *not* pass for this pull request. As noted in my original pull request description, "Other Eclipse leak warnings remain, but this is an initial step on the ones that I could fix in the obvious way."

I think it would be reasonable to hold this pull request back until those ECJ tests do pass. That is, I should finish the job I started for the leaks that I could *not* "fix in the obvious way."

msridhar:
Ah, ok.  I am ok with breaking things up and landing fixes in batches, and only bumping the CI ECJ version once everything is either fixed or suppressed.  Alternately, we can try to fix / suppress everything in this PR.  @liblit I'll leave it to you to decide

msridhar:
@liblit I tweaked the comment one more time to clarify.  Oddly, if I remove one "unnecessary" suppression in Shrike, the `./gradlew -PjavaCompiler=ecj linters compileJava compileTestJava` command passes for me.  As far as I can tell, I have chosen the latest version of ECJ, but maybe there are further inconsistencies as compared to the compiler in Eclipse 2020-03?

liblit:
Oh dear. If we can't figure out how to get the same compiler in Eclipse on the desktop and as the Gradle-driven ECJ ... then I'm not sure how we can do anything constructive. Bleh.

msridhar:
Well, bumping the ECJ version definitely caught _more_ things, like at least one of these resource leaks.  So it seems like it's a win to do that for now.

If you get time, if you could paste an issue you see in the Eclipse IDE that isn't getting caught by Gradle-driven ECJ, I'm curious how it looks.

liblit:
@msridhar wrote:
> maybe there are further inconsistencies as compared to the compiler in Eclipse 2020-03?

If I bring up Eclipse 2020-03's Help → About Eclipse IDE → Installation Details → Features, then look for "Eclipse Java Development Tools", I see that it is using version 3.18.300.v20200305-0155 of that feature. So maybe that tells us which ECJ version to request in Gradle to get matching behavior from the leak analysis? @msridhar, can you give that a try?

msridhar:
3.18.3 doesn't work, unfortunately.  The Gradle plugin is pulling in this artifact:

https://search.maven.org/artifact/org.eclipse.jdt/ecj

The latest version, released March 2020, is 3.21.0.  So I think that's what Eclipse must be using.  Maybe the differing warnings are due to different defaults in ECJ command line vs. how it gets configured in the IDE?

liblit:
I see a version 3.18.0 at https://search.maven.org/artifact/org.eclipse.jdt/ecj. My guess is that's the closest to the 3.18.300.v20200305-0155 in my Eclipse 2020-03 installation.

msridhar:
> I see a version 3.18.0 at https://search.maven.org/artifact/org.eclipse.jdt/ecj. My guess is that's the closest to the 3.18.300.v20200305-0155 in my Eclipse 2020-03 installation.

The release date on that version is June 16, 2019, so that doesn't seem right.  My guess is the ECJ version is just not aligned with other JDT version numbers for some reason.

liblit:
I think I’m done for today, but the next time I can a look at this I plan to:

1. Fix or suppress all remaining Java compilation warnings (most of which are leak warnings) reported by Eclipse 2020-03.

2. Systematically experiment with different ECJ versions to see if I find one that also reports no Java warnings on the same code.

Maybe I’ll find an ECJ version that matches Eclipse; maybe not. Should be informative either way. Hopefully we come out of this with a better understanding of how to run ECJ CI jobs that more closely reflect what Eclipse users can expect to see.

Once we see how the above plays out, we can then consider what to do about any additional warnings reported by the most recent ECJ but not by the most recent Eclipse.

msridhar:
Thanks, @liblit.  Let me know if I can help.

liblit:
@msridhar, you were right. After fixing all Java compilation errors in Eclipse 2020-03, standalone ECJ 3.21.0 also has no complaints. ECJ 3.{17,18,19,20}.0 still report problems. So standalone ECJ 3.21.0 is the right version for us to use in CI. All CI tests are now passing. I'm only seeing GitHub results reported above for some reason, but [Travis CI is happy too](https://travis-ci.org/github/liblit/WALA/builds/668189404). I think this is ready to merge.

travis/script-gradle
```diff
@@ -8,7 +8,7 @@ case "$(javac -version 2>&1)" in
   (javac\ 1.8.*)
      # linters only need to run on one operating system
      if [ "${TRAVIS_OS_NAME:-}" = linux ]; then
-       run_gradle -PjavaCompiler=ecj linters compileJava compileTestJava javadoc
+       run_gradle -PjavaCompiler=ecj linters compileJava compileTestJava compileTestFixturesJava compileTestSubjectsJava javadoc
```
msridhar: Could you update `CONTRIBUTING.md` to reflect the change in this command?  To avoid the need for this consistency check, it would be nice to create a top-level Gradle linting task that enables ECJ and runs all of these listed tasks.  But that can happen in a follow-up PR.

liblit: Thanks for catching that. Added in a552e86ceb92d037d48792e9ea594e54fc8de7ec.

PR #744, opened 2020-03-31
Run ECJ compiler as part of Gradle `check` tasks

### LINKED ISSUE #736: Integrate ECJ compilation into regular build valuation
Currently, compiling with ECJ instead of the standard `javac` requires a separate `./gradlew -PjavaCompiler=ecj ...` run. This makes thorough testing cumbersome: there is no one Gradle task and no one Gradle command line that can be used to perform all build validation steps we want.

Gradle has good abstractions for Java tool chains, which can be set on a task-by-task basis. In principle, it should be possible for both ECJ and standard compilation tasks to be active in a single Gradle run. They’d need different task names, different tool chains, and different output directories, but should otherwise coexist peacefully.

I have work in progress to make this happen and it’s looking good so far. I might need to request one small change from the maintainer of the ECJ Gradle plugin to make it easier for me to access the ECJ tool chain instance. But it all seems doable.

Question for @msridhar: assume I can rig this up. Which existing tasks would you want it to be a dependency of? The two obvious choices are `check` or `linters`. If triggered by `check`, then we will validate ECJ compilation as part of `./gradlew build` or `./gradlew check`, though not as part of `./gradlew test` as the latter is really just for running JUnit-style tests. Or if triggered by `linters` then we will validate ECJ compilation as part of `./gradlew linters` but not as part of `./gradlew check` or `./gradlew build`.

In the past, I favored putting all of the linters under `check` and therefore transitively under `build`. But @msridhar had different options and wanted the checks and the linters to be separate. Which would you prefer for ECJ compilation validation?

liblit:
We still use the default Java compiler for builds, tests, published jar archives, etc.  But now we *also* compile all Java using ECJ as part of the standard `check` validation tasks.

Set the `skipJavaUsingEcjTasks` project property if you want to skip ECJ compilation for some reason. For example, `./gradlew -PskipJavaUsingEcjTasks ...`.

Fixes #736.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -11,15 +11,6 @@ jobs:
           - ubuntu-latest
           - macos-latest
           - windows-latest
-        java-compiler:
-          - default
-          - ecj
-        exclude:
-          # testing ECJ compilation on any one OS is sufficient
-          - os: macos-latest
-            java-compiler: ecj
-          - os: windows-latest
-            java-compiler: ecj
       fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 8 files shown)


liblit:
Discarding in favor of #749.

PR #771, opened 2020-05-03
Re-enable Google Java Format checking on CI

msridhar:
It seems the `verifyGoogleJavaFormat` task was not running on CI.  I think it is because at the top level, the `build` task does not depend on `check`.  This PR re-enables the check and also fixes some formatting issues that have crept in.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -239,7 +239,7 @@ tasks.register('installGitHooks', Copy) {
 }
 
 // run all known linters
-tasks.register('check') {
+final check = tasks.register('check') {
 	group = 'verification'
 	dependsOn(
 			// 'lintGradle',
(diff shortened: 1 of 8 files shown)


msridhar:
@liblit there may be a better fix for this one Gradle-wise; I just hacked the Travis script to run the task directly

liblit:
Ah, sorry about that. The standard Gradle Java plugin creates `check` as a dependency of `build` in each subproject that loads the Java plugin. But the top-level `build.gradle` does not load the Java plugin, so that dependency relationship is never established. We create our own top-level `:check` task, but I neglected to have the top-level `:build` task depend on it.

All add the necessary Gradle code so that `:build` depends on `:check`. How important is it to you to suppress that on all CI platforms except one?

msridhar:
> Ah, sorry about that. The standard Gradle Java plugin creates `check` as a dependency of `build` in each subproject that loads the Java plugin. But the top-level `build.gradle` does not load the Java plugin, so that dependency relationship is never established. We create our own top-level `:check` task, but I neglected to have the top-level `:build` task depend on it.
> 
> I'll add the necessary Gradle code so that `:build` depends on `:check`. How important is it to you to suppress that on all CI platforms except one?

Not that important.  It can just always run as part of `:build`.

PR #775, opened 2020-05-18
#753 [Shrike] pop2 opcode not translated to Pop(2) instruction

### LINKED ISSUE #753: [Shrike] pop2 opcode not translated to Pop(2) instruction
Hello,
I currently got stuck understanding the PopInstruction size. It seems there are two possible values: 1 (single word) and 2 (double word) which are representing `pop` and `pop2` in the jvm bytecode respectively. When there is a method like this:
```java
public void twoWordPop() {
	Thread.activeCount();       // Pushes an int on the stack (single word)
	System.currentTimeMillis(); // Pushes a long on the stack (double word)
}
```
the following bytecode is compiled (output from eclipse internal class file viewer):
```
public void twoWordPop();
  0  invokestatic java.lang.Thread.activeCount() : int [31]
  3  pop
  4  invokestatic java.lang.System.currentTimeMillis() : long [15]
  7  pop2
  8  return
```
When I load the compiled class file with an OfflineInstrumenter, the MethodData holds the following instructions:
```
0: Invoke(STATIC,Ljava/lang/Thread;,activeCount,()I)
1: Pop(1)
2: Invoke(STATIC,Ljava/lang/System;,currentTimeMillis,()J)
3: Pop(1)
4: Return(V)
```
I am wondering why there is a Pop(1) for the pushed long value at index 3, since there is a `pop2` bytecode instruction in the compiled file. Is this correct or am I do something wrong?

RalfHerzog:
Hopefully fixes #753 but I think some other tests fail. Maybe it's because of some setup issues on my machine. Need external review


msridhar:
It looks like there are some tests failing, e.g., the `SlicerTests` in `wala.core`.  @RalfHerzog do you need help in debugging these test failures?

liblit:
Seeing no reply to [my earlier message](https://github.com/wala/WALA/pull/775#issuecomment-3194671142), I'm going to assume that this PR has been abandoned. @RalfHerzog (or anyone else), please let me know if we should reopen it at some future date.

PR #778, opened 2020-06-06
Upgrade to Gradle 6.5

liblit:
No required changes for us in this release. The new [file-system watching feature](https://blog.gradle.org/introducing-file-system-watching) is interesting, though. In my ad hoc experiments, turning this on reduced the incremental-rebuild-after-no-changes time from about 3 seconds to about 2.5 seconds.

I haven't turned this feature on by default, though, since it's still experimental.


liblit:
Huh, a check from DeepCode. @msridhar, did you enable that?

msridhar:
> Huh, a check from DeepCode. @msridhar, did you enable that?

Yeah figured I’d give it a try 🙂

PR #779, opened 2020-06-06
Help Gradle cache the verifyGoogleJavaFormat task

liblit:
This works around a missing feature that [has already been requested for
the Gradle google-java-format plugin](https://github.com/sherter/google-java-format-gradle-plugin/issues/43).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -229,6 +229,13 @@ googleJavaFormat {
 
 tasks.named('verifyGoogleJavaFormat') {
 	group 'verification'
+
+	// workaround for <https://github.com/sherter/google-java-format-gradle-plugin/issues/43>
+	final stampFile = project.layout.buildDirectory.file(name)
+	outputs.file stampFile
+	doLast {
+		stampFile.get().asFile.text = ''
+	}
 }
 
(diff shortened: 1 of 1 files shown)


msridhar:
I don’t really understand this change. Will the task still run again if a source file is modified?

liblit:
@msridhar wrote:
> I don’t really understand this change.

The standard `verifyGoogleJavaFormat` task is ineligible for caching because it produces no outputs. This change gives it an output, which then lets Gradle make a proper decision about whether that output is out-of-date with respect to the inputs.

> Will the task still run again if a source file is modified?

Yes. Fortunately `verifyGoogleJavaFormat` does correctly set its inputs to the collection of all Java files in the project. If any one of those changes contents, the task will rerun.

I've checked this empirically too. Before this change, repeatedly running `./gradlew verifyGoogleJavaFormat` always produced output showing that the task was being rerun:

```
1 actionable task: 1 executed
```

After this change, running that command a second time without Java changes instead prints:

```
1 actionable task: 1 up-to-date
```

However, modifying a Java file causes the next Gradle run to print:

```
1 actionable task: 1 executed
```

PR #788, opened 2020-07-20
Improvements to CallGraph2JSON

msridhar:
This PR includes various improvements to `CallGraph2JSON` for serializing call graphs built with the field-based call graph builder.  It also adds some tests for `CallGraph2JSON`.  Opportunistically clean up some related code (like removing prints, removing mutable static fields, and fixing a Javadoc warning).

I may land this without review as its on the critical path for a research project.  But, if any issues are spotted I can fix in a follow-up PR.


msridhar:
(note that commit history will be squashed when merging)

PR #793, opened 2020-07-25
Add method info to CallGraph2JSON output

msridhar:
We now output JSON in a format that includes information on the containing method for each call site, which makes determining things like method reachability from an entrypoint much easier.  Also, use Gson for serializing JSON rather than coding that ourselves.


msridhar:
I assume it's ok for WALA to add a dependence on [Gson](https://github.com/google/gson) given it's such a well-known library and we already depend on another well-known Google library, Guava.  @juliandolby please confirm this is ok from an IBM standpoint.

PR #799, opened 2020-08-18
Override the isA() method of CallStringContextPair 

ligenPro:
Override `com.ibm.wala.ipa.callgraph.propagation.cfa.CallStringContextSelector.CallStringContextPair#isA()` to prevent possible problems when dealing with reflection and other situations.

For example: in the case of N-CFA, `com.ibm.wala.analysis.reflection.GetClassContextInterpeter#understand()` will not match `JavaTypeContext`

I think this bug is relatively simple, so I did not add related test cases. If necessary, please contact me.

### DIFF
--- a/com.ibm.wala.core/src/main/java/com/ibm/wala/core/tests/callGraph/CallGraphTestUtil.java
+++ b/com.ibm.wala.core/src/main/java/com/ibm/wala/core/tests/callGraph/CallGraphTestUtil.java
@@ -19,9 +19,11 @@ import com.ibm.wala.ipa.callgraph.Entrypoint;
 import com.ibm.wala.ipa.callgraph.IAnalysisCacheView;
 import com.ibm.wala.ipa.callgraph.impl.Util;
 import com.ibm.wala.ipa.callgraph.propagation.InstanceKey;
+import com.ibm.wala.ipa.callgraph.propagation.PointerAnalysis;
 import com.ibm.wala.ipa.callgraph.propagation.SSAPropagationCallGraphBuilder;
 import com.ibm.wala.ipa.cha.IClassHierarchy;
 import com.ibm.wala.util.CancelException;
+import com.ibm.wala.util.collections.Pair;
 import com.ibm.wala.util.config.AnalysisScopeReader;
 import com.ibm.wala.util.io.FileProvider;
 import com.ibm.wala.util.perf.StopwatchGC;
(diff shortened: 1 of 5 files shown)


msridhar:
@ligenPro thanks for the contribution!  I am not super familiar with how exactly this `isA()` method is used, so I'm not sure this is a correct fix.  If you could provide one test case that shows a different analysis result before and after your change that would be helpful.  Thanks!

ligenPro:
@msridhar   First of all, if the `isA()` method of `CallStringContextPair` is not overridden, the problem will only occur when **dealing with reflection**. 
# Test code
My test case is as follows:
```
public class People {
    
        People p = new People();
        System.out.println(p.getClass());
    }
} 
```
My java program is as follows, most of the code comes from: `com.ibm.wala.examples.drivers.PDFCallGraph`:
- using N-CFA to build CallGraph, n=1
- using the default ReflectionOptions, which is `com.ibm.wala.ipa.callgraph.AnalysisOptions.ReflectionOptions#FULL`
```
public static Graph<CGNode> buildCallGraph(String appJar, File exclusionFile)
      throws WalaException, IllegalArgumentException, CancelException, IOException {
    AnalysisScope scope =
        AnalysisScopeReader.makeJavaBinaryAnalysisScope(
            appJar, exclusionFile != null ? exclusionFile : new File(CallGraphTestUtil.REGRESSION_EXCLUSIONS));

    ClassHierarchy cha = ClassHierarchyFactory.make(scope);

    Iterable<Entrypoint> entrypoints = Util.makeMainEntrypoints(scope, cha);
    
    //the default value of ReflectionOption is Full
    AnalysisOptions options = new AnalysisOptions(scope, entrypoints);

    // build the call graph
    com.ibm.wala.ipa.callgraph.CallGraphBuilder<InstanceKey> builder =
        Util.makeNCFABuilder(1, options, new AnalysisCacheImpl(), cha, scope);

    CallGraph cg = builder.makeCallGraph(options, null);
    return cg;
}
```
The CGNode Context corresponding to `java.lang.Object#getClass` is `CallStringContextPair`, the specific content is: 
image
- `CallString` comes from `com.ibm.wala.ipa.callgraph.propagation.cfa.CallStringContextSelector` 
- `JavaTypeContext` comes from `com.ibm.wala.analysis.reflection.GetClassContextSelector`

# IR difference before and after overwriting
As you already know, in the process of building the CallGraph, after the `Context` corresponding to the `CGNode` is obtained through the `ContextSelector`, the `IR` corresponding to the `CGNode` will be obtained in the `SSAPropagationCallGraphBuilder#processCallingConstraints()—>SSAPropagationCallGraphBuilder#getTargetPointerKey()` method.
In the process of obtaining `IR`, the following `SSAContextInterpreter` is passed:
image


## Before overwriting
## After overwriting

msridhar:
Thanks, @ligenPro!  If I understand correctly, this slightly modified example should be a good test:

```java
public class People {

    static void doSomething(Object o) {}  

        People p = new People();
        doNothing(p.getClass());
    }
}
```

With your change, I believe the final points-to set obtained for the parameter `o` of `doSomething` should be more precise?  Or maybe if there are further reflective operations it will lead to a difference in the call graph?

What I'm looking for is a unit test with an assertion, such that the assertion fails before your change and passes afterward.  I think you should be able to add another test method to [`ReflectionTest`](https://github.com/wala/WALA/blob/5f20c317f1d4de687bf4e665b5906fd2a514a3ec/com.ibm.wala.core/src/test/java/com/ibm/wala/core/tests/callGraph/ReflectionTest.java) to achieve this, maybe with an existing test input or maybe a new one.  Can you give it a shot?

msridhar:
Thanks @ligenPro for the contribution!

ligenPro:
You're welcome 😄 😄

ligenPro:
@msridhar You're welcome! 😃

PR #804, opened 2020-08-25
Add ability to serialize flow graph from JS field-based call graph builder to JSON

msridhar:
This required a refactoring to expose the `FlowGraph` in the relevant construction APIs.  Also, add the ability to print a nicer, source-level representation of `Vertex` nodes in the flow graph, rather than printing info about WALA SSA variables.


msridhar:
This required a refactoring to expose the `FlowGraph` in the relevant construction APIs.  Also, add the ability to print a nicer, source-level representation of `Vertex` nodes in the flow graph, rather than printing info about WALA SSA variables.

PR #830, opened 2020-10-07
Handle ARROW_FUNCTION in visitFunctionNode

izgzhen:
For example:

```js
var https=require('https');

https.get({}, (r)=>{}); // AST: v21 = dispatch v20:#get@18 v10,v22,v17:#null
```

But for:

```js
var https=require('https');

// AST: v21 = construct v24@16 v23:#example.js@42 exception:v2example.js [28->58] (line 2)
https.get({}, function (r) {}); // AST: v16 = dispatch v15:#get@17 v5,v17,v21
```

I think [this predicate](https://github.com/wala/WALA/blob/60af67444dafe4071cd571b538cc2aff78b9d0ca/com.ibm.wala.cast.js.rhino/src/main/java/com/ibm/wala/cast/js/translator/RhinoToAstTranslator.java#L1250) should consider `FunctionNode.ARROW_FUNCTION` as well.

### DIFF
--- a/com.ibm.wala.cast.js.rhino/src/main/java/com/ibm/wala/cast/js/translator/RhinoToAstTranslator.java
+++ b/com.ibm.wala.cast.js.rhino/src/main/java/com/ibm/wala/cast/js/translator/RhinoToAstTranslator.java
@@ -151,6 +151,10 @@ public class RhinoToAstTranslator implements TranslatorToCAst {
     }
   }
 
+  private int thisCounter = 0;
+  private final Map<JavaScriptTranslatorToCAst.WalkContext<WalkContext, Node>, String>
+      contextThisVarNameMap = HashMapFactory.make();
+
   /** context used for top-level script declarations */
   private static class ScriptContext extends FunctionContext {
     private final String script;
@@ -1223,8 +1227,24 @@ public class RhinoToAstTranslator implements TranslatorToCAst {
       return null;
     }
 
+    private String getContextThisVarName(
+        JavaScriptTranslatorToCAst.WalkContext<WalkContext, Node> context) {
+      String name = "%%saved_this_" + thisCounter;
+      thisCounter += 1;
+      contextThisVarNameMap.put(context, name);
+      return name;
+    }
+
     @Override
     public CAstNode visitFunctionNode(FunctionNode fn, WalkContext context) {
+      // Declare a unique "this" bind to the current context's this
+      // FIXME: Only create it when needed (e.g. for a nested arrow function)
+      String thisName = getContextThisVarName(context);
+      context.addNameDecl(
+          Ast.makeNode(
+              CAstNode.DECL_STMT,
+              Ast.makeConstant(new CAstSymbolImpl(thisName, JSAstTranslator.Any)),
+              visitKeywordLiteral(new KeywordLiteral(-1, -1, Token.THIS), context)));
       WalkContext child = new FunctionContext(context, fn);
       List<CAstNode> body = new ArrayList<>();
       body.add(visit(fn.getBody(), child));
@@ -1268,7 +1288,8 @@ public class RhinoToAstTranslator implements TranslatorToCAst {
 
       if (DEBUG) System.err.println(fne.getName());
 
-      if (fn.getFunctionType() == FunctionNode.FUNCTION_EXPRESSION) {
+      if (fn.getFunctionType() == FunctionNode.FUNCTION_EXPRESSION
+          || fn.getFunctionType() == FunctionNode.ARROW_FUNCTION) {
         CAstNode fun = Ast.makeNode(CAstNode.FUNCTION_EXPR, Ast.makeConstant(fne));
 
         context.addScopedEntity(fun, fne);
@@ -1370,6 +1391,15 @@ public class RhinoToAstTranslator implements TranslatorToCAst {
               arg.cfg().map(globalRef, globalRef);
               return globalRef;
             } else {
+              // Query context to decide if using the enclosing context is an arrow function that
+              // has associated "saved_this_X"
+              if (arg.top() instanceof FunctionNode) {
+                int curretnFunctionType = ((FunctionNode) arg.top()).getFunctionType();
+                if (curretnFunctionType == FunctionNode.ARROW_FUNCTION) {
+                  String thisVarName = getContextThisVarName(arg.getParent());
+                  return readName(arg, null, thisVarName);
+                }
+              }
               return Ast.makeNode(CAstNode.VAR, Ast.makeConstant("this"));
             }
           }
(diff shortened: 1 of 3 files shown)


msridhar:
Thanks for the contribution @izgzhen!  I think we need a bit more, though.  Arrow functions differ semantically from function expressions in important ways.  See this article:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Description

At the least, I think we need to make sure the binding of `this` for calls to arrow functions works correctly.  Are you interested in making further changes to make this work properly?  If so, I can try to guide you to the relevant code.

izgzhen:
> At the least, I think we need to make sure the binding of this for calls to arrow functions works correctly. Are you interested in making further changes to make this work properly? If so, I can try to guide you to the relevant code.

Ah, makes sense! I am happy to give it a try. Thanks!

msridhar:
Cool!  So we can learn something about what is needed based on how Babel translates away arrow functions.  There's an example [here](https://babeljs.io/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=G4QwTgBA9gRgVhAvBA3gKApiBjKBXAOwBcIAuCARgAYAaDLAEygGUoBbAUyIAsBLAgOYAZEEQ6RyAM0LYivKAQAUAShQQA9OuiSc-MAGcONCODBQA7hGkFZ8gvpNgOEAlBL68vMQytRInHigGfXpMUKxMQyIAFV5OfCJFCBUkAD5UDS19fmxnHmdTCysZOQUIcxAHbCdRDh9zLz4CCHyIACJYODbjLxN9D04HVs6OWQByBzaeXn028IiI6f0AOlxCIgBqDYBueYXMXHsoABsOZeOoAUUl1fxiZV39zABfYwBmKioH8Oe0X7ROssmKwAk1hLUwCptkA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=script&lineWrap=true&presets=env%2Ces2015%2Cenv&prettier=false&targets=&version=7.11.6&externalPlugins=).  Given the following input (from the MDN page):
```javascript
```

izgzhen:
(nothing important -- just that I was a bit busy recently -- hope to look into this issue around this weekend...)

izgzhen:
@msridhar I made some attempts. It is still a bit hacky, but want to hear your opinions first. Also, I was modifying an old test in order to trigger the code path, but I think I should add a new test for it. Is there a more complete guide on how to add a new unit test? I tried duplicating the class `com.ibm.wala.cast.js.test.TestForInLoopHack`, but my IntelliJ IDE suggests "No tests available" when I run to launch it.

com.ibm.wala.cast.js.rhino/src/main/java/com/ibm/wala/cast/js/translator/RhinoToAstTranslator.java
```diff
@@ -1225,6 +1225,13 @@ private String getParentName(AstNode fn) {
 
     @Override
     public CAstNode visitFunctionNode(FunctionNode fn, WalkContext context) {
+      // Declare a unique "this" bind to the current context's this
+      String thisName = "this_" + context.top().hashCode();
```
msridhar, line 1229: I wouldn't use `hashCode()` here, as it will be non-deterministic and make debugging harder.  There should only be one of these variables declared per scope, and we don't care about being able to serialize back to JS code.  So you could use a name like `%%saved_this`, which is guaranteed not to create name conflicts since `%` is not a legal character for a JS identifier.  You could then add a counter variable at the end, like `%%saved_this_1`, etc., with the counter stored as an instance field in `RhinoToAstTranslator`.

msridhar, line 1229: It would be cleaner if we only created this variable when we needed it (i.e., when there is a nested arrow function).  But if that's tricky, as a first cut I am ok with creating it always.

com.ibm.wala.cast.js.rhino/src/main/java/com/ibm/wala/cast/js/translator/RhinoToAstTranslator.java
```diff
@@ -1349,6 +1357,15 @@ public CAstNode visitKeywordLiteral(KeywordLiteral node, WalkContext arg) {
               arg.cfg().map(globalRef, globalRef);
               return globalRef;
             } else {
+              // Query context to decide if using the enclosing context defined "this_X".
+              if (arg.top() instanceof FunctionNode) {
+                int curretnFunctionType = ((FunctionNode) arg.top()).getFunctionType();
+                if (curretnFunctionType == FunctionNode.ARROW_FUNCTION) {
+                  // WIP: we can probably use a better approach than querying parent context's hash code...
```
msridhar, line 1364: Yes we should store the variable name on the context somewhere...

msridhar (commented):
Couple initial comments

msridhar:
> @msridhar I made some attempts. It is still a bit hacky, but want to hear your opinions first. 

It's a good start!  I added a couple initial comments.

> Also, I was modifying an old test in order to trigger the code path, but I think I should add a new test for it. Is there a more complete guide on how to add a new unit test? I tried duplicating the class `com.ibm.wala.cast.js.test.TestForInLoopHack`, but my IntelliJ IDE suggests "No tests available" when I run to launch it.

I wouldn't write a new test class for testing this.  I would instead add new test methods to [`TestSimpleCallGraphShapeRhino`](https://github.com/wala/WALA/blob/master/com.ibm.wala.cast.js.rhino/src/test/java/com/ibm/wala/cast/js/test/TestSimpleCallGraphShapeRhino.java), following the style of [`TestSimpleCallGraphShape`](https://github.com/wala/WALA/blob/master/com.ibm.wala.cast.js/src/testFixtures/java/com/ibm/wala/cast/js/test/TestSimpleCallGraphShape.java).  You can add the test input `.js` files to [this folder](https://github.com/wala/WALA/tree/master/com.ibm.wala.cast.js/src/test/resources/tests).  I would write the tests in such a way that the call graph will only be correct if arrow functions are handled correctly.

com.ibm.wala.cast.js.rhino/src/main/java/com/ibm/wala/cast/js/translator/RhinoToAstTranslator.java
izgzhen: Right, I will leave a FIXME and try to address it after the initial validation

izgzhen:
> > @msridhar I made some attempts. It is still a bit hacky, but want to hear your opinions first.
> 
> It's a good start! I added a couple initial comments.
> 
> > Also, I was modifying an old test in order to trigger the code path, but I think I should add a new test for it. Is there a more complete guide on how to add a new unit test? I tried duplicating the class `com.ibm.wala.cast.js.test.TestForInLoopHack`, but my IntelliJ IDE suggests "No tests available" when I run to launch it.
> 
> I wouldn't write a new test class for testing this. I would instead add new test methods to [`TestSimpleCallGraphShapeRhino`](https://github.com/wala/WALA/blob/master/com.ibm.wala.cast.js.rhino/src/test/java/com/ibm/wala/cast/js/test/TestSimpleCallGraphShapeRhino.java), following the style of [`TestSimpleCallGraphShape`](https://github.com/wala/WALA/blob/master/com.ibm.wala.cast.js/src/testFixtures/java/com/ibm/wala/cast/js/test/TestSimpleCallGraphShape.java). You can add the test input `.js` files to [this folder](https://github.com/wala/WALA/tree/master/com.ibm.wala.cast.js/src/test/resources/tests). I would write the tests in such a way that the call graph will only be correct if arrow functions are handled correctly.

Added a new test and made various fixes. Sorry for the procrastination ;)

PR #834, opened 2020-10-30
JavaScript: handle multiple function expressions with the same name in the same scope

msridhar:
Before, WALA would simply drop one of the functions with the same name, which was clearly incorrect.  This fix detects the case of a duplicate name in the same scope, and uses a modified unique name (appending the character offset) for any duplicates.  Ideally we would have a consistent naming convention for all function expressions, whether there are duplicates or not.  But modifying the naming scheme in the non-duplicate case could break clients, so we avoid that for now.


msridhar:
FYI @juliandolby

PR #848, opened 2020-12-26
Fix handling of default case in switch statements

msridhar:
Before, the explicit default case in the generated CFG was unreachable, leading to the code being deleted completely from the IR.  Add some comments and a JavaScript test.  Presumably Java (and other CAst languages) are also affected by this fix.

### DIFF
--- a/com.ibm.wala.cast/src/main/java/com/ibm/wala/cast/ir/translator/AstTranslator.java
+++ b/com.ibm.wala.cast/src/main/java/com/ibm/wala/cast/ir/translator/AstTranslator.java
@@ -4744,10 +4744,17 @@ public abstract class AstTranslator extends CAstVisitor<AstTranslator.WalkContex
     return hasCase;
   }
 
+  /**
+   * Translates a CAst SWITCH node using an {@link com.ibm.wala.ssa.SSASwitchInstruction} for the
+   * control flow
+   */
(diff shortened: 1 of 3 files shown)


com.ibm.wala.cast/src/main/java/com/ibm/wala/cast/ir/translator/AstTranslator.java
```diff
@@ -4757,41 +4764,56 @@ private void doSimpleSwitch(CAstNode n, WalkContext context, CAstVisitor<WalkCon
...
 
-    context.cfg().addInstruction(insts.GotoInstruction(context.cfg().currentInstruction, -1));
-    defaultHackBlock = context.cfg().getCurrentBlock();
+    // initialize the dummy default block with a goto instruction to -1 (its target index will get
+    // corrected in a later phase)
     context.cfg().newBlock(false);
+    context.cfg().addInstruction(insts.GotoInstruction(context.cfg().currentInstruction, -1));
+    dummyDefaultBlock = context.cfg().getCurrentBlock();
+    // make sure the dummy default block is reachable
+    context.cfg().addEdge(switchInstrBlock, dummyDefaultBlock);
```
msridhar, line 4788: This is the key control-flow edge that was missing before

PR #855, opened 2021-01-18
Handle returned values from Function.protoype.apply calls in ACG

msridhar:


msridhar:
CI failure was some weird issue updating Gradle caches, nothing to do with the change

Issue #892, opened 2021-06-10
Updating to Gradle 6.7 or above causes compile error in `buildSrc`


msridhar:
If I run `./gradlew wrapper --gradle-version 7.0.2` and then `./gradlew --dry-run`, I get errors when compiling `buildSrc`:

```
> Task :buildSrc:compileJava FAILED
/Users/msridhar/git-repos/WALA/buildSrc/src/main/java/com/ibm/wala/gradle/EclipseJavaCompilerFactory.java:5: error: cannot find symbol
import org.gradle.api.internal.tasks.compile.JavaCompilerFactory;
                                            ^
  symbol:   class JavaCompilerFactory
  location: package org.gradle.api.internal.tasks.compile
/Users/msridhar/git-repos/WALA/buildSrc/src/main/java/com/ibm/wala/gradle/EclipseJavaCompilerFactory.java:11: error: cannot find symbol
class EclipseJavaCompilerFactory implements JavaCompilerFactory {
                                            ^
  symbol: class JavaCompilerFactory
/Users/msridhar/git-repos/WALA/buildSrc/src/main/java/com/ibm/wala/gradle/EclipseJavaCompilerToolChain.java:5: error: cannot find symbol
import org.gradle.api.internal.tasks.AbstractJavaToolChain;
                                    ^
  symbol:   class AbstractJavaToolChain
  location: package org.gradle.api.internal.tasks
/Users/msridhar/git-repos/WALA/buildSrc/src/main/java/com/ibm/wala/gradle/EclipseJavaCompilerToolChain.java:15: error: cannot find symbol
public class EclipseJavaCompilerToolChain extends AbstractJavaToolChain {
                                                  ^
  symbol: class AbstractJavaToolChain
/Users/msridhar/git-repos/WALA/buildSrc/src/main/java/com/ibm/wala/gradle/EclipseJavaCompilerFactory.java:27: error: method does not override or implement a method from a supertype
  @Override
  ^
/Users/msridhar/git-repos/WALA/buildSrc/src/main/java/com/ibm/wala/gradle/EclipseJavaCompilerToolChain.java:22: error: method does not override or implement a method from a supertype
  @Override
  ^
/Users/msridhar/git-repos/WALA/buildSrc/src/main/java/com/ibm/wala/gradle/EclipseJavaCompilerToolChain.java:27: error: method does not override or implement a method from a supertype
  @Override
  ^
7 errors
```

[...]

PR #893, opened 2021-06-12
Minor Gradle 6 updates for better future-compatibility

liblit:
These tweaks don't actually allow us to upgrade to Gradle 7: see wala/WALA#892. But they do smooth the path for a future upgrade by modernizing some things that are deprecated in Gradle 6 or 7.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -90,7 +90,7 @@ subprojects { subproject ->
 
 		if (project.hasProperty('excludeSlowTests')) {
 			dependencies {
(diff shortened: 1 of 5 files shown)


msridhar (approved):

msridhar:
@liblit I can't remember, do you still like to not squash your commits?  I'll let you merge as you see fit 🙂

liblit:
I rebase if it takes me a few iterations to get one change right. In this case we have several changes, each correct on the first try, each independent from the others. Not a situation that I would squash.￼

PR #898, opened 2021-06-20
[WIP] Fix for updated Eclipse dependencies that require JDK 11

msridhar:
WALA master no longer builds on JDK 8, due to the fact that some Eclipse dependencies got updated to versions that require JDK 11.  This PR fixes the compilation problem, but it's kind of a hack and I hope there is a better approach.

@liblit I'd appreciate it if you could take a look at this and answer a couple of questions:
1. Why does the WALA build even pull new versions of these dependencies?  I don't understand how the `com.diffplug.eclipse.mavencentral` plugin works, but is there any way we can tell it to "freeze" the version of the Eclipse dependencies that we use and not update them unless we specifically ask?  Unintended updates to transitive dependencies have bit us before.
2. What is the "right" way to fix this?  My change works by forcing old versions of jars, which works for compiling, but at runtime things are messed up (the `com.ibm.wala.cast.java.ecj` test is failing as there are still more jars that require JDK 11).  We need a way to get a consistent set of Eclipse jars in the dependencies, while preventing them from being uploaded without us asking.

I meant to cut a new release today but got caught up in this issue.  We need to figure out a way to make things more dependable.  Maybe we need to not use `com.diffplug.eclipse.mavencentral`, at least for the ECJ module, or put in place some kind of transitive dependency freezing.  
 
Anyway thoughts welcome here.

### DIFF
--- a/com.ibm.wala.cast.java.ecj/build.gradle
+++ b/com.ibm.wala.cast.java.ecj/build.gradle
@@ -20,6 +20,9 @@ dependencies {
 			project(':com.ibm.wala.shrike'),
 			project(':com.ibm.wala.util'),
 	)
+	implementation('org.eclipse.platform:org.eclipse.equinox.common') {
+		version { strictly '3.14.100' }
+	}	
 	testImplementation(
 			'junit:junit:4.13.2',
 			testFixtures(project(':com.ibm.wala.cast.java')),
--- a/com.ibm.wala.ide.jdt.test/build.gradle
+++ b/com.ibm.wala.ide.jdt.test/build.gradle
@@ -33,6 +33,12 @@ dependencies {
 	testImplementation('javax.annotation:javax.annotation-api') {
 		version { strictly '1.3.2' }
 	}
+	testImplementation('org.eclipse.platform:org.eclipse.core.resources') {
+		version { strictly '3.14.0' }
+	}
+	testImplementation('org.eclipse.platform:org.eclipse.equinox.common') {
+		version { strictly '3.14.100' }
+	}
 }
 
 tasks.named('test') {
--- a/com.ibm.wala.ide.jdt/build.gradle
+++ b/com.ibm.wala.ide.jdt/build.gradle
@@ -32,4 +32,10 @@ dependencies {
 	implementation('javax.annotation:javax.annotation-api') {
 		version { strictly '1.3.2' }
 	}
+	implementation('org.eclipse.platform:org.eclipse.core.resources') {
+		version { strictly '3.14.0' }
+	}
+	implementation('org.eclipse.platform:org.eclipse.equinox.common') {
+		version { strictly '3.14.100' }
+	}
 }
--- a/com.ibm.wala.ide.jsdt.tests/build.gradle
+++ b/com.ibm.wala.ide.jsdt.tests/build.gradle
@@ -38,6 +38,12 @@ dependencies {
 	testImplementation('javax.annotation:javax.annotation-api') {
 		version { strictly '1.3.2' }
 	}
+	testImplementation('org.eclipse.platform:org.eclipse.core.resources') {
+		version { strictly '3.14.0' }
+	}
+	testImplementation('org.eclipse.platform:org.eclipse.equinox.common') {
+		version { strictly '3.14.100' }
+	}
 }
 
 tasks.named('test') {
--- a/com.ibm.wala.ide.jsdt/build.gradle
+++ b/com.ibm.wala.ide.jsdt/build.gradle
@@ -41,6 +41,12 @@ dependencies {
 	implementation('javax.annotation:javax.annotation-api') {
 		version { strictly '1.3.2' }
 	}
+	implementation('org.eclipse.platform:org.eclipse.core.resources') {
+		version { strictly '3.14.0' }
+	}
+	implementation('org.eclipse.platform:org.eclipse.equinox.common') {
+		version { strictly '3.14.100' }
+	}
 	testFixturesImplementation('javax.annotation:javax.annotation-api') {
 		version { strictly '1.3.2' }
 	}
--- a/com.ibm.wala.ide.tests/build.gradle
+++ b/com.ibm.wala.ide.tests/build.gradle
@@ -35,5 +35,10 @@ dependencies {
 	testImplementation('javax.annotation:javax.annotation-api') {
 		version { strictly '1.3.2' }
 	}
-
+	testImplementation('org.eclipse.platform:org.eclipse.core.resources') {
+		version { strictly '3.14.0' }
+	}
+	testImplementation('org.eclipse.platform:org.eclipse.equinox.common') {
+		version { strictly '3.14.100' }
+	}
 }
--- a/com.ibm.wala.ide/build.gradle
+++ b/com.ibm.wala.ide/build.gradle
@@ -32,6 +32,12 @@ dependencies {
 	implementation('javax.annotation:javax.annotation-api') {
 		version { strictly '1.3.2' }
 	}
+	implementation('org.eclipse.platform:org.eclipse.core.resources') {
+		version { strictly '3.14.0' }
+	}
+	implementation('org.eclipse.platform:org.eclipse.equinox.common') {
+		version { strictly '3.14.100' }
+	}	
 	testFixturesImplementation('javax.annotation:javax.annotation-api') {
 		version { strictly '1.3.2' }
 	}


liblit:
**Note:** everything I run below uses WALA master *without* the changes proposed in this pull request. It may well be that this pull request’s proposed changes are what we need, but for helping me understand what’s going on, it was cleaner to omit these changes in what follows.

Also, I appologize for the long and rambling nature of this message. It is, in part, a log of what I explored, tried, and learned along the way. **The TL;DR version is that I don’t understand everything, but I understand enough to offer a promising alternative that passes all automated tests while touching just a dozen lines of Gradle logic in a single file: #899.**

---

What we might think of as a single Eclipse release actually consists of dozens of individual artifacts. Each artifact has its own version number and dependencies on other artifacts, none of which bear any obvious resemblance to the overall version number of Eclipse itself. So you might intend to use "stuff from [Eclipse 4.15.0](https://projects.eclipse.org/projects/eclipse/releases/4.15.0)", but that’s a very high-level request that needs to be mapped down to specific versions of the many artifacts that were part of that release.

This is where the `com.diffplug.eclipse.mavencentral` plugin comes in. It lets us request a single high-level Eclipse version and it maps that down to versions of the various artifacts that made up that release. Once that mapping has been done, then Gradle’s regular mechanisms kick in for resolving, finding, and downloading artifacts from Maven Central. This is already a big improvement over the bad old days when Eclipse artifacts were not on Maven Central at all, but instead were only available from p2 repositories that nobody but Eclipse insiders really understands or supports.

Given that as background, here is where we set a project field to the Eclipse version that we want to use throughout:

https://github.com/wala/WALA/blob/44e276ebb57f28cb9ffbfc9901720afcb841b4c4/build.gradle#L37

There’s nothing magical about the `eclipseVersion` name: it’s just the arbitrary name we chose for a slot where we stash the version string. Any subproject can fetch that value later and know that it’s got the same value as seen in every other project.

Here’s a typical use of that version string:

https://github.com/wala/WALA/blob/44e276ebb57f28cb9ffbfc9901720afcb841b4c4/com.ibm.wala.ide/build.gradle#L8-L25

Notice the use of `release eclipseVersion`, which here is effectively `release '4.15.0'`. Each of the lines that follows lists some artifact that this subproject will need. The `com.diffplug.eclipse.mavencentral` plugin looks up the appropriate version of that artifact, and also determines each artifact’s Maven group. It then adds a dependency on the given artifact, with the group and version filled in for us.

That’s how it’s all *supposed* to work, at least.

---

As @msridhar noted, it seems that we’ve recently started picking up newer versions of some artifacts: too new for us, since we’re stuck at Java 1.8. Unfortunately, I don’t actually know where the `com.diffplug.eclipse.mavencentral` plugin gets its mapping from Eclipse version numbers to Maven artifact version numbers. From looking at some of the source code of that plugin I can tell that this information is downloaded from *somewhere* and cached under `~/.goomph/eclipse-release-metadata`. For example, I have a `~/.goomph/eclipse-release-metadata/4.15.0/artifacts.jar` file that contains a single `artifacts.xml` file. To pick one artifact as our focus, here’s what this Eclipse 4.15.0 `artifacts.xml` file tells us we should use for `org.eclipse.core.resources`:

```xml
<artifact classifier='osgi.bundle' id='org.eclipse.core.resources' version='3.13.700.v20200209-1624'>
  <properties size='9'>
    <property name='artifact.size' value='894121'/>
    <property name='download.size' value='894121'/>
    <property name='maven-groupId' value='eclipse.platform.resources'/>
    <property name='maven-artifactId' value='org.eclipse.core.resources'/>
    <property name='maven-version' value='3.13.700-SNAPSHOT'/>
    <property name='download.stats' value='org.eclipse.core.resources/3.13.700.v20200209-1624'/>
    <property name='download.md5' value='944de742270e1c24858059e22c6c37aa'/>
    <property name='download.checksum.md5' value='944de742270e1c24858059e22c6c37aa'/>
    <property name='download.checksum.sha-256' value='ce021447dbea30a4e5ddb3f52534cd2794fb52855071b8dcf257b936ab162168'/>
  </properties>
</artifact>
```

Notice the `maven-version` property. As far as I can tell, if we actually did use version 3.13.x of `org.eclipse.core.resources`, we’d be in fine shape. It looks like version 3.15.x of this artifact is where Java 1.8 incompatibility first arrives. And in fact, when Gradle fails to compile this subproject, it does so with a complaint that it’s trying (but failing) to use version 3.15.x of this artifact:

```
  bad class file: /home/liblit/.gradle/caches/modules-2/files-2.1/org.eclipse.platform/org.eclipse.core.resources/3.15.0/a1964a49431ffd325a20295f30ef3a0b595409b8/org.eclipse.core.resources-3.15.0.jar(org/eclipse/core/resources/IFile.class)
    class file has wrong version 55.0, should be 52.0
    Please remove or make sure it appears in the correct subdirectory of the classpath.
```

---

So how did we go from version 3.13.x (which should work) to version 3.15.x (which does not)? I don’t know. I hoped that Gradle could explain itself a bit better. In the `com.ibm.wala.ide` subdirectory, I ran the command:

```shell
../gradlew dependencyInsight --dependency org.eclipse.platform:org.eclipse.core.resources
```

The output is a bit more readable on a color terminal, but here’s what Gradle told me:

```
org.eclipse.platform:org.eclipse.core.resources:3.15.0
   variant "compile" [
      org.gradle.status              = release (not requested)
      org.gradle.usage               = java-api
      org.gradle.libraryelements     = jar (compatible with: classes)
      org.gradle.category            = library

      Requested attributes not found in the selected variant:
         org.gradle.dependency.bundling = external
         org.gradle.jvm.version         = 8
   ]
   Selection reasons:
      - By conflict resolution : between versions 3.13.700 and 3.15.0

org.eclipse.platform:org.eclipse.core.resources:3.13.700 -> 3.15.0
\--- compileClasspath

org.eclipse.platform:org.eclipse.core.resources:[3.14.0,4.0.0) -> 3.15.0
+--- org.eclipse.jdt:org.eclipse.jdt.core:3.26.0
|    +--- compileClasspath (requested org.eclipse.jdt:org.eclipse.jdt.core:3.21.0)
|    +--- org.eclipse.pde:org.eclipse.pde.core:3.13.300 (requested org.eclipse.jdt:org.eclipse.jdt.core:[3.2.0,4.0.0))
|    |    \--- compileClasspath
|    +--- org.eclipse.jdt:org.eclipse.jdt.launching:3.19.200 (requested org.eclipse.jdt:org.eclipse.jdt.core:[3.26.0,4.0.0))
|    |    \--- org.eclipse.pde:org.eclipse.pde.core:3.13.300 (requested org.eclipse.jdt:org.eclipse.jdt.launching:[3.2.0,4.0.0)) (*)
|    \--- org.eclipse.jdt:org.eclipse.jdt.debug:3.17.200 (requested org.eclipse.jdt:org.eclipse.jdt.core:[3.26.0,4.0.0))
|         \--- org.eclipse.jdt:org.eclipse.jdt.launching:3.19.200 (requested org.eclipse.jdt:org.eclipse.jdt.debug:[3.17.0,4.0.0)) (*)
+--- org.eclipse.jdt:org.eclipse.jdt.debug:3.17.200 (*)
\--- org.eclipse.jdt:org.eclipse.jdt.launching:3.19.200 (*)

org.eclipse.platform:org.eclipse.core.resources:[3.3.0,4.0.0) -> 3.15.0
\--- org.eclipse.platform:org.eclipse.team.core:3.9.0
     \--- org.eclipse.pde:org.eclipse.pde.core:3.13.300 (requested org.eclipse.platform:org.eclipse.team.core:[3.2.0,4.0.0))
          \--- compileClasspath

org.eclipse.platform:org.eclipse.core.resources:[3.9.0,4.0.0) -> 3.15.0
\--- org.eclipse.platform:org.eclipse.debug.core:3.18.100
     +--- org.eclipse.pde:org.eclipse.pde.core:3.13.300 (requested org.eclipse.platform:org.eclipse.debug.core:[3.5.0,4.0.0))
     |    \--- compileClasspath
     +--- org.eclipse.jdt:org.eclipse.jdt.launching:3.19.200 (requested org.eclipse.platform:org.eclipse.debug.core:[3.17.0,4.0.0))
     |    \--- org.eclipse.pde:org.eclipse.pde.core:3.13.300 (requested org.eclipse.jdt:org.eclipse.jdt.launching:[3.2.0,4.0.0)) (*)
     \--- org.eclipse.jdt:org.eclipse.jdt.debug:3.17.200 (requested org.eclipse.platform:org.eclipse.debug.core:[3.12.0,4.0.0))
          \--- org.eclipse.jdt:org.eclipse.jdt.launching:3.19.200 (requested org.eclipse.jdt:org.eclipse.jdt.debug:[3.17.0,4.0.0)) (*)

(*) - dependencies omitted (listed previously)

A web-based, searchable dependency report is available by adding the --scan option.
```

**Note:** as the last line of output above suggests, if you add `--scan` to that Gradle command line, you’ll get an uploaded dependency report that is a bit easier to search and work with. Try it!

Nearly everything in the above output looks compatible with us using acceptable versions of `org.eclipse.core.resources`. Ignore the first section, which begins with `org.eclipse.platform:org.eclipse.core.resources:3.15.0`. After that, I see a request for `3.13.700-SNAPSHOT`. I see several other lines that request various version ranges as old as 3.14.0, 3.3.0, or 3.9.0, but all strictly before 4.0.0. However, in each of these cases, observe that the *resolved* version is 3.15.0:

```
org.eclipse.platform:org.eclipse.core.resources:3.13.700 -> 3.15.0
[...]

org.eclipse.platform:org.eclipse.core.resources:[3.14.0,4.0.0) -> 3.15.0
[...]

org.eclipse.platform:org.eclipse.core.resources:[3.3.0,4.0.0) -> 3.15.0
[...]

org.eclipse.platform:org.eclipse.core.resources:[3.9.0,4.0.0) -> 3.15.0
[...]
```

So what’s going on with the first section?  This first section seems to be the root of our problems:

```
org.eclipse.platform:org.eclipse.core.resources:3.15.0
   variant "compile" [
      org.gradle.status              = release (not requested)
      org.gradle.usage               = java-api
      org.gradle.libraryelements     = jar (compatible with: classes)
      org.gradle.category            = library

      Requested attributes not found in the selected variant:
         org.gradle.dependency.bundling = external
         org.gradle.jvm.version         = 8
   ]
   Selection reasons:
      - By conflict resolution : between versions 3.13.700 and 3.15.0
```

This is clearly the part of dependency resolution that is forcing us to the 3.15.0 version that is incompatible with Java 1.8.

[This Gradle documentation](https://docs.gradle.org/current/userguide/dependency_management.html#sub:resolution-strategy) describes Gradle’s dependency resolution strategy.  There are several mentions of using the "highest" available version of a dependency. So is Gradle picking 3.15 simply because it is present in Maven Central, is higher than 3.13 or 3.14, but still satisfies the "strictly before 4.0.0" constraints?  Perhaps so, but if Gradle is respecting the "strictly before 4.0.0" constraint, then why is it disregarding what would *seem* to be an exact version 3.13.700 constraint reported in the `dependencyInsight` output?

Well [this documentation](https://docs.gradle.org/current/userguide/dependency_resolution.html#sub:resolution-strategy) would make me think that the non-range version 3.13.700 should be preferred:

> As you have seen, Gradle supports a concept of rich version declaration, so what is the highest version depends on the way versions were declared:
>
> [...]
>
> * If ranges are involved:
> 
>   * If there is a non range version that falls within the specified ranges or is higher than their upper bound, it will be selected.

Now look back at this part of the `dependencyInsight` output:

```
org.eclipse.platform:org.eclipse.core.resources:3.13.700 -> 3.15.0
\--- compileClasspath

org.eclipse.platform:org.eclipse.core.resources:[3.14.0,4.0.0) -> 3.15.0
[...]
```

The first line is a non-range version: 3.13.700. But below that we have a *conflicting* request for something at version 3.14.0 or later (but strictly before 4.0.0).  **So now I don’t understand how Gradle is able to select any version of `org.eclipse.core.resources` at all: there’s no version that satisfies equality to 3.13.700 while also being in the half-closed range from 3.14.0 to 4.0.0.**

Also in the `dependencyInsight` output, I see that a requested "org.gradle.jvm.version = 8" attribute was "not found in the selected variant". I’m not surprised that this attribute is not found: these extra attributes are part of a rich dependency variant selection mechanism that Gradle offers.  Using it to its fullest requires extra artifact metadata that is not typically found in Maven repositories. Gradle can generate and publish this extra metadata when it builds artifats, but it’s more commonly absent than present in packages that others have posted to Maven Central. So presumably Gradle has some fallback mechanism for allowing requested attributes to be missing.  [This is Gradle’s documentation on how the variants system does the best it can when working with Maven repositories](https://docs.gradle.org/current/userguide/variant_model.html#sec:mapping-maven-ivy-to-variants), but it didn’t help me understand what Gradle’s doing about the missing `org.gradle.jvm.version` attribute speciically.

---

OK, so there’s some stuff going on that I don’t understand, specifically around how the plugin-provided requirement for `org.eclipse.core.resources` version 3.13.700 ends up being conflict-resolved to version 3.15.0.  But we can certainly provide additional constraints telling Gradle that version 3.15.0 is not acceptable.  In fact, we really need to do without anything from version 3.14 onward, since 3.14 seems to be when this specific artifact started requiring Java 11+.  We must do the same for `org.eclipse.equinox.common`, for the same reason: 3.14+ require Java 11+.  @msridhar already *mostly* did that by adding blocks of code like these to various subprojects’ `build.gradle` files:

```groovy
implementation('org.eclipse.platform:org.eclipse.core.resources') {
	version { strictly '3.14.0' }
}
implementation('org.eclipse.platform:org.eclipse.equinox.common') {
	version { strictly '3.14.100' }
}	
```

Unfortunately, @msridhar found that while this fixed compilation failures, there were still test failures. I’m not completely sure why (no surprise), but I think there were just other ways that too-new versions of these two artifacts were slipping into the class path.  @msridhar was on the right track; he just didn’t quite cover everything that needed it.

**Fortunately, there is a very nice way that we can provide these version constraints in a single, centralized place.**  Gradle has a mechanism for providing what we might think of as [“latent” version constraints](https://docs.gradle.org/current/userguide/dependency_constraints.html#sec:adding-constraints-transitive-deps): constraints on the version of an artifact that are only used *if* that artifact is indeed a dependency. So we can configure each of WALA’s Java-compiling subprojects such that *if* `org.eclipse.core.resources` or `org.eclipse.equinox.common` is needed, only versions before 3.14 will be allowed. Furthermore, we can do this in just a single place rather than in multiple `*/build.gradle` files.

---

Circling back to @msridhar’s questions:

> Why does the WALA build even pull new versions of these dependencies?

Gradle’s dependency resolution logic is complex; I don’t understand all of it. In general, Gradle prefers the newest version of a dependency that satisfies all constraints, so that tends to move you toward nwewer versions as they become available. That being said, I do not understand why the `com.diffplug.eclipse.mavencentral` plugin’s constraints do not lock Gradle into a single version of each of the Eclipse artifacts that we use.

> I don't understand how the `com.diffplug.eclipse.mavencentral` plugin works, but is there any way we can tell it to "freeze" the version of the Eclipse dependencies that we use and not update them unless we specifically ask?

Yes, we can impose additional constraints. In the extreme, we could specify the exact version of every Eclipse artifact that we need, effectivly leaving the `com.diffplug.eclipse.mavencentral` plugin doing very little, or even nothing at all.  We could stop using the `com.diffplug.eclipse.mavencentral` plugin entirely if we decide that we are taking over full responsibility for selecting all exact Eclipse artifact versions. We could also keep using this plugin but add our own, additional constraints to forbid Gradle from using specific artifact versions (or version ranges) that we determine to be problematic, such as `org.eclipse.core.resources` or `org.eclipse.equinox.common` versions 3.14+.

We could also use [Gradle dependency locking](https://docs.gradle.org/current/userguide/dependency_locking.html).  This seems to be what @msridhar is asking for. Keep in mind, though, that once we start locking dependencies, every *intentional* update to a newer release of some dependency will become something we need to shepherd manually. So we are trading one kind of work (debugging goofy decisions by the automatic dependency resolution logic) for a different kind of work (making conscious, explicit choices for each any every dependency update, no matter how frequent or how small).

> What is the "right" way to fix this?

I think that you were headed in the right direction with the additional `org.eclipse.core.resources` and `org.eclipse.equinox.common` version constraints. You just missed a few somewhere. I am not entirely sure where, but it does not matter. I have an alternative way of applying these constraints wherever needed, throughout all of WALA, with the constraint logic added only in one, central place: #899.

msridhar:
Closing, merged #899 instead

PR #899, opened 2021-06-27
Avoid Eclipse artifact versions that require Java 11+

liblit:
This pull request is an alternative to #898 that centralizes constraint configuration in one place for all subprojects.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,6 +66,18 @@ subprojects { subproject ->
 		}
 	}
 
+    dependencies.constraints {
+    	[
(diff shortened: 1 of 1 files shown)


msridhar:
Thank you so much @liblit for digging in and finding this fix!  And thanks also for the detailed explanation at https://github.com/wala/WALA/pull/898#issuecomment-869236017.  Will merge this to solve the problem for now.  Afterward, I think we need to investigate dependency locking, to prevent this kind of thing from happening going forward.

PR #900, opened 2021-06-29
Introduce dependency locking

msridhar:
This was the simplest way I could find to introduce [dependency locking](https://docs.gradle.org/current/userguide/dependency_locking.html), which should prevent issues like #898 / #899 in the future.  It adds a bunch of empty lock files for configurations without dependencies, but since they are auto-generated I don't think this is a big deal; it will be more fragile to try to only generate non-empty lock files.

For now, we use the `LENIENT` lock mode to work around failure of certain checks in the default mode; see the comments for details.  Even in `LENIENT` mode, dynamic dependencies should be fixed at the versions in the lock files.


msridhar:
Unfortunately I am a bit stuck here.  The `'org.eclipse.platform:org.eclipse.swt.*'` dependencies vary by OS, so we can't lock that one as far as I can see (I don't see a way to specify a disjunction of dependencies in the lockfile).  Gradle 6.7 did add a way to ignore a particular dependency, but we cannot even update to Gradle 6.7 due to #892...

@liblit if you have any ideas here that are simpler than fixing #892 please share 🙂

msridhar:
@liblit I was able to find a temporary workaround for the SWT issue by switching to lenient mode; see the updated description for the PR and comments in the `build.gradle` file.

build.gradle
```diff
@@ -429,3 +451,14 @@ tasks.register('downloads') {
...
+
+////////////////////////////////////////////////////////////////////////
+//
+//  Helpers for dependency locking
+//
+
+// this task resolves dependencies in all sub-projects, making it easy to
+// generate lockfiles
+allprojects {
+    task allDeps(type: DependencyReportTask) {}
```
liblit, line 463: Use [the task configuration avoidance APIs](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html) here:

```groovy
tasks.register('allDeps', DependencyReportTask) {}
```

Or can we remove this task entirely? According to [the `DependencyReportTask` documentation](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.diagnostics.DependencyReportTask.html),

> An instance of this type is used when you execute the `dependencies` task from the command-line.

If that's the case, then why do we need an extra `allDeps` task?

liblit (changes requested):
Remove `allDeps` task if possible. If needed, explain why and switch to task configuration avoidance API.

build.gradle
msridhar: I think the key here is that we are adding the task to `allprojects`, which includes all sub-projects.  When I run `./gradlew dependencies` at the top level, I get:
```
$ ./gradlew dependencies
Starting a Gradle Daemon (subsequent builds will be faster)

> Task :dependencies

------------------------------------------------------------
Root project
------------------------------------------------------------

No configurations
```

I added the `allDeps` task as a way to run the `dependencies` task on all sub-projects; just running `./gradlew allDeps` does the trick.  Do you know of a different way to do this that doesn't require adding another task?  In Googling around I didn't see anything other than equivalents of the `allDeps` task.

msridhar: > Use [the task configuration avoidance APIs](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html) here:

4a0a78a05fabcef9eaee6b991886e13d0d90d480

msridhar:
@liblit gentle ping on this

liblit (changes requested):

liblit:
> @liblit gentle ping on this

Oops, sorry. I started a review but apparently never finished and posted it.

build.gradle
```diff
@@ -429,3 +451,14 @@ tasks.register('downloads') {
...
+
+////////////////////////////////////////////////////////////////////////
+//
+//  Helpers for dependency locking
+//
+
+// this task resolves dependencies in all sub-projects, making it easy to
+// generate lockfiles
+allprojects {
+    tasks.register('allDeps', DependencyReportTask) {}
```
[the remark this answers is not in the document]
msridhar: > Why does this look so different from the advice given [here](https://docs.gradle.org/6.6.1/userguide/dependency_locking.html#generating_and_updating_dependency_locks)?

If you're referring to the `resolveAndLockAll` task defined there, I tried it, and it didn't work.  I think it only resolves all configurations for the current project, _not_ sub-projects.  I guess it helps if you have some configurations that do not get resolved in some cases, but it didn't help with the sub-project problem.  Gradle could use some improved documentation on this feature 🙂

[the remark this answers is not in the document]
msridhar: I guess `resolveAndLockAll` would have worked if I had defined it within an `allprojects` block.  But (as you noted) `allDeps` is simpler, and I think it's good enough for our use case.

msridhar:
BTW, I tested this change locally by commenting out the constraints introduced in #899.  With the lockfiles present, the code compiled and tests passed without those constraints, indicating the lockfiles are working as intended.

liblit (approved):

PR #914, opened 2021-07-20
Dev env setup

VQTran123:
Created a file called GettingStarted.md that documents the developer environment setup process for WALACodeNetDev.


msridhar:
Going to close this out as I expect it's either outdated or belongs on another repo

PR #923, opened 2021-08-06
Add test for nestmates feature for JDK 11+

jerryxu158:


msridhar:
Couple high-level comments.

First, CI is failing because we need to exclude `NestmatesTest` from running on pre-11 JDKs in the `build.gradle`, like we did  for the private interface method tests.

Second, for testing this, I don't think we need to check for the presence of the same call graph edges we checked for private interface methods.  The key thing about nestmates is that inner classes can access private fields of enclosing classes without going through an accessor method.  So, besides just testing that the `Outer` and `Outer$Inner` classes get loaded by WALA without error, as a sanity check we could test that inside the `triple()` method there are no method calls at all (before JDK 11 there would be a call to an accessor method).  To do this you could call `getSuccNodes()` on the `CallGraph`, passing the node for `triple()`, and check that `hasNext()` is `false` on the returned `Iterator`.

com.ibm.wala.core/src/test/java/com/ibm/wala/core/tests/nestmates/NestmatesTest.java
```diff
@@ -0,0 +1,82 @@
...
+    MethodReference mm = MethodReference.findOrCreate(tm, "main", "([Ljava/lang/String;)V");
+    Assert.assertTrue("expect main node", cg.getNodes(mm).iterator().hasNext());
+    CGNode mnode = cg.getNodes(mm).iterator().next();
+
+    // should be from main to Triple()
+    TypeReference t1s =
+        TypeReference.findOrCreate(ClassLoaderReference.Application, "Lnestmates/Outer$Inner");
+    MethodReference t1m = MethodReference.findOrCreate(t1s, "triple", "()I");
+    Assert.assertTrue("expect Outer.Inner.triple node", cg.getNodes(t1m).iterator().hasNext());
+    CGNode t2node = cg.getNodes(t1m).iterator().next();
```
msridhar, line 59:
```suggestion
    CGNode t1node = cg.getNodes(t1m).iterator().next();
```
Rename for consistency

com.ibm.wala.core/src/test/java/com/ibm/wala/core/tests/nestmates/NestmatesTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+    // Check call from main to Triple()
+    Assert.assertTrue(
+        "should have call site from main to TestNestmates.triple()",
+        cg.getPossibleSites(mnode, t2node).hasNext());
+
+    // check that Iclass.getAllMethods() returns Triple node
+    TypeReference test1Type =
+        TypeReference.findOrCreate(ClassLoaderReference.Application, "Lnestmates/Outer$Inner");
+    IClass test1Class = cha.lookupClass(test1Type);
```
msridhar, line 69: I think we can get rid of this code; the check is not super useful

com.ibm.wala.core/src/test/java/com/ibm/wala/core/tests/nestmates/NestmatesTest.java
```diff
@@ -0,0 +1,82 @@
...
+    Assert.assertTrue(
+        "there should not be a call from triple() to an accessor method",
+        cg.getSuccNodes(t2node).hasNext() == false);
+    Collection<? extends IMethod> allMethods = test1Class.getAllMethods();
+    IMethod intMethod = test1Class.getMethod(t1m.getSelector());
+    System.out.println(intMethod);
+    Assert.assertTrue(
+        "Expecting an int method to show up in IClass.allMethods()",
+        allMethods.contains(intMethod));
+  }
```
msridhar, line 81: I think we can get rid of this code; the check is not super useful

com.ibm.wala.core/src/test/java/com/ibm/wala/core/tests/nestmates/NestmatesTest.java
```diff
@@ -0,0 +1,82 @@
...
+    Assert.assertTrue(
+        "should have call site from main to TestNestmates.triple()",
+        cg.getPossibleSites(mnode, t2node).hasNext());
+
+    // check that Iclass.getAllMethods() returns Triple node
+    TypeReference test1Type =
+        TypeReference.findOrCreate(ClassLoaderReference.Application, "Lnestmates/Outer$Inner");
+    IClass test1Class = cha.lookupClass(test1Type);
+
+    // check that there is no accessor method after triple()
```
msridhar, line 71:
```suggestion
    // check that triple() does not call an accessor method
```

msridhar (changes requested):
Looks good!  I have a couple comments

msridhar (approved):
LGTM!  Will land once CI passes

PR #932, opened 2021-08-30
Update to Gradle 7.2 and enable stricter locking

msridhar:
Now, any resolved dependence that does not match the lock files will cause the build to fail.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.2-bin.zip
```
msridhar, line 3: @liblit by default, running `./gradlew wrapper` to update pulls in the `bin` distribution, not `all`.  I don't quite see why we would need `all` here, but if you prefer it let me know.

liblit: I prefer `all`. This larger download lets IntelliJ IDEA do a better job offering Gradle API/DSL documentation, which is helpful to me when hacking on the build system.

msridhar: Switched in ac95c76b6a6f667c4bd9f889286a31e7bffb9bdd

liblit (approved):

PR #942, opened 2021-09-05
Save task providers for reuse a few lines later

liblit:
Previously we were looking up some task providers by name.  But we created those task providers just slightly earlier in this same
`build.gradle` script.  We may as well retain references to the providers we created instead of discarding them and looking them up again later.

### DIFF
--- a/com.ibm.wala.cast.java.ecj/build.gradle
+++ b/com.ibm.wala.cast.java.ecj/build.gradle
@@ -28,7 +28,7 @@ dependencies {
 
 application.mainClass.set('com.ibm.wala.cast.java.ecj.util.SourceDirCallGraph')
 
-tasks.named('run') {
+final run = tasks.named('run') {
 	// this is for testing purposes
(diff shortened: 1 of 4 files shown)


com.ibm.wala.cast.java.ecj/build.gradle
```diff
@@ -52,7 +52,7 @@ tasks.named('test') {
 	// temporarily turn off some tests on JDK 9+
 //	if (JavaVersion.current() <= JavaVersion.VERSION_1_8) {
 		// ensure the command-line driver for running ECJ works
```
msridhar, line 54: Maybe we can remove these commented lines while we are editing here?  Also the ones starting at line 56

msridhar (approved):
Approved with one optional suggestion

com.ibm.wala.cast.java.ecj/build.gradle
liblit: If we want to remove these lines, then we should do so separately, since that's unrelated to the narrowly focused purpose of this pull request.

PR #944, opened 2021-09-05
Don't compile JSDT projects with Gradle

msridhar:
The JSDT Gradle projects pull artifacts from the Alfresco Maven repository, which now does not work due to an expired certificate.  Rather than trying to fix this in Gradle, this PR disables Gradle on the modules.  The modules are still built and tested via the "Maven Eclipse tests" CI job, which pulls the relevant artifacts from a p2 repository (via Tycho) so it is unaffected.


liblit (changes requested):
Please revise your commit message to explain *why* you are making this change.

msridhar:
> Please revise your commit message to explain _why_ you are making this change.

When I merge I will squash the commits and use the PR description above as the commit message.  Sound good?

liblit:
> Sound good?

Yes. Thank you!

liblit (approved):

PR #939, opened 2021-09-05
Remove JUnit as an implementation dependence of cast.js

msridhar:
JUnit should only be required for tests.

Issue #937, opened 2021-09-03
Ecliplse import issue


bowenshen:
Hello Everyone

I successfully imported WALA several months before. But I met a issue now when I redo my steps.
I followed the instructions on https://github.com/wala/WALA/blob/master/README-Gradle.md#eclipse
I got pure WALA from github and installed required plugin of Eclipse.
Since I have multiple java version (sdkman management), gradle is setting to Java 11.0.2-open by default.
After importing, I tried to build project. There are 7 errors suck like
"Abstract methods do not specify a body"
"Illegal modifier for the interface method GetT; only public, abstract, default, static and strictfp are permitted"
"The type ReturnArray must implement the inherited abstract method PrivateInterface.GetT(Object)"
"build.properties does not exist" errors showing in	com.ibm.wala.core, com.ibm.wala.shrike, com.ibm.wala.util

I believe it might be related with Java version. So I switched back to import option and chose java 8.0.232-open instead. Several errors disappeared but 3 remained.
"build.properties does not exist" errors showing in	com.ibm.wala.core, com.ibm.wala.shrike, com.ibm.wala.util

Could you please help to resolve the issue?

The reason I need to build WALA myself instead of using WALA-start because I want to adding something in the source code for some specific purposes.

Thank you in advance

msridhar:
@bowenshen could you share the version of Eclipse you are using?

@liblit any ideas here?  I think you might have the most Eclipse setup expertise right now

bowenshen:
Thank you for reply. I'm using Eclipse IDE for Java Developers 2019-12 (4.14.0)

liblit:
I have not used Eclipse for WALA development for quite a few years. I don't think any of the WALA maintainers use Eclipse, so I'm not surprised that some regressions have slipped in. That being said, I think that I can speak on behalf of my fellow maintainers when I say that we would *like* WALA + Eclipse to be a working combination. So thank you for bringing these import problems to our attention, @bowenshen.

I've imported WALA into Eclipse 2021-06 (4.20.0). That Eclipse installation has Buildship 3.1.5.v20210113-0929 installed. My Eclipse is configured to know about Java 1.8, 11, and 16 SDKs, with Java 16 being the default. After import, the Problems view reports two warnings and two infos, neither of which I'm going to worry about for now. The Problems view also reports the following three errors:

| Description | Location |
|---|---|
[...]

PR #952, opened 2021-09-06
Some updates to Gradle lock files

msridhar:
These were generated by running `./gradlew allDeps --write-locks`.  The changes include:

1. A new lock file for `com.ibm.wala.core.java11`
2. Further updates for JUnit and Hamcrest dependencies that were missed in #939 
3. Various updates to Eclipse-related dependencies that I don't fully understand, but might be due to a changed version of the `com.diffplug.eclipse.mavencentral`

It's a bit puzzling to me that the CI jobs were all passing despite the issues in 2 and 3.  My best guess is that the build will only fail for _dynamic_ dependencies of configurations that get resolved, where the resolved dependence doesn't match the lock file.  So for 2 and 3, maybe the relevant configurations are never resolved on CI, or maybe these are changes in non-dynamic dependencies.

FWIW, this change doesn't seem to affect dependencies showing up in https://github.com/wala/WALA/issues/937#issuecomment-913363643; so that must be a separate issue.


liblit (approved):
Pipelined approval: OK to merge *if* the automated checks complete successfully.

liblit:
FYI, I have a fix in progress for https://github.com/wala/WALA/issues/937#issuecomment-913363643, which I also don't fully understand. But I'm going to wait until after #952 is merged, since #952 might affect the particular ways in which the problems of https://github.com/wala/WALA/issues/937#issuecomment-913363643 manifest and should be worked around.

PR #953, opened 2021-09-06
Give `autoLintGradle` task an empty output file

liblit:
`autoLintGradle` creates no output files, which causes Gradle to treat it as always out-of-date.  By creating a simple, empty stamp file to record that this task has run and succeeded, we allow Gradle to avoid rerunning this task unnecessarily.  This task will still be rerun when needed, though, such as when any of the `**/build.gradle` files change.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -282,6 +282,18 @@ final verifyGoogleJavaFormat = tasks.named('verifyGoogleJavaFormat') {
 	}
 }
 
+tasks.named('autoLintGradle') {
+	// `autoLintGradle` creates no output files, which causes Gradle to treat it as always
+	// out-of-date.  By creating a simple, empty stamp file to record that this task has run and
+	// succeeded, we allow Gradle to avoid rerunning this task unnecessarily.  This task will still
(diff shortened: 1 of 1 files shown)


msridhar (approved):
LGTM.  Are there docs on what `autoLintGradle` actually checks?  FWIW, I've never seen it fail.

liblit:
> Are there docs on what `autoLintGradle` actually checks?

[Yes.](https://github.com/nebula-plugins/gradle-lint-plugin/wiki)

> FWIW, I've never seen it fail.

It used to fail much more often. I've gotten better at Gradle.

msridhar:
> > FWIW, I've never seen it fail.
> 
> It used to fail much more often. I've gotten better at Gradle.

Haha I'm sure 🙂 But it looks like it might be disabled right now, or at least not checking any rules?

https://github.com/wala/WALA/blob/e29e2e97e679323bcfa7c48f7374f3f9baa0e3cf/build.gradle#L217-L231

liblit:
> But it looks like it might be disabled right now, or at least not checking any rules?

Thanks for pointing this out. I'll look into it separately.

Issue #916, opened 2021-07-21
Flaky test failures on Ubuntu JDK 11


msridhar:
On CI, we occasionally get a test failure in `com.ibm.wala.core` only on JDK 11, with a stack like the following:
```
com.ibm.wala.core.tests.callGraph.CallGraphTest > testJava_cup FAILED
341
    java.lang.AssertionError
342
        at com.ibm.wala.util.intset.BitVectorIntSet$1.hasNext(BitVectorIntSet.java:228)
343
        at com.ibm.wala.ipa.callgraph.propagation.rta.DelegatingExplicitCallGraph$DelegatingEdgeManager.getPredNodeNumbers(DelegatingExplicitCallGraph.java:243)
344
        at com.ibm.wala.ipa.callgraph.impl.ExplicitCallGraph$ExplicitEdgeManager.getPredNodes(ExplicitCallGraph.java:374)
345
        at com.ibm.wala.ipa.callgraph.impl.ExplicitCallGraph$ExplicitEdgeManager.getPredNodes(ExplicitCallGraph.java:343)
346
        at com.ibm.wala.util.graph.AbstractGraph.getPredNodes(AbstractGraph.java:66)
347
        at com.ibm.wala.util.graph.GraphIntegrity.checkEdges(GraphIntegrity.java:90)
348
        at com.ibm.wala.util.graph.GraphIntegrity.check(GraphIntegrity.java:41)
349
        at com.ibm.wala.core.tests.callGraph.CallGraphTest.doCallGraphs(CallGraphTest.java:408)
350
        at 
```
I do not know how this could be failing only on JDK 11.  I will try updating the JDK used in CI to see if it goes away due to a fixed JVM bug.  If it doesn't, I will have to investigate further.

msridhar:
Update: #917 did not fix this problem.  I also cannot reproduce on Mac.  It may be a combination of a WALA bug and some non-determinism that is only getting exposed on Linux.  It's a tough one to debug.

liblit:
Should we consider disabling this test when running on Java 11 until the problem can be fixed? This non-deterministic failure creates a speed bump for other changes: we need to rerun *all* GitHub CI jobs until we get lucky for this one flaky test.

msridhar:
The problem is it's not always the same test that fails.  In the example above, it was `test_Javacup`, but in the CI job that failed for #954 it was `testOcamlHash`.  We could try disabling both of these on Java 11 for now, I guess.

[...]

PR #954, opened 2021-09-06
Avoid conditional dependencies

liblit:
Gradle dependency version locksets work best when dependencies are fixed
across all builds.


msridhar:
The CI failure is due to a flaky test (#916).  A restart should hopefully fix.

I'm really puzzled by this issue.  I'd like to think a bit more about what could be the root cause before approving.

msridhar:
It seems the `-PexcludeSlowTests` flag is related to / causing this issue, at least that's what I see locally.  @liblit can you confirm that if you remove that option you don't get the same type of failure anymore?  I wonder if that property is causing additional dependencies to be pulled in.  If so, maybe it'd be better to pull in the deps unconditionally somehow?  If that's too tricky, I'm ok with this current fix, with some extra comments indicating why we need it.

liblit:
> It seems the `-PexcludeSlowTests` flag is related to / causing this issue, at least that's what I see locally.

Woah. 🤯 How did you manage to figure that out?

> @liblit can you confirm that if you remove that option you don't get the same type of failure anymore?

Confirmed.

> I wonder if that property is causing additional dependencies to be pulled in.

It must have something to do with this code that only executes when `excludeSlowTests` is set:

https://github.com/wala/WALA/blob/35b8e41f230abb6062a9a66c2f6c7ae4611036d2/build.gradle#L113-L115

> If so, maybe it'd be better to pull in the deps unconditionally somehow?

Indeed, I think that's exactly the right solution, rather than the workaround-without-true-understanding that I originally proposed here. I'll revise the code and this pull request accordingly. Good job, @msridhar!

msridhar (approved):
👍

liblit:
@msridhar wrote:
> The CI failure is due to a flaky test (#916). A restart should hopefully fix.

Thanks for that hint, too. I would have spent a nontrivial amount of time rediscovering that on my own.

PR #955, opened 2021-09-06
Update Java version selection strategy; resolve all Eclipse diagnostics

### LINKED ISSUE #937: Ecliplse import issue

liblit:
### Resolve Eclipse errors following a pristine import

IntelliJ IDEA can configure a different Java version for each `SourceSet` within each Gradle subproject.  Eclipse, unfortunately, can only configure one Java version per subproject.  Therefore, Java 11+ material that was previously in `com.ibm.wala.core` needs to pulled out into some other subproject.  The existing `com.ibm.wala.core.java11` seems like a sensible place to put that stuff.

Also change how we tell Gradle what version of Java to use where.  The old `java.sourceCompatibility` and `java.targetCompatibility` settings are still supported, but the preferred way to do this is now via the `java.toolchains` APIs.  I am pleased to report that both IntelliJ IDEA (with the Gradle plugin) and Eclipse (with the Buildship plugins) interpret the Java toolchain settings correctly.  In either IDE, you end up with nearly all subprojeccts configured to use Java 1.8, but with `com.ibm.wala.core.java11` configured to use Java 11.  Perfect!

If you prefer to try using some other Java version instead of 1.8, then set the `javaToolchainLanguageVersion` project property to the desired version.  For example, `./gradlew -PjavaToolchainLanguageVersion=11` will try to build and run *everything* using a Java 11 toolchain.  There are definitely no guarantees that this will succeed, but that's at least how you'd *attempt* it.  We know that `:com.ibm.wala.core:test` passes this way, but beyond that, you're on your own.

Note that [Gradle auto-provisioning of Java toolchains](https://docs.gradle.org/current/userguide/toolchains.html#sec:provisioning) is enabled by default.  If Java 11 is needed, but Gradle cannot [find an existing Java 11 JDK](https://docs.gradle.org/current/userguide/toolchains.html#sec:auto_detection), then Gradle will download a suitable JDK automatically.  This can be disabled, of course, but it's enabled by default.

Fixes #937.

### Remove an unused private method

Removing this unused code resolves one Eclipse warning.

### Suppress an Eclipse warning about dead code

The dead code here would be alive if certain `final static` constants were changed from `false` to `true`.  We presumably want to retain the option to change those flags, which would thereby bring this "dead" code back to life.

### Remove or disable unnecessary warning suppressions

These changes resolve two Eclipse informational diagnostics.


build.gradle
```diff
@@ -58,9 +58,9 @@ subprojects { subproject ->
 	apply plugin: 'maven-publish'
 	apply plugin: 'signing'
 
-	java.sourceCompatibility = JavaVersion.VERSION_1_8
-	java.targetCompatibility = JavaVersion.VERSION_1_8
-	
+	java.toolchain.languageVersion.set(
```
msridhar, line 61: I'm not 100% sure this is what we want.  I think for the foreseeable future, when we cut WALA releases, the target compatibility will be 1.8.  So, even when _running_ on a Java 11+ VM, we may still want the source and target compatibility flags for `javac` set to 1.8, to mimic running an actual WALA release on the more modern VM.  What do you think?

msridhar (commented):
Mostly looks great!  I just have one comment / question.

build.gradle
liblit: > even when _running_ on a Java 11+ VM, we may still want the source and target compatibility flags for `javac` set to 1.8, to mimic running an actual WALA release on the more modern VM. What do you think?

Bleh. But OK, I see what you mean. I've updated the code in this pull request to revert to the older style of setting `java.sourceCompatibility` and `java.targetCompatibility` to `JavaVersion.VERSION_1_8` in all subprojects except for `com.ibm.wala.java11`. In that last subproject, we need to set `sourceCompatibility` to `VERSION_11`, since we genuinely do have source code here that uses post-Java-1.8 syntax. Once we've done that, Eclipse import fails unless we also set `targetCompatibility` to `VERSION_11` in this subproject as well.

Now I'm having second thoughts, though. If your concern is what happens "when we cut WALA releases", then wouldn't it be sufficient to always cut WALA releases using a Java 1.8 toolchain? That would have been the default in my `java.toolchain`-using code: the code you reviewed when making this comment. If we moved that default to some later version in the future, we could have used `./gradlew -PjavaToolchainLanguageVersion=1.8` to explicitly request a Java 1.8 toolchain when building releases. If that succeeded at all, then presumably it would produce Java-1.8-compatible artifacts. Shouldn't that be everything we need, even when using the new Java toolchain API?

com.ibm.wala.core.java11/build.gradle
```diff
@@ -4,8 +4,55 @@ plugins {
 
 eclipse.project.natures 'org.eclipse.pde.PluginNature'
 
+java.sourceCompatibility = null
+java.targetCompatibility = null
```
msridhar, line 8: It's a bit gross to have to null these out.  Maybe we can move this language version configuration to the top-level `build.gradle` and do it conditionally based on the sub-project name?  Or will that mess up Eclipse / Buildship?

liblit: The only package that should have a different configuration for these settings is `com.ibm.wala.core.java11`. Is hard-coding that sub-project name in the top-level `build.gradle` less gross that reassigning these settings to `NULL` here?

msridhar: Fair enough.  We do some conditional stuff on project name [here](https://github.com/wala/WALA/blob/58ca8a9640f1b0e07ecd9e2cf775f8fba68201b2/build.gradle#L46-L53), but that's a bit different I guess.  It's your call on how best to do this, but if you decide to go with the `null` assignments maybe add a comment as to why it's needed.

msridhar (approved):

com.ibm.wala.core.java11/build.gradle
liblit: > if  you decide to go with the `null` assignments maybe add a comment as to why it's needed.

Done!

PR #956, opened 2021-09-06
Resolve linter-reported dependency problems

liblit:
If enabled, `lintGradle` detects several dependency-related problems: duplicate classes provided by multiple dependencies, declared-but-unused dependencies, used-but-undeclared dependencies, etc.  These fixes address all of those problems.

Unfortunately, we don't actually want to leave the Gradle linter plugin enabled.  If enabled, it produces warnings about APIs that are deprecated in Gradle 7 and that will be removed in Gradle 8.  The deprecation warning is non-specific by default, which means that it could easily mask any new deprecation problems that we might introduce into our build code.  By leaving the Gradle linter disabled in most
subprojects, we retain the potential to see a *new* deprecation warning where previously there was none, and for that to be something that we truly should act upon.


liblit:
FYI, I'm going to see if I can leave the linter plugin enabled, but just not have it run `autoLintGradle` for every build. That would satisfy my concerns about deprecation warnings, while still leaving this tool configured and available for anyone to run manually.

liblit:
> FYI, I'm going to see if I can leave the linter plugin enabled, but just not have it run `autoLintGradle` for every build.

Success.

build.gradle
```diff
@@ -88,6 +88,11 @@ subprojects { subproject ->
     }
 
 	configurations {
+		all {
+			resolutionStrategy.dependencySubstitution {
+				substitute module('org.hamcrest:hamcrest-core') using module('org.hamcrest:hamcrest:2.2')
+			}
+		}
```
msridhar, line 95: Can we add a comment on why we are doing this?

build.gradle
```diff
@@ -211,22 +216,31 @@ tasks.register('aggregatedJavadocs', Javadoc) { aggregated ->
 //  linters for various specific languages or file formats
 //
 
-// Gradle build scripts
-// disabled until <https://github.com/gradle/gradle/issues/6246> is fixed
-/*
+// We use no deprecated Gradle APIs ourselves, and we'd like to keep it that way.  Unfortunately,
+// the Gradle linter tasks produce a deprecation warning under Gradle 7.2.  The warning message
+// is not detailed by default, and could easily mask any new deprecation problems that we might
+// introduce ourselves.  So we don't run the linter by default for now.
```
msridhar, line 222: Should this comment go next to the `gradleLint.alwaysRun = false` line?  I think that's what it pertains to

com.ibm.wala.ide.tests/build.gradle
```diff
@@ -24,16 +24,19 @@ eclipseMavenCentral {
...
-	}
-
 }
+
+configurations.all {
+	resolutionStrategy.dependencySubstitution {
+		substitute module('org.eclipse.platform:org.eclipse.osgi.services') using module('org.eclipse.platform:org.eclipse.osgi:3.15.100')
+		substitute module('xml-apis:xml-apis-ext') using module('org.eclipse.birt.runtime:org.w3c.css.sac:1.3.1.v200903091627')
+	}
+}
```
msridhar, line 42: Could we also add a comment on this?  And maybe a trailing newline?

msridhar (changes requested):
Looks great!  Couple of minor comments

build.gradle
liblit: Added an explanation using the `because` construct that is optionally allowed here.

build.gradle
liblit: Good call. Fixed.

com.ibm.wala.ide.tests/build.gradle
liblit: Added an explanation using the `because` construct that is optionally allowed here.

Added a trailing newline.

build.gradle
```diff
@@ -88,6 +88,11 @@ subprojects { subproject ->
     }
 
 	configurations {
+		all {
+			resolutionStrategy.dependencySubstitution {
+				substitute module('org.hamcrest:hamcrest-core') using module('org.hamcrest:hamcrest:2.2') because 'junit depends on hancrest-core, but a ll hamcrest-core classes have been incorporated into hamcrest'
```
msridhar, line 93:
```suggestion
				substitute module('org.hamcrest:hamcrest-core') using module('org.hamcrest:hamcrest:2.2') because 'junit depends on hamcrest-core, but all hamcrest-core classes have been incorporated into hamcrest'
```

build.gradle
```diff
@@ -211,22 +216,32 @@ tasks.register('aggregatedJavadocs', Javadoc) { aggregated ->
 //  linters for various specific languages or file formats
 //
 
-// Gradle build scripts
-// disabled until <https://github.com/gradle/gradle/issues/6246> is fixed
-/*
 allprojects {
 	apply plugin: 'nebula.lint'
+
+    // We use no deprecated Gradle APIs ourselves, and we'd like to keep it that way.
```
msridhar, line 222: Indentation?

msridhar (approved):
Approved assuming minor nits below are addressed

build.gradle
liblit: Oops, inconsistent use of tabs (elsewhere) versus spaces (here). Fixed. Thanks!

PR #961, opened 2021-09-10
Don't run CI jobs on pushes to dependabot branches

msridhar:
Dependabot pushes to branches in the `wala/WALA` repo for its changes, which means that right now, two sets of (redundant) jobs run on these PRs: one set for the pull request, and one for the push to the branch.  This change eliminates the jobs on the branch push, which will save some CI resources.  Not a huge deal in the grand scheme of things, but I hate wasting compute 🙂

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,7 +1,9 @@
 name: Continuous integration
 on:
-  - pull_request
-  - push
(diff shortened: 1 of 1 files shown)


msridhar:
FWIW I tested out this change on a dependabot PR and it seemed to work as intended

liblit (approved):

PR #975, opened 2021-09-15
Revive IFDSExplorerExample

### LINKED ISSUE #921: wala-ide.jar
There is not wala-ide.jar in maven repository.
How should I get it ?
Can you upload it to the maven repository?

msridhar:
Fixes #921 

Previously we did not test that the SWT-related code in `com.ibm.wala.ide.tests` actually ran correctly, and we ended up with some classpath issues not detected by just building the code (e.g., a dependence that required Java 11).  This PR fixes these issues, and adds a unit test so we don't regress in the future.

Also, we add a Gradle task to run `IFDSExplorerExample` to make it easier for others to try out in the future.


.github/workflows/continuous-integration.yml
```diff
@@ -47,11 +47,11 @@ jobs:
           java-version: ${{ matrix.java }}
           distribution: 'adopt'
       - name: Build and test using Gradle and Java 8
-        run: ./gradlew --no-build-cache --no-daemon javadoc build lintGradle
+        run: xvfb-run --auto-servernum ./gradlew --no-build-cache --no-daemon javadoc build lintGradle
```
msridhar, line 50: `xvfb-run` is required for the SWT test (which has a GUI) to actually run

build.gradle
```diff
@@ -82,9 +82,37 @@ subprojects { subproject ->
     		}
     	}
         implementation('org.eclipse.platform:org.eclipse.core.commands') {
-                version { strictly '(,3.9.800)' }
-                because 'Versions 3.10.0 and later of this artifact require Java 11.'
+			version { strictly '(,3.9.800)' }
```
msridhar, line 85: This just makes the indentation consistent

build.gradle
```diff
@@ -82,9 +82,37 @@ subprojects { subproject ->
...
+			because 'Versions 3.10.0 and later of this artifact require Java 11.'
         }
+	    implementation('org.eclipse.platform:org.eclipse.e4.ui.workbench') {
+			version { strictly '(,1.11.0]' }
+			because 'Later versions of this artifact require Java 11.'
+	    }
+	    implementation('org.eclipse.platform:org.eclipse.equinox.registry') {
+			version { strictly '(,3.8.600]' }
+			because 'Dependencies from later versions introduce inconsistencies with constraints above.'
```
msridhar, line 94: I did my best to document these.  The real issue is that the Gradle constraint solver isn't "smart" enough, and greedily chooses the most recent available version of a dep consistent with the given dynamic range.  E.g., if you have dep A fixed at version X, A depends on B with some range, and some transitive dep C of B requires dep A at version X+1, constraint resolution fails.  Ideally Gradle would just choose an earlier version of C, which would be consistent with all constraints and avoid the failure, but it doesn't.  So we need to add a bunch of version restrictions.

com.ibm.wala.ide.tests/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -84,8 +84,8 @@ org.eclipse.jdt.core.compiler.problem.redundantNullAnnotation=error
 org.eclipse.jdt.core.compiler.problem.redundantNullCheck=error
 org.eclipse.jdt.core.compiler.problem.redundantSpecificationOfTypeArguments=error
 org.eclipse.jdt.core.compiler.problem.redundantSuperinterface=error
-org.eclipse.jdt.core.compiler.problem.reportMethodCanBePotentiallyStatic=error
-org.eclipse.jdt.core.compiler.problem.reportMethodCanBeStatic=error
+org.eclipse.jdt.core.compiler.problem.reportMethodCanBePotentiallyStatic=ignore
```
msridhar, line 87: Without this change ECJ complains that a JUnit test method should be static

com.ibm.wala.ide.tests/src/test/resources/wala.testdata.txt
```diff
@@ -0,0 +1,3 @@
+Primordial,Java,stdlib,none
```
msridhar, line 1: I wish we didn't have to duplicate this file, but didn't immediately see a better solution

com.ibm.wala.ide.tests/.settings/org.eclipse.jdt.core.prefs
liblit: If we want to receive these warnings for *most* of the code in this subproject, then it would be better to leave this Eclipse warning enabled and just suppress it for specific methods where we disagree with Eclipse's recommendation. Is that the case here? Or is this diagnostic wrong so often that it's impractical to suppress each wrong instance?

com.ibm.wala.ide.tests/src/test/resources/wala.testdata.txt
liblit: The original is `com.ibm.wala.core/src/test/resources/wala.testdata.txt`, right? We should be able to have Gradle copy that original file for us as part of the `:com.ibm.wala.ide.tests:processTestResources` task.

Try removing this manually copied file and instead adding something like the following to `com.ibm.wala.ide.tests/build.gradle`:

```gradle
tasks.named('processTestResources', Copy) {
	from project(':com.ibm.wala.core').sourceSets.test.resources
	include 'wala.testdata.txt'
}
```

liblit (changes requested):
Thank you for the extra explanatory comments!

com.ibm.wala.ide.tests/.settings/org.eclipse.jdt.core.prefs
msridhar: I think it's a lint that should be disabled on test code if it's not aware of JUnit, as `@Test` methods should _not_ be static (AFAIK).  Plus the motivation for this lint, which is mostly performance I think, is moot for test code.  Unfortunately, we don't have a good way to enable / disable ECJ checks at the source set granularity, so I'm in favor of just disabling the check for the module (it is disabled in at least some other WALA modules).

liblit: OK.

com.ibm.wala.ide.tests/src/test/resources/wala.testdata.txt
msridhar: It worked! 0d3cbea

liblit: W00t!

liblit (approved):

PR #976, opened 2021-09-16
Add CI step to cancel previous workflows for same branch / PR

msridhar:
Sometimes I end up pushing multiple times to a PR in a short time, and by default GitHub will run CI jobs for each push.  But only the most recent CI job "counts" in terms of checking the changes.  This changes adds a step to cancel running but outdated workflows; see https://github.com/styfle/cancel-workflow-action/blob/main/README.md


liblit (approved):
Nice!

PR #977, opened 2021-09-16
Remove useless empty @throws tags

liblit:
A Javadoc `@throws FooException` with no explanation as to the circumstances under which the exception arises is useless.  Presumably these were added at some point to pacify a linter's requirement that all thrown exceptions be documented.  An empty `@throws FooException` tag is *not* real documentation, and in fact, these empty tags are producing warning diagnostics of their own when building Javadoc.


msridhar (approved):
👍

PR #980, opened 2021-09-16
Disable snapshots job on CI

msridhar:
It seems to be failing at the moment.  Also disable the change from #976 as it doesn't seem to be working as intended.  I will debug later.


liblit (approved):
Thanks, @msridhar! Hopefully these are all just temporary measures.

PR #983, opened 2021-09-17
Fix hundreds of inactive or broken `@see` references

liblit:
Previously this code had 637 methods preceded by non-Javadoc comments of the form:

```java
/*
 * @see ...
 */
```

Replacing the first line of each of these with `/**` turns it into a Javadoc comment containing only a `@see`.  Sure, that's not *much* of a Javadoc comment, but it's better than nothing.  At least this gives the `@see` cross-references their intended semantics.  For example, IDEs can make these active for code navigation, the Javadoc compiler can validate that the symbols they refer to actually exist, etc.

There's no good reason to add explicit `@see` references to the method that owns a Javadoc comment, the super method of the method that owns a Javadoc comment, or the class that contains a Javadoc comment.  So remove these redundant `@see` references.

Correct many other broken `@see` references to symbols that do not exist.  I've tried my best to infer what the intended reference target is for each.

Remove some broken `@see` references to missing symbols.  In these cases, I've been unable to identify suitable replacements.  If there *is* a suitable replacement for any of these that I've overlooked, please let me know!  Keep in mind, however, that we don't want to add useless references as described two paragraphs earlier.


com.ibm.wala.core/src/main/java/com/ibm/wala/analysis/pointers/BasicHeapGraph.java
```diff
@@ -376,97 +376,73 @@ public int compare(Object arg1, Object arg2) {
     }
   }
 
-  /*
-   * @see com.ibm.wala.util.graph.NumberedNodeManager#getNumber(com.ibm.wala.util.graph.Node)
-   */
+  /** @see com.ibm.wala.util.graph.NumberedNodeManager#getNumber(Object) */
```
msridhar, line 379: Hrm, I'm not sure these `@see` links to the superclass method add much value.  Are these something we should delete instead, under the assumption that a developer will go to go to superclass / interface methods for "inherited" documentation?

com.ibm.wala.core/src/main/java/com/ibm/wala/cfg/AbstractCFG.java
```diff
@@ -492,25 +492,19 @@ public void addExceptionalEdge(T src, T dst) {
     }
   }
 
-  /*
-   * @see com.ibm.wala.util.graph.Graph#removeNode(com.ibm.wala.util.graph.Node)
-   */
+  /** @see com.ibm.wala.util.graph.Graph#removeNode(Object) */
```
msridhar, line 495: I think `Graph` still has a `removeNodeAndEdges` method? https://github.com/wala/WALA/blob/156d3266a974baabd22314f959604b10cc8eb339/com.ibm.wala.util/src/main/java/com/ibm/wala/util/graph/Graph.java#L29

msridhar (commented):
Thanks for this great cleanup!  Added a couple of comments, but it's hard to review all these changes.

Taking a step back, I don't think I've seen a single `@see` doc in this change that seems like it adds much value.  I wonder, should we just delete all of them?  Most just seem to link out to the superclass / interface method being overridden / implemented, but IDEs have a built-in way to navigate to such methods and it's a natural thing for devs to do when reading code.  And even the Javadoc will have links to superclasses / interfaces by default.  So I don't see the value in also having an `@see` link.  @liblit thoughts?

com.ibm.wala.core/src/main/java/com/ibm/wala/analysis/pointers/BasicHeapGraph.java
liblit: This is not a `@see` link to the superclass method. `BasicHeapGraph` does not implement `NumberedNodeManager`.

com.ibm.wala.core/src/main/java/com/ibm/wala/cfg/AbstractCFG.java
liblit: Yes, `Graph::removeNodeAndEdges` still exists, and this method implements that interface method. What of it? The original `@see` tag was not a reference to `Graph::removeNodeAndEdges`. It was a reference to `Graph::removeNode`, except with the wrong parameter type. I have corrected that parameter type, as this seems most consistent with the original intent of the existing `@see` tag.

liblit:
@msridhar wrote:
> Taking a step back, I don't think I've seen a single @see doc in this change that seems like it adds much value. I wonder, should we just delete all of them?

That seems rather extreme. I am not convinced that *every* such link is useless, and I have no handy mechanism to distinguish the useless from the useful ones. Lacking such a mechanism, I prefer to err on the side of caution: preserve information that *somebody* apparently considered useful at some point in time.

> Most just seem to link out to the superclass / interface method being overridden / implemented, but IDEs have a built-in way to navigate to such methods and it's a natural thing for devs to do when reading code. And even the Javadoc will have links to superclasses / interfaces by default. So I don't see the value in also having an @see link.

I agree that there is no value in a `@see` link to the superclass or interface method that is being overridden or implemented. I have used IntelliJ IDEA's "Unnecessary Javadoc link" inspection to identify and remove many of these. Per that inspection's description:

> Reports the Javadoc `@see`, `{@link}`, and `{@linkplain}` tags that refer to the method owning the comment, the super method of the method owning the comment, or the class containing the comment.
>
> Such links are unnecessary and can be safely removed with this inspection's quick-fix. The quick-fix will remove the entire Javadoc comment if it refers to its own content.

I am not sure that this inspection is as thorough as we would want it to be, though. It definitely detects `@see` referring to a method from the immediate superclass or any directly implemented interface. I am not certain that it finds `@see` referring to methods from transitively implemented interfaces, though. If it did, then I would be happy to remove those too.

That being said, I am not convinced that all of the `@see` links are as redundant as they seem. You pointed out one that you thought was redundant, wherein the Javadoc for `BasicHeapGraph::getNumber` has a `@see` link to `NumberedNodeManager::getNumber`. But `BasicHeapGraph` is not an implementation of `NumberedNodeManager`; this is not a trivially redundant link of the sort that you suggest removing. We could discuss whether it is useful or not to point out `NumberedNodeManager::getNumber` as a method of interest when one is looking at `BasicHeapGraph::getNumber`: maybe that `@see` link is informative, maybe it isn't. But it's not trivially redundant. It does not just "link out to the superclass / interface method being overridden / implemented".

msridhar:
> Lacking such a mechanism, I prefer to err on the side of caution: preserve information that _somebody_ apparently considered useful at some point in time.

Ok, fair enough.  My strong suspicion is that these tags were generated by some IDE at some point and they add basically no value (at least I haven't seen a useful one).  But it doesn't hurt much to leave them in so we can err on the side of caution.

msridhar (approved):

PR #986, opened 2021-09-19
Conform to Java package naming conventions

liblit:
Java package names conventionally start with lower-case letters.


com.ibm.wala.core/src/main/java/com/ibm/wala/dataflow/ifds/BackwardsSupergraph.java
```diff
@@ -8,7 +8,7 @@
  * Contributors:
  *     IBM Corporation - initial API and implementation
  */
-package com.ibm.wala.dataflow.IFDS;
+package com.ibm.wala.dataflow.ifds;
```
msridhar, line 11: I'm not 100% sure about this change.  It will "break" any WALA client using the IFDS stuff.  The fix is easy but will take up some time for any WALA user.  Since IFDS is an acronym anyway, I wonder is this worth it?  @liblit thoughts?

liblit: “IFDS” is an acronym, but so are “IBM” and “WALA”. We don’t name our packages starting with `com.IBM.WALA`, so I don’t find this acronym argument convincing.

The case for backward compatibility is stronger, though. I agree that this would be a rude change if the next official release will be 1.5.8. It might be a justifiable change if the next release is 1.6.0, but even then, the benefit seems smaller than the disruption.

I will revise this pull request to leave “IFDS” unchanged in package names.

msridhar (approved):

PR #989, opened 2021-09-19
Cancel outdated jobs using workflow concurrency

msridhar:
This is an improved version of #976 using [the `concurrency` feature](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency) of GH Actions workflows.  This ensures our CI workflow only runs for the latest commit on any branch / PR, and also that only one workflow for each branch is running at any given time.  The latter is useful for the WALA master branch in particular, since before it was possible for workflows to upload Javadoc and snapshot builds out of order if one ran particularly slowly.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -23,6 +23,9 @@ jobs:
             java: 11
       fail-fast: false
     runs-on: ${{ matrix.os }}
+
+
+    
     steps:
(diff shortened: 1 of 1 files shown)


msridhar:
Confirmed via dummy commits that the workflow cancellation works both in my fork and for the pull request

liblit (approved):

Issue #888, opened 2021-05-26
Building Callgraph: RTA crashing when using dex frontend


greenerpilotx:
Hello,

I clone the latest Wala a few days ago to work with Android. However, I have been running into some issues using _DexFileModule_. Using the latest version of Wala, each  time I try to build a call graph using a dex file,  I get this particular exception: 

    Exception in thread "main" java.lang.ClassCastException: [Lcom.ibm.wala.dalvik.dex.instructions.Instruction; cannot be cast to [Lcom.ibm.wala.shrikeBT.IInstruction;
        at com.ibm.wala.classLoader.ShrikeIRFactory.makeIR(ShrikeIRFactory.java:51)
        at com.ibm.wala.ssa.DefaultIRFactory.makeIR(DefaultIRFactory.java:70)
        at com.ibm.wala.ssa.SSACache.findOrCreateIR(SSACache.java:69)
        at com.ibm.wala.ipa.callgraph.AnalysisCache.getIR(AnalysisCache.java:74)
        at com.ibm.wala.ipa.callgraph.propagation.cfa.ContextInsensitiveSSAInterpreter.getIR(ContextInsensitiveSSAInterpreter.java:45)
        at com.ibm.wala.ipa.callgraph.propagation.cfa.DefaultSSAInterpreter.getIR(DefaultSSAInterpreter.java:53)
        at com.ibm.wala.ipa.callgraph.propagation.cfa.DelegatingSSAContextInterpreter.getIR(DelegatingSSAContextInterpreter.java:53)
        at com.ibm.wala.ipa.callgraph.impl.ExplicitCallGraph$ExplicitNode.getIR(ExplicitCallGraph.java:295)
        at com.ibm.wala.ipa.callgraph.propagation.SSAPropagationCallGraphBuilder.getTargetPointerKey(SSAPropagationCallGraphBuilder.java:2287)
        at com.ibm.wala.ipa.callgraph.propagation.SSAPropagationCallGraphBuilder.processCallingConstraints(SSAPropagationCallGraphBuilder.java:1722)
        at com.ibm.wala.ipa.callgraph.propagation.SSAPropagationCallGraphBuilder.processResolvedCall(SSAPropagationCallGraphBuilder.java:1671)
        at com.ibm.wala.ipa.callgraph.propagation.SSAPropagationCallGraphBuilder.access$200(SSAPropagationCallGraphBuilder.java:99)
[...]

PR #994, opened 2021-09-20
AllowContextInsensitiveRTAInterpreter to call the correct CodeScanner method when working with Dex Files

greenerpilotx:
This is a proposed fix for #888. 

Currently, adding dex files to the AnalysisScope and then attempting to build a callgraph results in a java.lang.ClassCastException when using the RTABuilder or the context-sensitive CFAZeroOneBuilder/the deprecated CFAZeroOneContainerBuilder. 

The issue lies in the CodeScanner.java file  and the ContextInsensitiveRTAInterpreter.java file. The ContextInsensitiveRTAInterpreter calls the public method CodeScanner.getNewSites(). However, that method is geared toward either SyntheticMethods or ShrikeBTMethods, not DexIMethods. There is a secondary CodeScanner.getNewSites() that would work perfectly with DexIMethods; the only problem is that it has private access and takes a SSAInstruction array, rather than an IMethod, as an argument. 

The proposed changes are to: 
1) make the relevant getNewSites() method public &
2) have the ContextInsensitiveRTAInterpreter call the getNewSites() method that is compatible with DexIMethods when encountering a method that is not a SyntheticMethod or a ShrikeBTMethod

### DIFF
--- a/com.ibm.wala.core/src/main/java/com/ibm/wala/classLoader/CodeScanner.java
+++ b/com.ibm.wala.core/src/main/java/com/ibm/wala/classLoader/CodeScanner.java
@@ -301,7 +301,7 @@ public class CodeScanner {
    * @param statements list of ssa statements
    * @return List of InvokeInstruction
    */
-  private static List<NewSiteReference> getNewSites(SSAInstruction[] statements) {
+  public static List<NewSiteReference> getNewSites(SSAInstruction[] statements) {
     final List<NewSiteReference> result = new LinkedList<>();
     Visitor v =
         new Visitor() {
--- a/com.ibm.wala.core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/rta/ContextInsensitiveRTAInterpreter.java
+++ b/com.ibm.wala.core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/rta/ContextInsensitiveRTAInterpreter.java
@@ -14,13 +14,16 @@ package com.ibm.wala.ipa.callgraph.propagation.rta;
 import com.ibm.wala.classLoader.CodeScanner;
 import com.ibm.wala.classLoader.IClass;
 import com.ibm.wala.classLoader.NewSiteReference;
+import com.ibm.wala.classLoader.ShrikeBTMethod;
 import com.ibm.wala.ipa.callgraph.CGNode;
 import com.ibm.wala.ipa.callgraph.IAnalysisCacheView;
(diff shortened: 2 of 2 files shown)


msridhar:
Thanks so much for the contribution, @greenerpilotx!  I've read your approach and I believe it will work correctly.  However, I'd like to propose an alternate approach more aligned with the current code.  For Java bytecodes, `CodeScanner` currently obtains the new sites directly from the `ShrikeCTMethod` for efficiency, as constructing an IR has some cost.  It would be great if we could do the same for Dex.  We could do this using the following steps:

1. Add a method `Collection<NewSiteReference> getNewSites()` to the `IBytecodeMethod` interface, around [here](https://github.com/wala/WALA/blob/7b20562027ddf2e7067705895a28c6ade2010372/com.ibm.wala.core/src/main/java/com/ibm/wala/classLoader/IBytecodeMethod.java#L42). 
2. Implement `getNewSites()` in `DexIMethod`, in a manner similar to [this method](https://github.com/wala/WALA/blob/7b20562027ddf2e7067705895a28c6ade2010372/com.ibm.wala.dalvik/src/main/java/com/ibm/wala/dalvik/classLoader/DexIMethod.java#L4612-L4643).
3. Modify `CodeScanner#getNewSites(com.ibm.wala.classLoader.IMethod)` to just invoke `IBytecodeMethod.getNewSites()` in its `else` case.

The advantages of this approach are:
1. We don't force RTA to construct IRs for Dex methods unnecessarily, which is more efficient.
2. We fix any caller who is trying to use `CodeScanner.getNewSites()` with Dex bytecodes, not just RTA.

What do you think?

msridhar:
Also, we should also add a regression test for using RTA on Dex as part of this change.

greenerpilotx:
I think this sounds great! Initially, I was attempting to use DexIMethod directly (like ShrikeBTMethod) but was having trouble because of the separation between com.ibm.wala.core & com.ibm.wala.dalvik. I see now that relying on the IBytecodeMethod interface should have been the approach to take.

msridhar:
@greenerpilotx do you plan to open a fresh PR using the new approach?  I don't think the changes really overlap with those in this PR

greenerpilotx:
@msridhar Yes, definitely, I'll open a new PR.

msridhar:
@greenerpilotx just checking in, do you need any help with putting together a PR on the alternate approach discussed above?

greenerpilotx:
@msridhar Apologies for being late on this. I can put something together in the next day or two.

PR #1017, opened 2021-09-29
Use official Gradle GitHub actions

liblit:
Presumably the Gradle developers know best how to configure things like caching.  I'm particularly interested in seeing our GitHub actions use better caching because we run those actions more frequently now due to our stricter `master`-branch merge policies.

Historically, we have disabled Gradle caching during CI/CD builds: `./gradlew --no-cache ...`.  Presumably we wanted to get a "clean" build every time for the strictest possible checking.  However, clean builds with no cache take a long time!  Furthermore, I believe that our build caching is extremely trustworthy.  I cannot remember the last time that something broken ended up in my cache.  Likewise, I cannot remember the last time that a build succeeded by using the cache, but should have failed if the cache were empty.

I propose that we allow caching, as in this commit.  We can do cache-free hermetic builds when preparing a new official release, but for the many CI/CD builds that happen in between releases, I think that some caching may give us a nice speedup.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -32,39 +32,33 @@ jobs:
           path: ~/.goomph
           key: ${{ runner.os }}-goomph-${{ hashFiles('build.gradle') }}
           restore-keys: ${{ runner.os }}-goomph-
-      - name: Cache Gradle caches
-        uses: actions/cache@v1
-        with:
-          path: ~/.gradle/caches
-          key: ${{ runner.os }}-gradle-caches-${{ hashFiles('**/*.gradle') }}
-          restore-keys: ${{ runner.os }}-gradle-caches-
-      - name: Cache Gradle wrapper
-        uses: actions/cache@v1
-        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -165,4 +135,6 @@ jobs:
         env:
           ORG_GRADLE_PROJECT_SONATYPE_NEXUS_USERNAME: ${{ secrets.SONATYPE_NEXUS_USERNAME }}
           ORG_GRADLE_PROJECT_SONATYPE_NEXUS_PASSWORD: ${{ secrets.SONATYPE_NEXUS_PASSWORD }}
-        run: ./gradlew publishAllPublicationsToMavenRepository
+        uses: gradle/gradle-build-action@v2
+        with:
+          arguments: publishAllPublicationsToMavenRepository
```
msridhar, line 140: Note to self: double check that the artifacts actually get uploaded after this change

msridhar (approved):
Nice, hope this gives us a good CI speedup!

Issue #963, opened 2021-09-11
LibraryVersionTest does not pass for libraries from JDK 9+


msridhar:
Currently it's disabled.  For JDK 8 and earlier, we would read the Specification-Version from the manifest file of `rt.jar`:

https://github.com/wala/WALA/blob/748b55661f47c099288374d0d4de734916bb714f/com.ibm.wala.core/src/main/java/com/ibm/wala/ipa/callgraph/AnalysisScope.java#L373-L375

But there is no such manifest file in the JDK `.jmod` files.  However, there is a `classes/module-info.class` file.  If we add enough support to Shrike to at least read out the module name from such a file, I think we might be able to figure out the spec version from that.  Here's what I see when running `javap` on the `module-info.class` for `java.base.jmod` on JDK 11:

```
module java.base@11.0.11 {
  exports java.io;
  exports java.lang;
  exports java.lang.annotation;
  exports java.lang.invoke;
  ...
```

It's a bit hacky but it should work.

PR #1023, opened 2021-10-10
Remove some test exclusions for JDK 11, and document the remaining one

msridhar:
These tests that were excluded seem to be passing now.  There is one test that still fails on JDK 11, due to #963


msridhar:
Nevermind, messed up my local testing

PR #1026, opened 2021-10-10
Run all CI jobs on JDK 11

msridhar:
It seems that with the current Gradle config, all Gradle tasks getting run on JDK 8 on Linux also pass on JDK 11.  So, this PR enables them for both versions.  Note that there is still some work to do to get Dalvik tests running on JDK 11, but they are disabled in the Gradle config for `com.ibm.wala.dalvik` on JDK 11 so this change still works.

We also update the JDK distribution to "temurin", the new name for AdoptOpenJDK:

https://github.com/actions/setup-java#supported-distributions

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -36,20 +36,14 @@ jobs:
         uses: actions/setup-java@v2
         with:
           java-version: ${{ matrix.java }}
-          distribution: 'adopt'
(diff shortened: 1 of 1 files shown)


liblit:
> It seems that with the current Gradle config, all Gradle tasks getting run on JDK 8 on Linux also pass on JDK 11.

Nice!

PR #1046, opened 2021-11-08
Use `constrainTransitivesToThisRelease` to fix Eclipse transitive dependencies

msridhar:
This is a new feature in the Eclipse Maven Central plugin; see https://github.com/diffplug/goomph/issues/149.  By enabling it, we can avoid all the pain with Eclipse's dependency ranges, and also get rid of a bunch of ugly Gradle configuration and our lockfiles.


liblit (approved):
You had me at “+8 -883” (GitHub’s tally of lines added/removed).

PR #1052, opened 2021-11-14
Temporary fix for building CAst native code on M1 Macs

### LINKED ISSUE #1045: WALA does not compile on Apple Silicon Macs
I currently know of two issues.

First, the CAst C++ code does not compile, I believe due to a Gradle issue.  I filed https://github.com/gradle/gradle/issues/18876 on that.

Second, the wala.ide code requires a native build of SWT, but such builds only starting getting distributed with Eclipse 2021-09 (version 4.21).  I would like to get code like `IFDSExplorer` running on Apple Silicon at least, but this should probably wait until #462 is resolved, to minimize the amount of related code to deal with.

msridhar:
This trick convinces Gradle to run the built-in Clang / gcc on M1 Macs to compile native CAst code, when executing with a native ARM JVM.  I don't think it should cause any harm on other platforms, since all we support otherwise is 64-bit x86.  If / when https://github.com/gradle/gradle/issues/18876 is fixed, we can remove this change.

Fixes #1045

### DIFF
--- a/com.ibm.wala.cast/cast/build.gradle
+++ b/com.ibm.wala.cast/cast/build.gradle
@@ -5,7 +5,9 @@ plugins {
 library {
 	// Temporary change to build on M1 Mac machines, until https://github.com/gradle/gradle/issues/18876
 	// is fixed
-	targetMachines.add(machines.macOS.x86_64)
+	if (osName.equals('Mac OS X') && archName.equals('aarch64')) {
+		targetMachines.add(machines.macOS.x86_64)
+	}
 	binaries.whenElementFinalized {
 		compileTask.get().configure {
 			macros.put('BUILD_CAST_DLL', '1')
--- a/com.ibm.wala.cast/cast/build.gradle
+++ b/com.ibm.wala.cast/cast/build.gradle
@@ -3,6 +3,9 @@ plugins {
 }
 
 library {
+	// Temporary change to build on M1 Mac machines, until https://github.com/gradle/gradle/issues/18876
+	// is fixed
+	targetMachines.add(machines.macOS.x86_64)
 	binaries.whenElementFinalized {
 		compileTask.get().configure {
 			macros.put('BUILD_CAST_DLL', '1')
--- a/com.ibm.wala.cast/smoke_main/build.gradle
+++ b/com.ibm.wala.cast/smoke_main/build.gradle
@@ -5,6 +5,9 @@ plugins {
 evaluationDependsOn(':com.ibm.wala.cast:xlator_test')
 
 application {
+	// Temporary change to build on M1 Mac machines, until https://github.com/gradle/gradle/issues/18876
+	// is fixed
+	targetMachines.add(machines.macOS.x86_64)
 	dependencies {
 		implementation project(':com.ibm.wala.cast:cast')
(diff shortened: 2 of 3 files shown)


liblit (approved):

msridhar:
Sorry for requesting the review @liblit, this clearly needs some more work

msridhar:
Ok, I think this works now

liblit:
> Sorry for requesting the review @liblit, this clearly needs some more work

No problem at all!

liblit (approved):

bradAtTraceable:
I'm having a similar problem.  I'm running Gradle on an Ubuntu Docker container; the host is a Mac using the Apple M1 Pro chip, which causes the Ubuntu container to have an architecture of 'aarch64'.  The Gradle project has some c++ code, and gcc is installed on the Ubuntu Docker image.  The problem is that Docker won't invoke gcc to compile the c++ code; the message is: "Don't know how to build for host operating system 'Linux' architecture 'aarch64'".

Based on your comments, it looks like we should be able to override this by updating 'targetMachines'.  Is that correct?

Another problem is that the build.gradle file is written in Kotlin, which I am not familiar with.  If updating targetMachines will fix my problem, how can I code this with Kotlin?

liblit:
> Another problem is that the build.gradle file is written in Kotlin, which I am not familiar with. If updating targetMachines will fix my problem, how can I code this with Kotlin?

All WALA `build.gradle` scripts are written in Groovy, not Kotlin.

msridhar:
> Based on your comments, it looks like we should be able to override this by updating 'targetMachines'. Is that correct?

This seemed to work for us, but it was for building natively on Mac.  You'd have to test to see if it works for building on Linux with aarch64.  I'd expect this to be an issue anytime Linux is running on ARM, not just inside Docker containers on the M1.

bradAtTraceable:
It's not letting my script change the targetMachines.  I'm getting the following message:

A problem occurred configuring project ':jni-libtraceable-cpp'.
> The value for property 'targetMachines' is final and cannot be changed any further.

msridhar:
Ok sorry not sure how to help you further

msridhar:
FYI @bradAtTraceable our scripts seemed broken for some Gradle releases but seem to be working again in Gradle 7.5.1

dfsutherland:
FWIW:  This continues to fail on M1 Macs.

This still fails with Gradle 8.0.1 on MacOS 13.2.1 running on an M1Pro-based laptop. The error message is as follows:

```ld: warning: ignoring file /Library/Java/JavaVirtualMachines/temurin-11.jdk/Contents/Home/lib/server/libjvm.dylib, building for macOS-x86_64 but attempting to link with file built for macOS-arm64
Undefined symbols for architecture x86_64:
  "_JNI_CreateJavaVM", referenced from:
      launch_jvm(char*) in launch.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

msridhar:
@dfsutherland did you get that error while trying to build WALA?  If so, what Gradle command did you run?

dfsutherland:
Yes. The gradle command was ./gradlew build

PR #1053, opened 2021-11-14
Remove an unused XML file

liblit:
IntelliJ IDEA's headless inspections report several errors in this file, but we no longer use it, so it may as well go away.


msridhar (approved):
Can we try removing the `jardesc` files too?  I have no idea what those are for.  In any case this change can land separately.

liblit:
> Can we try removing the `jardesc` files too?

Will do.

PR #1054, opened 2021-11-14
Hash and equality-compare arrays deeply

liblit:
"Deeply" here means considering the hash codes or equality comparisons of array elements.  This change doesn't fix any reported bug that I am aware of, but it does resolve two IntelliJ IDEA warnings.

I could also just suppress those warnings if we are convinced that deep hashing and equality-comparison are never needed here.


msridhar (approved):
Looks like it's fixing a real bug to me!

PR #1055, opened 2021-11-14
Remove unused `*.jardesc` files

liblit:
As suggested [here](https://github.com/wala/WALA/pull/1053#pullrequestreview-805505696).


msridhar (approved):
👍

PR #1056, opened 2021-11-14
Preserve IntelliJ IDEA inspection results as workflow artifact

liblit:
If any inspections fail, it can be useful to examine the complete inspection results manually. This change has GitHub preserve those results whenever we run the IntelliJ IDEA inspections workflow.

Artifacts are preserved for 90 days by default, but that seems much longer than we'll need these results to be retained. We run this workflow once per week, so let's retain those results for two weeks.


msridhar (approved):
Depending on the size of this file, I'd be fine with a longer retention time, but either way is fine

liblit:
> Depending on the size of this file, I'd be fine with a longer retention time, but either way is fine

This file is currently under 67 KB. So yes, we could afford to keep many of them around. I just don't anticipate any need to do so.

PR #1064, opened 2021-11-26
Fix SourceMapTest so it can run on CI

msridhar:
In order for the `SourceMapTest` tests to pass, they must be able to see files inside the wala.core `testSubjects` source directory.  This PR adjusts the `testRuntimeOnly` path to include this directory, so the tests now pass.  This change enables removing the `analyzingJar` configuration that was added to skip the tests.

### DIFF
--- a/com.ibm.wala.core.java11/build.gradle
+++ b/com.ibm.wala.core.java11/build.gradle
@@ -47,7 +47,6 @@ tasks.named('processTestResources') {
 
 tasks.named('test') {
 	maxHeapSize = '1500M'
-	systemProperty 'com.ibm.wala.junit.analyzingJar', 'true'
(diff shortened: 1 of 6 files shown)


msridhar:
In order for the `SourceMapTest` tests to pass, they must be able to see files inside the wala.core `testSubjects` source directory.  This PR adjusts the `testRuntimeOnly` path to include this directory, so the tests now pass.  This change enables removing the `analyzingJar` configuration that was added to skip the tests.

liblit (approved):
Approving with a few tweaks from me. I took the liberty of pushing those tweaks directly to the `msridhar:fix-source-filename-test` branch rather than going through the extra step of asking @msridhar to apply them on my behalf.

msridhar:
Thanks, @liblit!

PR #1066, opened 2021-11-27
Revert unpacking NodeJs code to test resources rather than main resources

### LINKED ISSUE #1033: NullPointerException Introduced by a breaking change
Hi,

I find that WALA (4aa169c82fe61a08c8e83b9c0a908b254339bc80) throws NPE when parsing node.js programs, which is caused by this method
https://github.com/wala/WALA/blob/4aa169c82fe61a08c8e83b9c0a908b254339bc80/com.ibm.wala.cast.js.nodejs/src/main/java/com/ibm/wala/cast/js/nodejs/NodejsRequiredCoreModule.java#L92-L96
Further investigation points that https://github.com/wala/WALA/commit/7a029b7054e41107f13b7692ebf9b6c00124bd1c does not work with the above code (no core-modules is downloaded to resource/ )

I manually copy core-modules to com.ibm.wala.cast.js.nodejs/src/main/resources and now it works.

Please consider fixing it. thanks.

msridhar:
This reverts commit a9157d8cce595798baffef04ae62e27a9e96c909.

Fixes #1033.  Unfortunately, I am not sure why I made this change in the first place, but it seems wrong to me now, as the built `wala.cast.js.nodejs` artifact won't work without these resources packaged in.  I may have been thinking about how we could package up `nodejs` for release to Maven Central (since we wouldn't want to re-distribute NodeJs code in such a case).  But since we still haven't done that, for now revert the change.

### DIFF
--- a/com.ibm.wala.cast.js.nodejs/build.gradle
+++ b/com.ibm.wala.cast.js.nodejs/build.gradle
@@ -36,7 +36,7 @@ final unpackNodeJSLib = tasks.register('unpackNodeJSLib', Copy) {
 }
 
 tasks.named('processResources') {
-	dependsOn 'unpackNodeJSLib'
+	dependsOn unpackNodeJSLib
 }
 
 tasks.named('test') {
--- a/com.ibm.wala.cast.js.nodejs/build.gradle
+++ b/com.ibm.wala.cast.js.nodejs/build.gradle
@@ -31,6 +31,12 @@ final unpackNodeJSLib = tasks.register('unpackNodeJSLib', Copy) {
 		}
 	}
 
+	// It is important to unpack the the NodeJs library files into the main resources directory,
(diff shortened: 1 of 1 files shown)


liblit:
Are we missing an automated test that would have told us that the previous commit had a problem?

msridhar:
> Are we missing an automated test that would have told us that the previous commit had a problem?

It's a bit tricky, as this would require something like an integration test.  What is required is that these unpacked files be available for loading by a classloader.  Unpacking them to the test resources directory will satisfy this requirement for unit tests, but not for a case where the packaged jar is being used by third-party code.

I don't think we need to add an integration test for this case, but I will add a comment to the `build.gradle` indicating why it is important to unpack these files to the main resources directory.

liblit:
> I will add a comment to the `build.gradle` indicating why it is important to unpack these files to the main resources directory.

Good call. Thanks!

liblit (approved):

PR #1075, opened 2021-12-05
Extract Gradle logic for Java-containing subjects

liblit:
Previously, the top-level `build.gradle` iterated over all subprojects and applied lots of additional build-configuration logic to subprojects that contain Java source code.  Now, that Java-specific configuration logic is extracted into its own `wala-java.gradle` script plugin, which each individual subproject can load if appropriate.  This change makes the top-level `build.gradle` a bit smaller and easier to read through.


gradle-mvn-push.gradle
```diff
@@ -4,6 +4,8 @@ apply plugin: 'signing'
 final isSnapshot = version.contains('SNAPSHOT')
 
 final skipTestFixtures() {
+	println "*** $project"
```
msridhar, line 7: Is this an intended change?

wala-java.gradle
```diff
@@ -0,0 +1,106 @@
+apply plugin: 'eclipse'
+apply plugin: 'java-library'
+apply plugin: 'java-test-fixtures'
+apply plugin: 'maven-publish'
```
msridhar, line 4: Will this lead us to publish more artifacts than we did before?  I didn't think we included `maven-publish` for every project with Java source before.

msridhar (commented):
Couple of questions

gradle-mvn-push.gradle
liblit: Oops, nope. I'll remove this shortly. Thanks for being so attentive!

wala-java.gradle
liblit: I think we were already including `maven-publish` for every subproject with Java source:

https://github.com/wala/WALA/blob/17064550328e82c68fb1a8464b3c26ad56c83538/build.gradle#L57-L61

msridhar: Ah yes, I was confused. `gradle-mvn-push.gradle` is only included in some places

msridhar (approved):

liblit:
This change worked fine on a side branch, but caused [a regression in a task that only runs on `master`](https://github.com/wala/WALA/runs/4424048891?check_suite_focus=true). I am investigating.

PR #1085, opened 2021-12-23
Updated Native Models

MadhuNimmo:
Added stubs/models to support more native method calls.


msridhar:
@MadhuNimmo many regression tests are failing.  Can you take a look and fix?

PR #1086, opened 2022-01-03
Remove GitHub action that hangs on Ubuntu

liblit:
Running `./gradlew runInspections` works fine on Fedora 35.  However, on Ubuntu 20.04, it hangs.  The headless IntelliJ IDEA subprocess fails to exit once the inspections have all been run.  I don't know why.  Until we can make this work, we shouldn't run this as a weekly GitHub action.


msridhar (approved):
Bummer; hope we can figure this out sometime in the future

PR #1093, opened 2022-03-03
Resource Leak Fix for BufferedInputStream in captureStdErr

Aramismas:


com.ibm.wala.util/src/main/java/com/ibm/wala/util/processes/BasicLauncher.java
```diff
@@ -10,6 +10,7 @@
  */
 package com.ibm.wala.util.processes;
 
+import java.io.BufferedInputStream;
```
msridhar, line 13: This change seems irrelevant?

com.ibm.wala.util/src/main/java/com/ibm/wala/util/processes/Launcher.java
```diff
@@ -178,7 +178,7 @@ void blockingDrain() throws IOException {
     return result;
   }
 
-  protected Drainer captureStdErr(Process p) {
+  protected Drainer captureStdErr(Process p) throws IOException {
```
msridhar, line 181: why this change?

msridhar (changes requested):
Let's remove unnecessary changes before I review further

msridhar:
Also, please run `./gradlew googleJavaFormat` and check in the changes, to pass the formatting checks on CI.

com.ibm.wala.util/src/main/java/com/ibm/wala/util/processes/Launcher.java
Aramismas: This change was related to the first solution but should be removed as it is irrelevant now.

com.ibm.wala.util/src/main/java/com/ibm/wala/util/processes/BasicLauncher.java
Aramismas: I will remove this import.

msridhar:
I believe we have decided these `close()` calls are not necessary, so I am closing out this PR.  Still @Aramismas we appreciate the contribution!

PR #1102, opened 2022-06-26
Avoid configuring some tasks prematurely

liblit:
For the best Gradle performance, we should always use Gradle task-configuration-avoidance APIs to configure tasks only when truly needed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -59,7 +59,7 @@ tasks.register('aggregatedJavadocs', Javadoc) { aggregated ->
 	options.author true
 
 	subprojects.each { proj ->
-		proj.tasks.withType(Javadoc) { javadocTask ->
+		proj.tasks.withType(Javadoc).configureEach { javadocTask ->
 			aggregated.source += javadocTask.source
 			aggregated.classpath += javadocTask.classpath
 			aggregated.excludes += javadocTask.excludes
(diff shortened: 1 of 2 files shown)


msridhar (approved):
How do you even find these issues?  Is there some kind of static / dynamic analysis to identify them?  Would love to fix these up on some other projects I work on

liblit:
> How do you even find these issues? Is there some kind of static / dynamic analysis to identify them?

This specific PR was prompted by [this blog post](https://melix.github.io/blog/2022/05/gradle-laziness.html), which prompted me to search WALA's build logic for any uses of `tasks.withType(…)` instead of `tasks.withType(…) { … }`. There were not many, so it was easy to examine each more closely.

See also [Gradle’s documentation on task configuration avoidance](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html) for much more information, including [a useful table of APIs to avoid and their suggested replacements](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:old_vs_new_configuration_api_overview). This page also shows [how to use `--scan` to identify tasks that were configured prematurely during a specific build](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:task_configuration_avoidance_troubleshooting).

PR #1119, opened 2022-08-26
Improved equal function in SSAInstruction

felixmaechtle:
Currently, two instructions are equal if they have the same `instructionIndex` within their function. Hence the first instruction of a method is equivalent to any other first instruction. 

This fix adds a comparison of the hash values as well as the string representation.


msridhar:
Generally you should not be comparing equality of `SSAInstruction`s across IRs I think. I would first check that the IR is the same (or that the CHNode is the same) and then the instruction. I'm concerned this change will impact performance. Do you have a specific use case for it?

felixmaechtle:
Hello,

Your suggestion is my workaround now. I use `SSAInstruction`s as keys in a `HashMap`. This led to weird behaviour, as some instructions pointed to the same value within the `HashMap`.

The reason is that there are some collisions in the hash values. To intercept this, HashMap uses `.equals` internally.

Do you think it will be a considerable drop in performance? As Java is lazy, it should only impact equal instructions, shouldn't it?

msridhar:
> Do you think it will be a considerable drop in performance? As Java is lazy, it should only impact equal instructions, shouldn't it?

This was a long-standing design decision of WALA, and I'd rather not tinker with it unless we have a very strong reason to do so.  I've always been able to implement my analyses comparing instructions by CGNode + SSAInstruction and I haven't run into any issues.

I am going to close for now, but if you strongly disagree please feel free to continue the discussion.

PR #1123, opened 2022-09-13
Update various dependencies

msridhar:
Limited to what seemed like "safe" updates that I don't expect to cause test failures.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,7 +11,7 @@ plugins {
 	id 'all.shared.gradle.file-lister' version '1.0.2'
 	id 'com.diffplug.eclipse.mavencentral' version '3.33.2' apply false
 	id 'com.dorongold.task-tree' version '2.1.0'
-	id 'com.github.ben-manes.versions' version '0.39.0'
+	id 'com.github.ben-manes.versions' version '0.42.0'
 	id 'com.github.sherter.google-java-format' version '0.9'
 	id 'de.undercouch.download'
 	id 'nebula.lint' version '17.2.3'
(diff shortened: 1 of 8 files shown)


liblit (approved):
Approved, but it looks like the macOS build needs to be restarted after a temporary network outage.

msridhar:
@liblit I think I made the failed download more reliable by switching to https. Can I get another review? I'll do a follow up where I search for other download URLs that we should switch.

liblit:
> @liblit I think I made the failed download more reliable by switching to https.

If I remember correctly, the failed build was reporting "no route to host" or something along those lines.  That's a failure down at the level of IP routing: below TCP and below application protocols like HTTP or HTTPS.  I see no reason why switching to HTTPS would improve reliability.  More likely you just got luckier with the retry, as we've seen with many other transient network outages in our GitHub actions.

msridhar:
> If I remember correctly, the failed build was reporting "no route to host" or something along those lines. That's a failure down at the level of IP routing: below TCP and below application protocols like HTTP or HTTPS. I see no reason why switching to HTTPS would improve reliability. More likely you just got luckier with the retry, as we've seen with many other transient network outages in our GitHub actions.

Looks like you may have been right.  I still think the HTTPS change is a good one, but I can do that in a separate PR.  If you could re-stamp the original set of changes that would be great :-)

liblit (approved):

liblit:
> I still think the HTTPS change is a good one, but I can do that in a separate PR.

No objection from me. I'll point out, though, that we verify downloads' signatures against expectations. So even with HTTP, we are fairly safe from men-in-the-middle.

PR #1127, opened 2022-09-20
Make demand-driven points-to analysis robust to missing call sites

msridhar:
This can arise in rare cases where the up-front call graph is a `CHACallGraph`, which may consider call sites not present after constructing IR (which does some dead-code elimination).  It's hard to create a test for this scenario, but the code changes are safe, in that they are adding nullness / array length checks that were not present before.


msridhar:
Waiting on some external testing before merging

liblit (approved):

PR #1128, opened 2022-09-29
Add Error Prone to build

msridhar:
We only run Error Prone on JDK 11 builds, as recent Error Prone releases do not run on JDK 8.  We also disable WARNING level checks for now, to reduce the noise in the build output (we get hundreds of WARNING-level messages from Error Prone, which we should try to fix gradually).

This PR also fixes all ERROR-level issues reported by Error Prone.

### DIFF
--- a/com.ibm.wala.cast/src/main/java/com/ibm/wala/cast/loader/AstMethod.java
+++ b/com.ibm.wala.cast/src/main/java/com/ibm/wala/cast/loader/AstMethod.java
@@ -167,10 +167,6 @@ public abstract class AstMethod implements IMethod {
     return cfg;
   }
 
-  public boolean hasCatchBlock() {
-    return hasCatchBlock();
-  }
-
   public SymbolTable symbolTable() {
     return symtab;
   }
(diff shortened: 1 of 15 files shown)


com.ibm.wala.scandroid/src/main/java/org/scandroid/util/AndroidAnalysisContext.java
```diff
@@ -150,36 +148,6 @@ public static SSAPropagationCallGraphBuilder makeVanillaZeroOneCFABuilder(
         ZeroXInstanceKeys.ALLOCATIONS | ZeroXInstanceKeys.CONSTANT_SPECIFIC);
   }
 
-  /**
```
msridhar, line 153: Deleting since this method just calls itself in an infinite recursion.  Clearly no other code has been calling it

com.ibm.wala.cast/src/main/java/com/ibm/wala/cast/loader/AstMethod.java
```diff
@@ -167,10 +167,6 @@ protected AstMethod(
     return cfg;
   }
 
-  public boolean hasCatchBlock() {
```
msridhar, line 170: Deleting since this method just calls itself in an infinite recursion. Clearly no other code has been calling it

liblit (changes requested):

msridhar:
Thanks for the careful review @liblit!  I believe I have addressed all of your comments.

liblit (approved):

PR #1135, opened 2022-10-09
Address UnusedVariable warnings from Error Prone and turn up the check

msridhar:
I have reviewed these manually and suppressed some warnings, where it seemed to make more sense to keep the unused code.

Also, turn off the ECJ errors on unused locals.  I believe Error Prone should catch all of the same issues (and more), and also Error Prone supports not warning on a variable named `ignored`, for cases where you want to indicate you are deliberately ignoring a returned value.


com.ibm.wala.ide.jdt.test/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -89,7 +89,7 @@ org.eclipse.jdt.core.compiler.problem.reportMethodCanBeStatic=ignore
 org.eclipse.jdt.core.compiler.problem.specialParameterHidingField=disabled
 org.eclipse.jdt.core.compiler.problem.staticAccessReceiver=ignore
 org.eclipse.jdt.core.compiler.problem.suppressOptionalErrors=enabled
-org.eclipse.jdt.core.compiler.problem.suppressWarnings=enabled
+org.eclipse.jdt.core.compiler.problem.suppressWarnings=disabled
```
msridhar, line 92: For this one ECJ was complaining that a `@SuppressWarnings` was unnecessary, but it was necessary for the Error Prone checker.  This is for code compiled by Maven only so I'm not concerned about dialing down this check (which is only for unnecessary suppressions anyway)

liblit (approved):

PR #1140, opened 2022-10-14
Add a bounded version of the ACG algorithm

MadhuNimmo:
Implemented Bounded Call Graph Strategy allowing a call graph to be built with a limit on how deep it can analyze function call instructions.  The new algorithm is exposed through the `buildBoundedCG()` and `buildScriptDirBoundedCG()` APIs in the `FieldBasedCGUtil` class.


msridhar (approved):

msridhar:
This change looks good to me!  Note that I reviewed previous versions of this change before it was opened as this PR.

PR #1145, opened 2022-10-29
Fix Error Prone JdkObsolete warnings

msridhar:
See https://errorprone.info/bugpattern/JdkObsolete

Mostly involved replacing `LinkedList` usage with either `ArrayList` or `ArrayDeque` based on context.  Where we had to suppress, we left a comment.  Also disabled the ECJ error on unhandled tokens inside `@SuppressWarnings` since it doesn't know about JdkObsolete.


liblit:
Do we have any benchmarks that would tell us whether these data-structure changes affected performance?

msridhar:
> Do we have any benchmarks that would tell us whether these data-structure changes affected performance?

Nothing rigorous unfortunately. What I can do is run the call graph tests before and after. Those are the most performance-sensitive tests we have. May not cover everything but I think it's the best we can do. I'll try it.

msridhar:
@liblit I ran this command a few times to try to measure perf before and after:
```
time ./gradlew cleanTest :com.ibm.wala.core:test --tests com.ibm.wala.core.tests.callGraph.CallGraphTest --console=plain --no-build-cache
```
Ignoring the first run for each config (where everything is being compiled), this command takes roughly 58-59s on my machine, both on the master branch and the branch for this PR.  There may be other perf regressions that this test doesn't expose, but I think it gives some confidence that there is not a major regression in core WALA functionality (since this test hits IR construction, call graph construction, and pointer analysis heavily).

liblit:
I'm convinced. Thanks, Manu!

liblit (approved):

PR #1150, opened 2022-11-10
Bump Error Prone Gradle plugin version to 3.0.1

msridhar:
This version allows for a slight simplification in configuration, but otherwise works the same.


liblit (approved):

msridhar:
@liblit could I get another approval from you?  Sorry, the PR had an unintended change before.

PR #1153, opened 2022-11-20
Remove Gradle subprojects that don't do anything

liblit:
These subprojects have no sources and no `build.gradle` files.  They define no useful tasks and provide nothing of value to the rest of the Gradle build configuration.  There's really no benefit in treating these subdirectories as subprojects.

Could these subdirectories be removed entirely?  Or are they important somehow when this project is opened in Eclipse instead of IntelliJ IDEA?

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,7 +1,6 @@
 rootProject.name = 'com.ibm.wala'
 
 include(
-	'com.ibm.wala-repository',
 	'com.ibm.wala.cast',
 	'com.ibm.wala.cast:smoke_main',
 	'com.ibm.wala.cast:xlator_test',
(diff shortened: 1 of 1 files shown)


msridhar (approved):
If you like, we can try removing these directories as well.  If all the CI jobs pass (including the Maven job), I'm fine with it.

liblit:
> If you like, we can try removing these directories as well. If all the CI jobs pass (including the Maven job), I'm fine with it.

Understood!

As it turns out:

* `com.ibm.wala-repository` is used by the Maven job,
* `com.ibm.wala.ide_feature` is used by the Maven job,
* `com.ibm.wala.tests.ide_feature` is not used by anything, and
* `com.ibm.wala_feature` does not exist.

I've submitted a separate pull request to remove `com.ibm.wala.tests.ide_feature`: #1155.

PR #1155, opened 2022-11-20
Remove an unbuilt, unused subdirectory

liblit:
Nothing builds in this subdirectory using either Gradle or Maven.  The contents of this directory do not appear to be used by anything else.


msridhar (approved):
Thanks for the cleanup!

PR #1157, opened 2022-11-22
Add NullAway checks to "com.ibm.wala.util` module.

nimakarimipour:
This PR enables [NullAway](https://github.com/uber/NullAway) checks on `com.ibm.wala.util` module. 

Enabling NullAway for this module required injections of multiple `@Nullable`, `@NullUnmarked` and `@SuppressWarnings` annotations directly to the source code. All annotations are added automatically using [Annotator](https://github.com/ucr-riple/NullAwayAnnotator). Please note that all warnings from NullAway are resolved through annotations and it does not contain any behavior or code change.

These annotations are inferred based on the usage of the elements within the `util` module and 14 other downstream dependencies.

-----
Build configuration modifications will be changed later after annotations are approved by the review process.


liblit:
> Build configuration modifications will be changed later after annotations are approved by the review process.

I'm not sure how to interpret this. Should we provide feedback on changes to the Gradle build infrastructure? Or should we ignore all of that and focus only on annotations added to Java files?

msridhar:
Hi @liblit yes I think what @nimakarimipour is asking for is feedback on the annotations first (which I can look at), but we won't approve the PR until all the build config changes are finalized.  The reason @nimakarimipour wants to keep the build config changes is in case we find a problem with the annotations reflecting a problem in his tool, which would then require re-running the tool.

@nimakarimipour looks like CI is failing for the PR; can you take a look?

liblit:
OK, sounds good to me.

com.ibm.wala.util/src/main/java/com/ibm/wala/fixpoint/IntSetVariable.java
```diff
@@ -118,6 +120,7 @@ public boolean contains(int b) {
   }
 
   /** @return the value of this variable as a MutableSparseIntSet ... null if the set is empty. */
+  @NullUnmarked /* Annotator://local 1 */
```
msridhar, line 123: @nimakarimipour one quick note here is that the `// local` comment is arguably slightly misleading.  It is in fact true that if `@NullUnmarked` is removed, an error will appear on this function since the return is not `@Nullable`.  But, the local fix is easy; probably the real issue is that callers are using the return value without checking if it is null, etc.  I don't know of any kind of easy fix for this; just food for thought.

com.ibm.wala.qual/src/main/java/com/ibm/wala/qual/Initializer.java
```diff
@@ -0,0 +1,41 @@
...
+ * FOR A PARTICULAR PURPOSE AND FURTHER DISCLAIMS ANY STATUTORY
+ * WARRANTY OF NON-INFRINGEMENT. THE SOFTWARE AND ACCOMPANYING
+ * DOCUMENTATION, IF ANY, PROVIDED HEREUNDER IS PROVIDED "AS
+ * IS". REGENTS HAS NO OBLIGATION TO PROVIDE MAINTENANCE, SUPPORT,
+ * UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
+ */
+
+package com.ibm.wala.qual;
+
+public @interface Initializer {}
```
msridhar, line 41: Instead of this annotation, can we add a `compileOnly` dependence on `"com.facebook.infer.annotation:infer-annotation:0.11.0"` to get it?

com.ibm.wala.qual/src/main/java/com/ibm/wala/qual/NullUnmarked.java
```diff
@@ -0,0 +1,103 @@
...
+ * within null-unmarked code may have unspecified nullness. (<a
+ * href="https://bit.ly/3ppb8ZC">Why?</a>)
+ *
+ * <p>Unspecified nullness is (will be) explained comprehensively in the <a
+ * href="http://jspecify.org/docs/user-guide">User Guide</a>.
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({TYPE, METHOD, CONSTRUCTOR, PACKAGE})
+public @interface NullUnmarked {}
```
msridhar, line 103: Instead of adding this here can we just depend on `"org.jspecify:jspecify:0.3.0-alpha-2"`?

com.ibm.wala.util/src/main/java/com/ibm/wala/dataflow/graph/DataflowSolver.java
```diff
@@ -138,7 +144,7 @@ private int mapIt(int i, Object[] allVars, Map<Object, V> varMap) {
      * record that variable (n1, in1) is the same as variable (n2,in2), where (x,true) = IN(X) and
      * (x,false) = OUT(X)
      */
-    public void union(Object n1, Object n2) {
+    public void union(@Nullable Object n1, @Nullable Object n2) {
```
msridhar, line 147: A bit strange that the tool made these parameters `@Nullable`; I would not have done this since the first thing this function does is assert they are not null.  I think this is a case of the tool propagating `@Nullable` and not seeing any errors, but a human would have done it differently

com.ibm.wala.util/src/main/java/com/ibm/wala/util/collections/AbstractMultiMap.java
```diff
@@ -82,7 +83,7 @@ public Set<V> get(K key) {
    * @see AAA.util.MultiMap#put(K, V)
    */
   @Override
-  public boolean put(K key, V val) {
+  public boolean put(K key, @Nullable V val) {
```
msridhar, line 86: This is another case where I would have made a different decision; I don't think we want to allow nulls here (though in this case I'm actually not sure, maybe some code relies on that

com.ibm.wala.util/src/main/java/com/ibm/wala/util/collections/NonNullSingletonIterator.java
```diff
@@ -12,17 +12,18 @@
...
  * allows this class to be slightly more efficient than Collections.iterator()
  */
 public class NonNullSingletonIterator<T> implements Iterator<T> {
 
-  private T it;
+  @Nullable private T it;
 
   /** @param o the single object in this collection, must be non-null */
-  public NonNullSingletonIterator(T o) {
+  public NonNullSingletonIterator(@Nullable T o) {
```
msridhar, line 26: Another case where the parameter clearly shouldn't be `@Nullable`

com.ibm.wala.util/src/main/java/com/ibm/wala/util/collections/TwoLevelVector.java
```diff
@@ -32,6 +34,8 @@
   private int maxPage = -1;
 
   /** @see com.ibm.wala.util.intset.IntVector#get(int) */
+  @NullUnmarked
+  @Nullable
```
msridhar, line 38: Strange to make this both `@NullUnmarked` _and_ `@Nullable`

msridhar (commented):
Overall this looks really promising!  I think I'd like a way to force some elements not to be `@Nullable` and then re-run the tool.  @nimakarimipour will discuss with you

msridhar:
Here is how I think we should proceed on this pull request.  I see some changes I'd like to make to the generated annotations.  But, for research purposes, it will be helpful to us if we can cleanly compare the initial annotations generated by the tool with the final version that gets committed.  So, I propose the following:

1. @nimakarimipour, you get the PR cleaned up by removing all the extra build infra you added to run the annotator.
2. @liblit assuming you have the time, you then review and finally sign off on any build script changes.
3. Finally, I will commit some changes to the annotations that I want, and then we will land.

This way, we can keep the state after step 2 pushed on a branch, which will facilitate a clean comparison with whatever manual changes I needed to do.  Sound good?

Also, @nimakarimipour regarding the `com.ibm.wala.qual` module I think we can get rid of it.  You can move the `@Initializer` annotation to `com.ibm.wala.util` (which everything else depends on anyway).  And let's use both `@NullUnmarked` and `@Nullable` from the latest JSpecify artifact, so we don't need the dependence on `''com.google.code.findbugs:jsr305:3.0.2'` either.  For switching up the `@Nullable` annotation used, it may be easiest to re-run the annotator one more time.

com.ibm.wala.qual/src/main/java/com/ibm/wala/qual/Initializer.java
nimakarimipour: Sure, removed it be740cf0bd769d857d5a3d5f4c71ed1758890741

nimakarimipour:
@msridhar Thank you for the instructions. I cleaned up locally declared annotations and used `specify` and `facebook.infer` annotations. I am rerunning Annotator now and will push the new changes once it is finished.

nimakarimipour:
@msridhar I re-ran the Annotator with the requested annotations and cleaned up its scripts. Please let me know if there is anything else.

com.ibm.wala.util/src/main/java/com/ibm/wala/Initializer.java
```diff
@@ -0,0 +1,24 @@
...
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ */
+
+package com.ibm.wala;
```
msridhar, line 12: Let's put this in the `com.ibm.wala.qual` package

msridhar:
@liblit I think this is ready for you to review the build configuration changes

buildSrc/src/main/groovy/wala-java.gradle
```diff
@@ -27,9 +27,18 @@ repositories {
 // only use Error Prone on JDK 9+ JVMs
 final useErrorProne = JavaVersion.current().isJava9Compatible()
 
+// checks if NullAway should be used.
+def useNullAway(project) {
```
liblit, line 31: Use `final` here instead of `def`: we don't expect to reassign `useNullAway` to have a different value other than the function being created here.

buildSrc/src/main/groovy/wala-java.gradle
```diff
@@ -27,9 +27,18 @@ repositories {
...
+	final nullawayEnrolledModules = ["com.ibm.wala.util"]
+	return nullawayEnrolledModules.contains(project.name)
```
liblit, line 33: I don't care for using a centralized list of all enrolled modules here. Instead, this should be delegated down to the individual subprojects.

Let's create a `buildSrc/main/groovy/wala-NullAway.gradle` script. Any project that loads that script as a plugin will have the corresponding configuration changes applied to it unconditionally. For example, that script might begin with:

```groovy
plugins {
	id 'java'
}

dependencies {
	annotationProcessor 'com.uber.nullaway:nullaway:0.10.5'
}

// ...continue with any other NullAway-only project tweaks as needed...
```

For now, only `com.ibm.wala.util/build.gradle` would apply this plugin script. If we add NullAway to other subprojects later, then their corresponding `build.gradle` scripts can apply `wala-NullAway.gradle` too.

liblit (changes requested):

msridhar:
@liblit I think your comments on the build scripts have been addressed, if you'd like to take another look.

buildSrc/src/main/groovy/wala-java.gradle
```diff
@@ -25,7 +25,7 @@ repositories {
 }
 
 // only use Error Prone on JDK 9+ JVMs
-final useErrorProne = JavaVersion.current().isJava9Compatible()
+project.ext.useErrorProne = JavaVersion.current().isJava9Compatible()
```
msridhar, line 28: @liblit is this the right way to do this?  I tried just doing `ext.useErrorProne = ...` but that did not make the variable visible inside `wala-NullAway.gradle`.

liblit: Looks fine to me. I'm not sure why it didn't work without `project.`. But I'm also not very concerned about that: once we require Java 11+, this `useErrorProne` check can go away entirely.

liblit (dismissed):

liblit (approved):

liblit (approved):

PR #1162, opened 2022-12-12
Require Java 11 or later

liblit:
Let's party like it's 2018! 🎉

Build all of WALA using Java 11+, for Java 11+. Even the Gradle build engine itself must be running on a Java 11+ JVM. There is no attempt at backward compatibility for Java 1.8: after these changes, WALA jars will contain v55 bytecode for Java 11+ JVMs, and that's that.

These changes also bump WALA's version to 1.6.0; we should presumably release 1.5.9 as the last WALA for Java 1.8 before merging these changes to `master`.

Unfortunately, the Maven GitHub build action is currently failing, with [error diagnostics](https://github.com/wala/WALA/actions/runs/3671408424/jobs/6206590660#step:6:8) that don't give me any clues that I understand. I'm hoping that @msridhar may have some insights. Obviously we won't actually approve this PR or merge these changes until all tests pass.


.github/workflows/continuous-integration.yml
```diff
@@ -9,18 +9,13 @@ concurrency:
   cancel-in-progress: true
 jobs:
   build_gradle:
-    name: "JDK ${{ matrix.java }} on ${{ matrix.os }}"
+    name: "${{ matrix.os }}"
```
msridhar, line 12: Can we actually keep the JDK version in this matrix?  I dream of running tests on JDK 17 in the not-too-distant future, at which point we'd just have to add this all back

com.ibm.wala.core/src/test/java/com/ibm/wala/core/tests/callGraph/ReflectionTest.java
```diff
@@ -844,12 +844,12 @@ public void testForNameThrownExceptions()
     CallGraph cg = CallGraphTestUtil.buildZeroCFA(options, new AnalysisCacheImpl(), cha, false);
     IMethod mainMethod = entrypoints.iterator().next().getMethod();
     List<CGNode> mainCallees =
-        Iterator2List.toList(cg.getSuccNodes(cg.getNode(mainMethod, Everywhere.EVERYWHERE)));
+        Iterator2Collection.toList(cg.getSuccNodes(cg.getNode(mainMethod, Everywhere.EVERYWHERE)));
```
msridhar, line 847: Is this related to JDK 11?  Or some ECJ change?

com.ibm.wala.core/src/testFixtures/java/com/ibm/wala/core/tests/util/WalaTestCase.java
```diff
@@ -63,22 +66,20 @@ protected static void justThisTest(Class<?> testClass) {
...
-        if (result == null) {
-          result = new StringBuilder(s);
-        } else {
-          result.append(File.pathSeparator).append(s);
-        }
-      }
-    }
-    assert result != null : "cannot find " + elt;
-    return result.toString();
+    final String normalizedElt = Paths.get(elt).normalize().toString();
```
msridhar, line 69: Why is this change in this PR?

msridhar (changes requested):
Mostly LGTM!  Just some minor things.

Note that I won't land this until after cutting a new WALA release, as @liblit noted.  Also, we should remember to turn back on the inferrable type arguments check and fix those issues as a follow-up.

.github/workflows/continuous-integration.yml
liblit: OK.

com.ibm.wala.core/src/test/java/com/ibm/wala/core/tests/callGraph/ReflectionTest.java
liblit: Presumably it was motivated by some new diagnostic. But this change certainly is compatible with Java 1.8. Therefore, I will pull it out into a separate, tiny pull request of its own that we can merge before the big Java 11 change.

com.ibm.wala.core/src/testFixtures/java/com/ibm/wala/core/tests/util/WalaTestCase.java
liblit: Per the commit message:

> ### Revamp classpath search to use normalized paths
>
> On Windows with Java 11, the `java.class.path` property contains backslash-escaped backslashes, as in "`C:\\foo\\bar`".  These doubled backslashes interfere with the logic for finding classpath entries that contain the requested element.  Using normalized paths for all comparisons seems to fix the problem, and will likely be more robust in general.

That being said, this change should also work fine under Java 1.8. Therefore I will pull it out into a separate, small pull request that we can merge before the big Java 11 change.

liblit:
Reminder: we don't intend to merge these changes to `master` until after we have released WALA 1.5.9 as the last WALA release that will support Java 1.8.

msridhar (commented):
This now LGTM!  I'm going to hold off on proper "approval" until after we cut the 1.5.9 release

msridhar (approved):
WALA 1.5.9 has been released so this can now land!

liblit:
Thanks, @msridhar!

PR #1163, opened 2022-12-12
Remove Maven build

liblit:
It's time.


msridhar:
Are there now projects with code that is never tested, since before it was only tested under Maven?  E.g., I don't think the JSDT code will be tested at all by regression tests now.  I would like to delete code that is completely untested (it will still be available in the Git history).  WDYT @liblit?  We could also do it in a follow-up.

liblit:
I'm fine with removing JSDT or anything else that is now untested. It would be nice to identify that code in a systematic manner. Perhaps I can rig up some coverage tool to help us find whole subprojects that no tests touch.

msridhar:
Upon further thought, let's be a bit careful about deleting code.  I looked through some projects and it's not completely obvious what can be removed (e.g., `com.ibm.wala.ide` has some code that I'd like to keep, even if it's not currently getting tested in CI).  So we can go ahead with just this change right now that deletes the Maven build.

msridhar (approved):

msridhar (changes requested):
Actually, @liblit can you check that the `change-version.sh` script still works after this change?

liblit:
`change-version.sh` script still works after this change, but was pointlessly searching for `pom.xml` files that no longer exist. I've updated this pull request to remove that obsolete part of this script's logic.

msridhar (approved):
🎉

msridhar:
@liblit I pushed one more change removing a `mvn` command from `change-version.sh`; hope that looks reasonable to you.

You may need to update the required CI builds to remove "Maven Eclipse Tests" before this can land.

PR #1167, opened 2022-12-19
Remove -Wall compiler flag

khatchad:
I had to make this change to get the builship import to work.


khatchad:
Now I'm seeing in the buld a bunch of warnings pertaining to test code, which probably isn't what we want.

khatchad:
My feeling is that buidhship isn't properly working here, and the only viable way to import WALA into Eclipse is through using `./gradlew eclipse`.

khatchad:
And even `./gradlew eclipse` doesn't completely work. It still needs some finagling to get it building.

msridhar:
Thanks for all the digging @khatchad.  If you have time to update our docs on what is necessary here that would be much appreciated.

PR #1174, opened 2022-12-23
Port Gradle build system from Groovy to Kotlin

liblit:


msridhar (approved):
Approving after a fairly superficial review. Since all the file names changed, GitHub does not really show me a proper "diff" on each file, so it's hard to see if anything non-trivial happened in the translation from Groovy to Kotlin.  @liblit if you think there is anything subtle here that is worthy of a closer look, let me know.  Otherwise I am fine to land this and deal with any issues as they arise :-)

msridhar:
Whoops, didn't see that auto-merge was enabled, but @liblit the offer still stands if you think there was something worth a particular look here

liblit:
Thanks, @msridhar. I'm OK with merging after your cursory review. There was nothing in particular that I wanted closer scrutiny of.

PR #1175, opened 2022-12-23
Remove redundant modifiers

liblit:


msridhar (approved):
How did you detect these? Any way to prevent backsliding? Maybe there's an Error Prone check for it

liblit:
> How did you detect these?

InteliJ IDEA has a Java inspection that detects and optionally repairs all of these.

> Any way to prevent backsliding? Maybe there's an Error Prone check for it

There’s no Eclipse setting to detect these, so presumably we cannot configure our ECJ compilations to flag them. I’m also not seeing an equivalent Error Prone check [here](https://errorprone.info/bugpatterns), but perhaps I missed something?

[Checkstyle’s `RedundantModifier` check](https://checkstyle.sourceforge.io/config_modifier.html#RedundantModifier) seems equivalent, but would require adding a whole new linter to WALA’s builds.

PR #1182, opened 2022-12-24
Use String.repeat() where applicable

liblit:


com.ibm.wala.core/src/main/java/com/ibm/wala/core/util/strings/StringStuff.java
```diff
@@ -50,11 +50,7 @@ public static void padWithSpaces(StringBuilder b, int length) {
     if (b == null) {
       throw new IllegalArgumentException("b is null");
     }
-    if (b.length() < length) {
-      for (int i = b.length(); i < length; i++) {
-        b.append(' ');
-      }
-    }
+    b.append(" ".repeat(Math.max(0, length - b.length())));
```
msridhar, line 53: To me, it would be more readable to keep `if (b.length() < length) {` and put the `append()` call inside that condition than to use `Math.max()`

com.ibm.wala.util/src/main/java/com/ibm/wala/util/tables/Table.java
```diff
@@ -159,10 +159,6 @@ public static void padWithSpaces(StringBuilder b, int length) {
     if (b == null) {
       throw new IllegalArgumentException("b is null");
     }
-    if (b.length() < length) {
-      for (int i = b.length(); i < length; i++) {
-        b.append(' ');
-      }
-    }
+    b.append(" ".repeat(Math.max(0, length - b.length())));
```
msridhar, line 162: same as comment above

msridhar (dismissed):

com.ibm.wala.core/src/main/java/com/ibm/wala/core/util/strings/StringStuff.java
liblit: Sure, will do.

com.ibm.wala.util/src/main/java/com/ibm/wala/util/tables/Table.java
liblit: Sure, will do.

msridhar (approved):

PR #1185, opened 2022-12-30
Enlarge Gradle daemon's metaspace size limit

liblit:
I was seeing out-of-memory errors during concurrent Java compilation, specifically reporting depletion of the metaspace.  That space defaults to 256 MB, so let's enlarge it by 50% and see if the errors go away.


msridhar (approved):

liblit:
[Possibly related.](https://youtrack.jetbrains.com/issue/KT-56093)

PR #1190, opened 2022-12-30
Gather dependency versions in a version catalog

### LINKED ISSUE #1183: use Gradle version catalog
WALA's Gradle build system should [use a version catalog to coordinate dependency versions between subprojects](https://docs.gradle.org/current/userguide/platforms.html). A catalog will make it easier to update dependencies to newer versions consistently throughout WALA's code base.

[Dependabot does not yet support version catalogs](https://github.com/dependabot/dependabot-core/issues/3121), but [we stopped using Dependabot some time ago](https://github.com/wala/WALA/commit/505ada7914f87cbfeb0394e7c42db8fbbd72855f). Instead we use [the Gradle Versions Plugin](https://github.com/ben-manes/gradle-versions-plugin) (added in #654), which does support version catalogs. Other Gradle plugins could be added in the future to automate updates further, such as [this one](https://github.com/littlerobots/version-catalog-update-plugin).

liblit:
Resolves #1183.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -7,21 +7,22 @@ import com.github.sherter.googlejavaformatgradleplugin.VerifyGoogleJavaFormat
 import com.ncorti.ktfmt.gradle.tasks.KtfmtCheckTask
 import com.ncorti.ktfmt.gradle.tasks.KtfmtFormatTask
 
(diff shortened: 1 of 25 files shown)


msridhar (approved):
Really nice cleanup, thanks!

liblit:
> Really nice cleanup, thanks!

You're welcome! I'm glad you like it.

PR #1192, opened 2022-12-31
Update to Error Prone 2.17.0

msridhar:
Just to keep up

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -11,7 +11,7 @@ eclipse-ecj = "org.eclipse.jdt:ecj:3.21.0"
 eclipse-osgi = "org.eclipse.platform:org.eclipse.osgi:3.17.0"
 eclipse-wst-jsdt-core = { module = "org.eclipse.wst.jsdt:core", version.ref = "eclipse-wst-jsdt" }
 eclipse-wst-jsdt-ui = { module = "org.eclipse.wst.jsdt:ui", version.ref = "eclipse-wst-jsdt" }
-errorprone-core = "com.google.errorprone:error_prone_core:2.15.0"
+errorprone-core = "com.google.errorprone:error_prone_core:2.17.0"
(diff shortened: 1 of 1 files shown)


liblit (approved):

msridhar:
This is a bit weird:

https://github.com/wala/WALA/actions/runs/3809918658/jobs/6481638942#step:5:500

Somehow, for the CI job for this change, we get a build cache hit for `:com.ibm.wala.core:test`, even though `:com.ibm.wala.core:compileJava` and `:com.ibm.wala.core:testClasses` had to re-build.  @liblit does this look suspicious to you?

msridhar:
It's possible there is some special logic in the Gradle ErrorProne plugin that tells Gradle that even though the code needs to be re-compiled, the tests don't need to run as there is no semantic change?  I can look into that.  Want to understand before landing this.

liblit:
> It's possible there is some special logic in the Gradle ErrorProne plugin that tells Gradle that even though the code needs to be re-compiled, the tests don't need to run as there is no semantic change?

Gradle's caching logic is fairly fancy, with plenty of opportunities to leverage alternate (e.g., semantic) definitions for equivalence. See [Better reuse via input normalization](https://docs.gradle.org/current/userguide/build_cache_concepts.html#normalization). I am inclined to trust that Gradle did the right thing here.

msridhar:
> Gradle's caching logic is fairly fancy, with plenty of opportunities to leverage alternate (e.g., semantic) definitions of equivalence. See [Better reuse via input normalization](https://docs.gradle.org/current/userguide/build_cache_concepts.html#normalization). I am inclined to trust that Gradle did the right thing here.

Oh, I get it!  The thing here is that even though the compile tasks re-ran, the output _bytecodes_ were identical.  So, no need to re-run the tests, as the inputs are the same!  Very clever.  I am satisfied 🙂

liblit:
We've come a long way since the days of the Maven build, where the only trustworthy build was a clean rebuild, eh? :brain:

PR #1195, opened 2022-12-31
Move core.java11 into core

liblit:
All of WALA now requires Java 11+, so there's no longer any reason to separate Java 11 material from the rest of WALA.

Note that this pull request also *removes* several files that seem to be unused:

- `com.ibm.wala.core.java11/data/fact.class`
- `com.ibm.wala.core.java11/data/fact.java`
- `com.ibm.wala.core.java11/data/wala.testdata.txt`
- `com.ibm.wala.core.java11/src/main/java/com/ibm/wala/core/java11/LibraryStuff.class`
- `com.ibm.wala.core.java11/src/module-info.java`

My claim that these are unused is supported by the fact that all automated tests still pass after removing them. If I'm mistaken, and some of these files are actually needed, please let me know!


com.ibm.wala.core.java11/src/module-info.java
```diff
@@ -1,5 +0,0 @@
-module com.ibm.wala.core.java11 {
```
msridhar, line 1: We can delete this, but I'd like to think about the advantages / disadvantages of adopting modules now that we require JDK 11.  Not for this PR

msridhar (approved):
LGTM!

PR #1196, opened 2023-01-01
Remove obsolete Travis CI material

liblit:
We stopped using Travis CI nearly two years ago: #853 on January 2, 2021.  However, we still had a few leftover remnants of Travis-specific logic.  Let's get rid of that obsolete material too.

Note that we previously had a few Rhino tests that we pretended to skip only on Travis, but which we were actually skipping *everywhere* due to setting `$TRAVIS` in the test environment.  Those tests are now re-enabled, except for one which times out.  I'm not sure whether that timeout is correct or a regression.  And in general, I'm not sure whether we want to keep and re-enable these tests, keep them disabled, or remove them entirely.


com.ibm.wala.cast.js.rhino/src/test/java/com/ibm/wala/cast/js/rhino/callgraph/fieldbased/test/FieldBasedCGGamesTest.java
```diff
@@ -13,79 +13,61 @@ public class FieldBasedCGGamesTest extends AbstractFieldBasedTest {
 
   @Test
   public void testBunnyHunt() throws IOException, WalaException, Error, CancelException {
-    runTestExceptOnTravis(
-        new URL("http://www.themaninblue.com/experiment/BunnyHunt/"),
-        new Object[][] {},
-        BuilderType.OPTIMISTIC);
+    URL url = new URL("http://www.themaninblue.com/experiment/BunnyHunt/");
```
msridhar, line 16: I would prefer to have all these tests that load the input from a URL to _not_ run on CI; it's just a recipe for flakiness.  Can we somehow keep them disabled on CI?

liblit: Agreed. Implemented in cac148ea66f9876b1021a9784f078a6b9314bb41.

msridhar (approved):
Thanks for this cleanup!

liblit:
You're welcome!

PR #1200, opened 2023-01-03
Update to newer version of NullAway

liblit:


com.ibm.wala.util/src/main/java/com/ibm/wala/util/processes/Launcher.java
```diff
@@ -32,11 +32,11 @@ public abstract class Launcher {
 
   @Nullable protected Map<String, String> env = null;
 
-  @Nullable protected byte[] stdOut = null;
+  protected byte @Nullable [] stdOut = null;
```
msridhar, line 35: We actually didn't intend to force this change in the latest NullAway release; we're figuring out what to do about it in https://github.com/uber/NullAway/pull/706.  I'm moderately confident that this change will continue to work in the next NullAway release, but until we finalize (probably tomorrow), let's hold off, just in case I missed something.  I agree it's the right fix, though.

msridhar: I updated the PR to use the new release 0.10.7 that fixes this.  Still, these annotations are correct and should work, so I will leave this change.

msridhar (approved):

PR #1207, opened 2023-01-15
Publish test fixtures everywhere and simplify signature configuration

### LINKED ISSUE #1201: Publish artifacts for test fixtures
Artifacts relating to test fixtures are not currently published to the remote Sonatype repository. We should publish these too, though. See further discussion [here](https://github.com/wala/WALA/issues/1173#issuecomment-1374881868_).

liblit:
## Publish test fixtures everywhere

Resolves #1201.

## Simplify signature configuration

Previously we had two nearly identical publications: `local` and `remote`.  `local` was unsigned, and was only intended for local installation, but we could not really enforce that very well. Conversely, `remote` was signed (for non-snapshots) and was only intended for remote installation, which we likewise could not enforce very well.

Now we just have a single `maven` publication.  We always attempt to sign this publication's artifacts.  However, signature misconfiguration is quietly allowed unless we are publishing a non-snapshot to a real Maven remote repository.  Near as I can tell, this is the idiomatic way to sign official remote releases but not sign local installations or snapshot releases.

**Note:** [this wiki documentation](https://github.com/wala/WALA/wiki/Creating-Releases-and-Snapshots#publishing-released-versions-locally) should be updated after this commit is merged into `master`.  In fact, that entire section can probably just be deleted:  after this change, the obvious `publishToMavenLocal` target works as expected.  No signatures are required when installing only into the local Maven repository, even for non-snapshot releases.

### DIFF
--- a/buildSrc/src/main/kotlin/com/ibm/wala/gradle/publishing.gradle.kts
+++ b/buildSrc/src/main/kotlin/com/ibm/wala/gradle/publishing.gradle.kts
@@ -14,160 +14,137 @@ val isSnapshot = "SNAPSHOT" in version as String
 
 val javaComponent = components["java"] as AdhocComponentWithVariants
 
-fun skipTestFixtures() {
-  listOf(
-          configurations["testFixturesApiElements"],
-          configurations["testFixturesRuntimeElements"],
-      )
-      .forEach { javaComponent.withVariantsFromConfiguration(it) { skip() } }
-}
+val allTestFixturesSource = the<SourceSetContainer>()["testFixtures"].allSource
 
-fun MavenPublication.sharedPublicationConfiguration() {
-  from(javaComponent)
-  pom {
-    name.set(project.properties["POM_NAME"] as String)
-    description.set("T. J. Watson Libraries for Analysis")
-    inceptionYear.set("2006")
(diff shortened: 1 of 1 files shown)


msridhar (approved):
LGTM!  Some of the test fixture stuff can only be fully tested when we cut the next release.  Let's at least take a look at the snapshot artifacts after this lands and make sure the test fixtures artifacts are showing up there

liblit:
Agreed!

msridhar:
> **Note:** [this wiki documentation](https://github.com/wala/WALA/wiki/Creating-Releases-and-Snapshots#publishing-released-versions-locally) should be updated after this commit is merged into `master`. In fact, that entire section can probably just be deleted: after this change, the obvious `publishToMavenLocal` target works as expected. No signatures are required when installing only into the local Maven repository, even for non-snapshot releases.

For now, I have added a note that the paragraph only applies to releases 1.5.9 and earlier.  Once we cut our next release, I will further clarify.

liblit:
> For now, I have added a note that the paragraph only applies to releases 1.5.9 and earlier. Once we cut our next release, I will further clarify.

Excellent! Thank you for keeping this documentation in such good shape, @msridhar.

Issue #1206, opened 2023-01-15
Use Gradle configuration cache


liblit:
WALA's Gradle build system is complex, to put it mildly. We should try enabling [Gradle's configuration cache](https://docs.gradle.org/current/userguide/configuration_cache.html). If we can get that working, incremental builds should be significantly faster.

liblit:
Assigning to me. I'll take a first swing at this, see how far I get, and document both progress and blockers here.

liblit:
The [`gradle-configuration-cache`](https://github.com/wala/WALA/tree/gradle-configuration-cache) branch enables the Gradle configuration cache. [Builds on this branch currently fail](https://github.com/wala/WALA/actions/runs/3933570033), because we still have some work to do. We already handled problems with the Google Java format plugin in #1209, so that's progress. Errors reported on [the current `HEAD` of the `gradle-configuration-cache` branch](https://github.com/wala/WALA/commit/6d2b0d4985fd550dd5a36b65ae61d695a9bc5415) include:

> 63 problems were found storing the configuration cache, 9 of which seem unique.
> - Task `:com.ibm.wala.cast:cast:compileDebugCpp` of type `org.gradle.language.cpp.tasks.CppCompile`: cannot serialize object of type 'java.io.FileDescriptor', a subtype of 'java.io.FileDescriptor', as these are not supported with the configuration cache.
>   See https://docs.gradle.org/7.6/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
> - Task `:com.ibm.wala.cast:cast:compileDebugCpp` of type `org.gradle.language.cpp.tasks.CppCompile`: cannot serialize object of type 'java.io.PrintStream', a subtype of 'java.io.OutputStream', as these are not supported with the configuration cache.
>   See https://docs.gradle.org/7.6/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
> - Task `:com.ibm.wala.cast:cast:compileDebugCpp` of type `org.gradle.language.cpp.tasks.CppCompile`: cannot serialize object of type 'java.io.RandomAccessFile', a subtype of 'java.io.RandomAccessFile', as these are not supported with the configuration cache.
>   See https://docs.gradle.org/7.6/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
[...]

PR #1209, opened 2023-01-15
Use Spotless for Java and Kotlin formatting

liblit:
Spotless integrates nicely with the Java- and Kotlin-specific formatters we were already using, and also with Gradle.  Furthermore, it *appears* that Spotless may be ready for use with the Gradle configuration cache, which isn't the case for the Google Java formatter plugin we were using previously.  Thus, this change should move us a bit closer to #1206.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,10 +21,11 @@ Beyond tests, other checks run as part of `./gradlew check` and `./gradlew build
 
 1. Compilation with the Java compiler from [Eclipse JDT Core](https://www.eclipse.org/jdt/core/),
 which runs additional lint checks
-2. Checking that all code is formatted according to [Google Java
-  Format](https://github.com/google/google-java-format) standards
+2. Checking that all Java and Kotlin code is formatted according to [Google Java
+   Format](https://github.com/google/google-java-format) and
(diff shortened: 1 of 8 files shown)


msridhar:
Does the spotless plugin automatically add code formatting checks as a verification task, executed on a `./gradlew check`?  If not, how does one run the spotless checks?

msridhar:
> Does the spotless plugin automatically add code formatting checks as a verification task, executed on a `./gradlew check`? If not, how does one run the spotless checks?

Nevermind, based on the initial example [here](https://github.com/diffplug/spotless/tree/main/plugin-gradle) it's clear that if there are formatting violations, `./gradlew build` will fail.

msridhar (commented):
Changes LGTM, but can you also update `CONTRIBUTING.MD` to refer to the relevant spotless tasks instead of the tasks from the old Java formatting plugin?

liblit:
Ah yes, good catch. Thank you, once again, for being such a good guardian of WALA's documentation.

msridhar:
> Ah yes, good catch. Thank you, once again, for being such a good guardian of WALA's documentation.

Haha, thanks, the docs are too few, but I try to keep up the ones that we have 🙂

msridhar (approved):

PR #1210, opened 2023-01-17
Use third-party Gradle plugin for ShellCheck

liblit:
Also, require that ShellCheck pass for all macOS and Linux builds, not just builds where a `shellcheck` executable can be found.  If you want to build WALA, you need to install ShellCheck.  Fortunately, [ShellCheck is quite easy to
install](https://github.com/koalaman/shellcheck#installing) on all supported platforms.

Unfortunately, this ShellCheck plugin [does not work reliably on Windows](https://github.com/felipefzdz/gradle-shellcheck-plugin/issues/7). Until that bug is fixed, it's simpler to just skip ShellCheck on Windows.  We still run ShellCheck on macOS and Linux, so there's no risk of a bad shell script slipping into `master`.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -22,9 +22,6 @@ jobs:
       fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
-      - name: Install ShellCheck on macOS
-        run: brew install shellcheck
-        if: runner.os == 'macOS'
       - name: Check out WALA sources
         uses: actions/checkout@v3
       - name: Cache Goomph
@@ -42,7 +39,7 @@ jobs:
         uses: gradle/gradle-build-action@v2
         with:
           gradle-executable: xvfb-gradle.sh
-          arguments: aggregatedJavadocs build publishAllPublicationsToFakeRemoteRepository
+          arguments: aggregatedJavadocs build publishAllPublicationsToFakeRemoteRepository shellcheck
         # testing ECJ compilation on any one OS is sufficient; we choose Linux arbitrarily
         if: runner.os == 'Linux'
       - name: Build and test using Gradle but without ECJ
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -22,6 +22,9 @@ jobs:
       fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
+      - name: Install ShellCheck on macOS
+        run: brew install shellcheck
+        if: runner.os == 'macOS'
       - name: Check out WALA sources
         uses: actions/checkout@v3
       - name: Cache Goomph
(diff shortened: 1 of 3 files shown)


liblit:
It's worth mentioning that this new approach to ShellCheck is compatible with the Gradle configuration cache. Thus, this change helps us make incremental progress toward #1206.

msridhar:
Question: how much does using the 3rd-party shellcheck plugin buy us over implementing this ourselves?  I am not sure how maintained that plugin is (no recent commits it seems).  It's possible it's just "done", but I do worry slightly that in the future we may have issues if Gradle changes lead to incompatibilities.  If it's providing a significant amount of functionality, then it's worthwhile; just checking.

liblit:
> Question: how much does using the 3rd-party shellcheck plugin buy us over implementing this ourselves?

Well, it makes our build logic smaller (+19 lines −35 lines). And it's compatible with the Gradle configuration cache, which our previous approach was not. Presumably we could have figured out what we needed to change for the configuration cache, but this third party already has. Other than that, it's not bringing any major new functionality.

If we need to go back to doing this ourselves, I don't think it will be hard to recover or recreate the logic we had before.

liblit:
I guess I should mention that this third-party plug-in does provide a significant piece of functionality that we are not currently using, but that I would not want to create from scratch myself. It has the ability to install and run ShellCheck from a Docker image. This feature allows for more hermetic, repeatable builds: no unmanaged dependency on what version of ShellCheck happens to be sitting around on the build machine.

I opted not to use this for a WALA right now, because Docker is pretty heavyweight, and our builds are chunky enough already. But if we think that this feature is something we might want in the future, then this plugin gives it to us.

msridhar:
Ok, seems reasonable to use a plug in.  One more question; if a user tries to build without `shellcheck` being installed, can we provide them a reasonable error message indicating it must be in the path to complete the build?

It is mildly annoying to have another external requirement to build WALA besides an appropriate JDK.  Was there a reason we had to drop the behavior of skipping the check if `shellcheck` is not present?

msridhar:
Thinking more here, I think my preference here is that the `shellcheck` task _not_ run by default when running `./gradlew build`; instead it should be separately invoked.  Our CI jobs can run `shellcheck`, as we can ensure `shellcheck` is installed.  I imagine that nearly all users building WALA from source are concerned with modifying Java source code, not our shell scripts.  So I think not running `shellcheck` by default is reasonable.  I know that one can run tasks like `assemble`, `classes`, `test`, etc. and the verification tasks will not run, but I think many users will be used to just running `./gradlew build` and it's best to keep that experience smooth.

msridhar:
@liblit is this ready for review again?

liblit:
> @liblit is this ready for review again?

Yep. Thanks!

msridhar (approved):
LGTM!

PR #1221, opened 2023-01-24
Replace `buildSrc` with an included build

liblit:
Gradle automatically looks for a `buildSrc` subdirectory containing build logic, such as custom plugins.  If found, Gradle builds this supporting project before building the main project.  In many ways, `buildSrc` acts like an included build: a self-standing project that is incorporated into some larger project, turning the latter into a so-called "composite build".

However, `buildSrc` is not *exactly* an included build, and from what I am seeing, the differences between the two make a real included build preferable to `buildSrc`:

* [Stop using Gradle `buildSrc`. Use composite builds instead](https://proandroiddev.com/stop-using-gradle-buildsrc-use-composite-builds-instead-3c38ac7a2ab3)

* [How to use Composite builds as a replacement of buildSrc in Gradle](https://medium.com/bumble-tech/how-to-use-composite-builds-as-a-replacement-of-buildsrc-in-gradle-64ff99344b58)

In particular, apparently any change to `buildSrc` invalidates the entire build cache for the main project.  When using a composite build instead, cache invalidation can be more selective.

Therefore, this commit replaces `buildSrc` with an included build. Following conventions I've seen elsewhere, the included build is called `build-logic`.  Fortunately, this is a *very* easy change.  We just needed to rename the `buildSrc` directory to `build-logic`, and add a single `includeBuild("build-logic")` directive to the main `settings.gradle.kts`.  That's it!


msridhar (approved):
Nice find!

PR #1227, opened 2023-02-04
Update gradle-download-task plugin

liblit:
Also use some new features added to this plugin's latest release that let us avoid referring to the current `Project` during the action method of `VerifiedDownload` instances.  This change makes our `VerifiedDownload` tasks compatible with the Gradle configuration cache.

Unfortunately we also need to change how we set the inputs for a few tasks that postprocess downloads.  It's not clear to me whether this is due to a problem in our `VerifiedDownload` task, in the `gradle-download-task` plugin, or in Gradle itself.  I can live with this change for now, but it would be nice to understand it better.


msridhar (approved):
👍

liblit:
Big thanks to `gradle-download-extension` maintainer @michael-kraemer for being receptive to our requests and making the configuration-cache changes that we needed here: michel-kraemer/gradle-download-task#284.

PR #1236, opened 2023-02-12
Remove disused Eclipse material

liblit:
These files all relate to building Eclipse plugins, but we stopped doing that a while ago.  Nobody seems to know whether any of these files are actually useful for anything at this point.


msridhar (approved):
We can always get these files back from git history if needed

PR #1237, opened 2023-02-13
Remove obsolete version-changing script

liblit:
We now store the version number in one place, and one place only: the `VERSION_NAME` property in the top-level `gradle.properties` file.  Any other use of the version number should be derived from this setting at build time.  Thus, we no longer need a helper script to remember other places where the version number is embedded in source code:  there are no such other places.

I thought we had some documentation on creating new WALA releases.  That documentation will probably need to be updated.  I can't find it though. Where should I be looking?


msridhar (approved):
@liblit this is the doc that needs to be update: https://github.com/wala/WALA/wiki/Creating-Releases-and-Snapshots

liblit:
[Wiki documentation updated](https://github.com/wala/WALA/wiki/Creating-Releases-and-Snapshots/_compare/ae13f2a5c98f8942234b2b0ea023ec02de5a4870...a8df9a9cf01caa8fcc6d820d3f9229c93b637a58?short_path=b63b7e1#diff-b63b7e112443adf237ff90054a747405bc1840e0a196397c03b5873fb3274661). Thanks for the link, @msridhar!

PR #1242, opened 2023-02-19
Upgrade to Gradle 8.0.1

liblit:
Gradle 8.0 includes Kotlin 1.8.0, and Kotlin 1.8.0 deprecates `String.toLowerCase()`.  So in the `NullAway` script plugin, instead of using `name.toLowerCase().contains("test")`, we now use `name.contains("test", true)`.  The `true` argument requests case-insensitive `String` comparison, which perfectly corresponds to what the earlier code was doing anyway.

Gradle 8.0 adds an annoying extra check for `JavaCompile` tasks.  If such a task is configured to fork a subprocess, Gradle 8.0 now requires that the configured subprocess executable match the configured toolchain compiler executable.  That's a problem for our ECJ Java compile tasks: they want to fork a `java` subprocess, which does not match `javac` as the toolchain compiler executable.  Working around this requires that we adapt our `JavaCompileUsingEcj` task implementation *yet again*, as we seemingly have to do with every new Gradle release, with gradle/gradle#15942 still not implemented.  This time around, we avoid relying on the standard `JavaCompile.compile()` task action method, since calling that leads to the new executable equality check.  Instead, we provide our own `compile()` task action method that directly runs the ECJ subprocess with a suitable command line.

Along with that change, we also slightly change how we define and resolve the ECJ jar.  Previously we used a detached `Configuration` for this purpose, but that turns out to be a bit tricky to manage with the Gradle configuration cache.  Now we use a normal, named `Configuration` that is defined and configured in the shared `java.gradle.kts` script plugin.

We also move the `skipJavaUsingEcjTasks` property check out of the `JavaCompileUsingEcj` task constructor, because having it there produced an IntelliJ IDEA warning about calling a non-`final` method in a constructor.  Instead, we configure this check in an existing bit of `java.gradle.kts` logic that was already being used to configure all `JavaCompileUsingEcj` tasks.


msridhar (approved):
Thanks for this!  LGTM

liblit:
You're welcome, @msridhar!

PR #1246, opened 2023-02-26
Aggregate Javadoc only from Java subprojects

liblit:
Previously we built the whole-WALA Javadoc HTML pages by aggregating Javadoc from all subprojects. That, in turn, required creating a `javadocClasspath` and `javadocSource` `Configuration` in each subproject, including a few subproejcts with no Java code at all.

Now we aggregate only from subprojects that have loaded, or will load, the `java-base` plugin.  This change nicely restricts aggregation to only those subprojects that might actually contain Javadoc documentation.


msridhar (approved):
If you could eyeball the generated Javadoc to check that it seems to cover the same classes that would be great

liblit:
> If you could eyeball the generated Javadoc to check that it seems to cover the same classes that would be great

I did, and it does. 👍

PR #1258, opened 2023-03-24
Fix link to site-specific search

khatchad:
Looks like Google isn't indexing GitHub wiki pages (maybe there's a `robots.txt?`). Switch to GitHub native wiki search.


msridhar (approved):
Thanks a lot!

PR #1261, opened 2023-04-10
Fix NPE in EclipseSourceFileModule ctor

khatchad:
The `getLocation()` API is returning `null`, resulting in an NPE. Use `getFullPath()` instead, which seems to work.


msridhar (approved):
Would normally ask for a test for this, but don't think that will be easy, so we can go ahead and land

PR #1286, opened 2023-07-05
Add ability to only analyze a subset of standard lib from running JVM

msridhar:
Before, when performing Java bytecode analysis using the standard library from the running JVM, we would always include _all_ the standard library jar / jmod files.  Now that WALA only runs on JDK 11+, which has a modularized standard library, it makes sense to support strategies where only a subset of the modules are analyzed.

This PR adds support for such alternate strategies as follows.  First, we add a `base` option for the `stdlib` line in scope files, such that when used, only the `java.base` module is added to the `AnalysisScope`.  Additional JDK modules can be added via the scope file via a new `jdkModule` line and the module name.  We also add corresponding APIs, `AnalysisScopeReader.makeBasePrimordialScope` and `AnalysisScope.addJDKModuleToScope`.  By default, WALA client behavior remains the same; clients must make use of these new features / APIs to opt into new behavior.

The `core` module regression tests have been rewritten to minimize the number of modules included for the analyses; most tests can pass when just including `java.base`.  Surprisingly, the tests still take significantly longer to run on JDK 17 than on JDK 11; we should investigate why in the future.

Some code only relevant when running on pre-JDK-11 JVMs has also been cleaned up / removed.


util/src/main/java/com/ibm/wala/util/PlatformUtil.java
```diff
@@ -47,38 +46,44 @@ public static boolean onIKVM() {
...
+   *     module
+   * @return array of {@code .jmod} module files
+   * @throws IllegalStateException if modules cannot be found
    */
-  public static String[] getBootClassPathJars() {
-    String classpath = null;
-    String javaVersion = System.getProperty("java.specification.version");
-    if (!javaVersion.equals("1.8")) {
-      // java11 support for jmod files
+  public static String[] getJDKModules(boolean justBase) {
```
liblit, line 56: I'm all about that bass, 'bout that bass, no treble
I'm all about that bass, 'bout that bass, no treble
I'm all about that bass, 'bout that bass (bass, bass, bass, bass)

🎶

liblit (approved):

liblit:
Nice work, @msridhar!

msridhar:
Thanks @liblit!  At least for some clients this should give a nice perf win

PR #1289, opened 2023-07-23
Use JUnit 5 to find and run tests

liblit:
The tests themselves are still written using JUnit 4, so we're using JUnit 5's Vintage component to discover and run those tests.  Hooray for backward compatibility!

Eventually we can try porting JUnit 4 tests forward to JUnit 5, but this is a decent first step.


msridhar (approved):
👍

PR #1293, opened 2023-07-23
Migrate many tests from JUnit 4 to JUnit 5

liblit:
Most of these migrations were mechanized using IntelliJ IDEA followed by some manual cleanups.  A few were migrated entirely manually, such as the expected-to-fail tests in `FieldBasedComparisonTest`.


liblit:
Cool, we have [the test summary](https://github.com/wala/WALA/pull/1293#issuecomment-1650830377) that #1297 added! And it's doing exactly what I wanted, which is to show that this pull request causes no net change in the number of tests being run.

msridhar (approved):
Obviously didn't review every line of this one, but did some spot checking and it LGTM!

liblit:
Thanks, @msridhar. This PR was for the big bulk of migrations that were mostly mechanized. I separated these migrations out specifically so that you could just skim, which is exactly what you did. Several other migrations will follow, all _vastly_ smaller, for tests that required manual conversion. Those are more worth your careful attention than this bulk batch. :+1:

PR #1294, opened 2023-07-25
Avoid printing many call graphs during testing

liblit:
These graph dumps add hundreds of megabytes to test logs, and presumably are manually inspected approximately never.  Let's turn them off. Anyone chasing a specific bug or regression can add more debug output in their local working tree, as needed, but there's no benefit to dumping all of this data on every automated test run.

Furthermore, I'm exploring an option for getting a nice test summaries after each GitHub actions workflow run.  Unfortunately, the tool I am considering has some limits on the sizes `CDATA` sections it can read in JUnit's XML test logs.  When our test logs include call graph dumps, several of them exceed these limits.


msridhar (approved):
LGTM

PR #1297, opened 2023-07-26
Generate a summary test report for all builds

liblit:
My intent is for this summary report to increase our confidence that we're not accidentally dropping any tests or dramatically hurting test performance as we [migrate from JUnit 4 to JUnit 5](https://github.com/orgs/wala/projects/2).  The summary report may also turn out to be useful if anyone posts a pull request that causes new test failures, though that's rare from what I've observed.

Full disclosure:  I do not actually understand all the instructions on configuring the report generator to [Support fork repositories and dependabot branches](https://github.com/EnricoMi/publish-unit-test-result-action#support-fork-repositories-and-dependabot-branches) or for [Running with multiple event types (pull_request, push, schedule, …)](https://github.com/EnricoMi/publish-unit-test-result-action#support-fork-repositories-and-dependabot-branches). I'm probably not doing everything correctly here.  It seems to be working just fine, though, even in my forked repository.  So let's just charge ahead and see what fails, if anything.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -55,6 +55,31 @@ jobs:
         run: ./check-git-cleanliness.sh
         # not running in Borne or POSIX shell on Windows
         if: runner.os != 'Windows'
+      - name: Upload test results
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: Test results for JDK ${{ matrix.java }} on ${{ runner.os }}
+          path: '**/build/test-results/test/TEST-*.xml'
+  publish_test_results:
+    name: Repoort test results
+    needs: build_gradle
+    if: always()
+    runs-on: ubuntu-latest
+    permissions:
+      checks: write
+      pull-requests: write
+      contents: read
+      issues: read
+    steps:
+      - name: Download artifacts
+        uses: actions/download-artifact@v3
+        with:
+          path: artifacts
+      - name: Publish test results
+        uses: EnricoMi/publish-unit-test-result-action@v2.8.0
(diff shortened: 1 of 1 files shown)


msridhar:
This looks really cool!  Where will we be able to see the reports?

liblit:
Eventually the test report should look [something like this](https://github.com/liblit/WALA/actions/runs/5662887967/attempts/1#summary-15344212486), and should also appear directly on the pull request page. I believe that it's not appearing there now because [the workflow that creates it is itself new to this pull request](https://github.com/wala/WALA/pull/1297/files#diff-208a7f7ef0a1174d68708f24502994bd2a136c75128806eb0cf63c0a70679321). Once this pull request is merged into `master`, then I _think_ we'll start seeing those reports for future pull requests, somewhere on the pull request's own page. :crossed_fingers:

liblit:
The report is generated using a workflow that is [triggered by a `workflow_run` event](https://github.com/wala/WALA/pull/1297/files#diff-208a7f7ef0a1174d68708f24502994bd2a136c75128806eb0cf63c0a70679321R3-R7). According to [the `workflow_run` documentation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run):

> Note: This event will only trigger a workflow run if the workflow file is on the default branch.

This note aligns with my understanding that we will only start getting the test reports after _this_ pull request is merged into `master`.

msridhar (approved):

liblit:
[Test reporting is now working exactly as intended.](https://github.com/wala/WALA/pull/1293#issuecomment-1650832150) Yay!

PR #1298, opened 2023-07-26
Migrate parameterized DroidBench tests from JUnit 4 to JUnit 5

liblit:
IntelliJ IDEA does not know how to migrate parameterized tests, so these test classes were migrated by hand.

### DIFF
--- a/dalvik/build.gradle.kts
+++ b/dalvik/build.gradle.kts
@@ -93,6 +93,7 @@ dependencies {
 
   testImplementation(libs.android.tools)
   testImplementation(libs.dexlib2)
+  testImplementation(libs.junit.jupiter.params)
   testImplementation(projects.core)
   testImplementation(projects.dalvik)
   testImplementation(projects.shrike)
--- a/dalvik/src/test/java/com/ibm/wala/dalvik/test/callGraph/DroidBenchCGTest.java
+++ b/dalvik/src/test/java/com/ibm/wala/dalvik/test/callGraph/DroidBenchCGTest.java
@@ -11,7 +11,7 @@
 package com.ibm.wala.dalvik.test.callGraph;
 
 import static com.ibm.wala.dalvik.test.util.Util.walaProperties;
-import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import com.ibm.wala.classLoader.IClass;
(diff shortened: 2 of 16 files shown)


msridhar:
Do we expect the discrepancy between removed and added tests in the test results report?

liblit:
> Do we expect the discrepancy between removed and added tests in the test results report?

We do not. I'm still investigating, which is why I haven't requested your review yet. I'm glad to see that the new test summary report is serving as an extra check: exactly what I intended.

liblit:
In brief: we're not actually losing any tests in this PR. We're simply losing spurious variations in test _names_.

---

The test summary _before_ this change reports 2,804 test runs and 846 tests. The 846 count is based on distinct test names. A single test name leads to multiple test runs because we test on multiple operating systems and Java versions.

The test summary _after_ this change reports 2,804 test runs and 730 tests. So we actually have the same number of test runs across all tested platforms, but these test runs consist of fewer distinct test _names_.

A detailed, manual review of the before/after test names shows that several tests' before names included absolute path names, but are reported using shorter, relative path names after this change. Absolute paths vary by operating system, while the relative names do not. Thus, the before tests had 846 distinct names that reduced to just 730 distinct names after this change. To give one concrete example, tests using the `com.ibm.wala.dalvik.test.callGraph.droidbench.ThreadingTest` class include:

| revision | JDK | OS | Test Name |
|---|---|---|---|
| before | 11 | Linux | `runTest[DroidBench: /home/runner/work/WALA/WALA/dalvik/build/DroidBench/apk/Threading/Looper1.apk]` |
| before | 17 | Linux | `runTest[DroidBench: /home/runner/work/WALA/WALA/dalvik/build/DroidBench/apk/Threading/Looper1.apk]` |
| before | 11 | macOS | `runTest[DroidBench: /Users/runner/work/WALA/WALA/dalvik/build/DroidBench/apk/Threading/Looper1.apk]` |
| after | 11 | Linux | `[5] Threading/Looper1.apk` |
| after | 17 | Linux | `[5] Threading/Looper1.apk` |
| after | 11 | macOS | `[5] Threading/Looper1.apk` |

msridhar (approved):
Thanks for the change and for digging in to check it!

liblit:
You're welcome, @msridhar!  I'm really liking these "Test Results" summaries. We’re still getting used to them, but they’re going to be a nice piece of extra protection for us.

PR #1300, opened 2023-07-26
Migrate slow tests from JUnit 4 to JUnit 5

liblit:
These tests were all migrated by hand, including updating the Gradle logic for conditionally excluding them.


msridhar (approved):
👍

PR #1301, opened 2023-07-26
Port a few JUnit 4 odds and ends to JUnit 5

liblit:
These miscellaneous uses of JUnit 4 escaped my attention in other recent conversion passes.

### DIFF
--- a/cast/js/src/testFixtures/java/com/ibm/wala/cast/js/test/TestForInBodyExtraction.java
+++ b/cast/js/src/testFixtures/java/com/ibm/wala/cast/js/test/TestForInBodyExtraction.java
@@ -22,8 +22,8 @@ import com.ibm.wala.classLoader.SourceModule;
 import com.ibm.wala.util.io.FileUtil;
 import java.io.File;
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.util.regex.Pattern;
-import org.junit.ComparisonFailure;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
@@ -67,14 +67,9 @@ public abstract class TestForInBodyExtraction {
       expected = new CAstDumper().dump(parseJS(tmp, ast));
       expected = eraseGeneratedNames(expected);
 
-      assertEquals(expected, actual, testName);
+      assertEquals(expected, actual, () -> "Comparison Failure in " + testName + '!');
     } catch (IOException e) {
-      e.printStackTrace();
-    } catch (ComparisonFailure e) {
-      System.err.println("Comparison Failure in " + testName + "!");
-      System.err.println(expected);
-      System.err.println(actual);
-      throw e;
+      throw new UncheckedIOException(e);
     } finally {
       if (tmp != null && tmp.exists()) tmp.delete();
     }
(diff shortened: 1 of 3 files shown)


cast/js/src/testFixtures/java/com/ibm/wala/cast/js/test/TestForInBodyExtraction.java
```diff
@@ -70,7 +70,7 @@ public void testRewriter(String testName, String in, String out) {
       assertEquals(expected, actual, testName);
     } catch (IOException e) {
       e.printStackTrace();
-    } catch (ComparisonFailure e) {
+    } catch (AssertionFailedError e) {
```
msridhar, line 73: The pre-existing code is a bit strange.  Since we're editing here, can we clean it up by changing the `try` to a try-with-resources for the `tmp` variable?  I think then there will also be no need to catch the `IOException`; right now the test passes if there is an `IOException` which I think is a mistake.

msridhar: Also I think the `assertEquals` call should be changed to pass `"Comparison Failure in " + testName + "!"` as the last argument, and then all the `System.err.println` calls in the catch block can go away.

msridhar (commented):
One further cleanup suggestion

cast/js/src/testFixtures/java/com/ibm/wala/cast/js/test/TestForInBodyExtraction.java
liblit: If we turn this `try`/`finally` into `try`-with-resources, then the temporary `File` referenced by `tmp` will only be closed, not deleted. We should retain the `try`/`finally` structure used here for reliable and timely deletion.

---

There are multiple calls within the `try` body that could potentially throw `IOException`. If we remove the `catch IOException` clause, then this method must itself be declared as `throws IOException`. That change will in turn have cascading effects on all callers.

However, I agree that the test should fail if `IOException` arises. I will revise this code to catch `IOException` and rethrow it wrapped in a new `UncheckedIOException`.

---

I agree that catching `AssertionFailedError` to add extra output can be replaced with passing a more informative description to the `assertEquals` call itself. Furthermore, there's no good reason for that description to include the `expected` and `actual` values, since those will be included automatically by `assertEquals`. For example, this code:

```java
assertEquals("foo", "bar", "Mismatch!");
```

produces this output:

```plain
Mismatch! ==> expected: <foo> but was: <bar>
```

I'll apply this change to the next revision of this pull request.

msridhar: > If we turn this `try`/`finally` into `try`-with-resources, then the temporary `File` referenced by `tmp` will only be closed, not deleted. We should retain the `try`/`finally` structure used here for reliable and timely deletion.

Fair enough, though I think with the current code the `File` object does not get closed?  For deletion, a built-in solution seems to be the `@TempDir` annotation for JUnit 5:

https://www.baeldung.com/junit-5-temporary-directory#2-tempdir-on-an-instance-field

But we don't need to change this within this PR.

msridhar (approved):

PR #1304, opened 2023-07-31
Remove JUnit 4 dependency

liblit:
All tests and test infrastructure have been migrated or ported to JUnit 5. Nothing remains that uses JUnit 4.

### DIFF
--- a/build-logic/src/main/kotlin/com/ibm/wala/gradle/java.gradle.kts
+++ b/build-logic/src/main/kotlin/com/ibm/wala/gradle/java.gradle.kts
@@ -42,11 +42,9 @@ dependencies {
   "javadocSource"(sourceSets.main.get().allJava)
 
   testFixturesImplementation(platform(findLibrary("junit-bom")))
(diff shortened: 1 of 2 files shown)


msridhar:
Shall we double check that the test numbers match in the report before merging this one?  Or is it guaranteed safe as long as the test code compiles?

liblit:
No harm waiting to review the test summary. We need to wait until the "Continuous integration" runs finish anyway, and the report is produced immediately after that.

msridhar:
> No harm waiting to review the test summary. We need to wait until the "Continuous integration" runs finish anyway, and the report is produced immediately after that.

The PR was set to auto-merge, so I was wondering if with an approving review the change would get merged independent of the test report contents, as long as the tests passed (which may not be what we want).

liblit:
Good point. I've turned off auto-merge. You can review and approve more, or at your convenience. I'll postpone merging until I've looked over the test report.

msridhar (approved):

PR #1306, opened 2023-08-05
Let JUnit 5 manage tests' temporary files

liblit:
This change implements [a suggestion previously made by @msridhar](https://github.com/wala/WALA/pull/1301#discussion_r1275297725), not just in the class originally under review, but in several other test (and test fixture) classes as well.


msridhar (dismissed):
Thanks for this cleanup!

liblit:
@msridhar, [this is what changed since your original review](https://github.com/wala/WALA/pull/1306/commits/9bf2772e8d5a50c2ac454a99f7b9e022398295a9). I recommend hiding whitespace-only changes when looking at this commit. You can find that setting under the “⚙️ ▾” menu near the top of the page, immediately to the right of “Jump to ▾”.

msridhar (approved):

PR #1310, opened 2023-08-20
Upgrade to Gradle 8.3

liblit:
Beyond updating Gradle itself, this change required two groups of changes to our build logic:

1. `project.buildDir` is deprecated. The preferred replacement is `project.layout.buildDirectory`, although that has a slightly different type.

1. The configuration cache complains about using a `java.net.URL` as an `@Input` property, due to various oddities in that type's Java serialization behavior. However, `java.net.URI` is just fine. So our various download tasks now use `URI`s instead of `URL`s to identify the resource to be downloaded.


msridhar (approved):
Thanks!

PR #1317, opened 2023-09-04
Correct redundant escaping

liblit:
Whoever wrote these `String.replaceAll` calls was a bit confused about how backslashes behave in the replacement text.  They actually behave less like backslashes in generic `String` literals, and more like backslashes in regular expressions.  For example, to replace a newline with a backslash followed by an `n`, the replacement text should be `\\\\n`, not `\\n`.

I've added a test for
`com.ibm.wala.types.generics.MethodTypeSignature.getArguments()` to affirm that my changes there leave us with an implementation that has the intended effects.  I haven't added a
`com.ibm.wala.cast.js.html.DomLessSourceExtractor.HtmlCallback.quotify(String)` test, though, because this method is buried too deep inside `protected` APIs that are difficult for a unit test to get access to.  Instead, to argue that the new code matches the original `quotify` authors' likely intent, contrast these two JShell approximations:

* old code:

  ```plain
  jshell> System.out.println("foo\n\"bar".replaceAll("\"", "\\\"").replaceAll("\n", "\\n"))
  foon"bar
  ```

* new code:

  ```plain
  System.out.println("foo\n\"bar".replaceAll("\"", "\\\\\"").replaceAll("\n", "\\\\n"))
  foo\n\"bar
  ```

### DIFF
--- a/cast/js/src/main/java/com/ibm/wala/cast/js/html/DomLessSourceExtractor.java
+++ b/cast/js/src/main/java/com/ibm/wala/cast/js/html/DomLessSourceExtractor.java
@@ -199,11 +199,11 @@ public class DomLessSourceExtractor extends JSSourceExtractor {
         quote = '"';
       } else {
         quote = '"';
-        value = value.replaceAll("\"", "\\\"");
+        value = value.replaceAll("\"", "\\\\\"");
       }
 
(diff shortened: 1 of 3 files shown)


msridhar (approved):
Good catch!

PR #1319, opened 2023-09-05
Suppress warnings about unstable Gradle APIs

liblit:
We're intentionally using [an unstable Gradle Kotlin feature that allows direct `... = ...` assignment to `Property` instances](https://blog.gradle.org/simpler-kotlin-dsl-property-assignment). This feature may be considered unstable, but it's stable enough to have been [enabled by default since Gradle 8.2](https://docs.gradle.org/8.2.1/release-notes.html#simple-property-assignment-in-kotlin-dsl-enabled-by-default).


msridhar (approved):
A bit annoying we can't add this suppression globally and need to do it in each file

PR #1320, opened 2023-09-05
Avoid memory leak for thread-local values

liblit:
Setting a thread-local value to `null` is not the same as removing it. The former retains an entry mapping the current thread's value to `null` in a hidden global map.  The latter removes the current thread's entry entirely.


msridhar (approved):
Nice catch!

PR #1339, opened 2023-12-01
Support building on JDK 21, and don't crash on JDK 21 bytecodes

msridhar:
With these changes, WALA builds and runs on JDK 21, and our regression tests pass, except for the Dalvik tests.  The Dalvik test issue is tracked as #1349.  I think it's best to land these changes now, since they include important fixes for JDK 21 support.  We can then fix Dalvik in a follow-up.

The main logic fix was to update the native models for `java.lang.Thread` to account for implementation changes there.  Also there are a bunch of changes due to the deprecation of the `URL(String)` constructor.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -21,6 +21,8 @@ jobs:
             java: 11
           - os: ubuntu-latest
             java: 17
+          - os: ubuntu-latest
+            java: 21
       fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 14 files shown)


msridhar:
Once this PR is approved, I will add the JDK 21 / ubuntu CI job as required before landing it.

liblit (approved):
I've been silently reviewing this PR's progress with each successive commit. That puts me in a position to approve this pull request now with little delay. Trust me that I have indeed looked over everything you're changing here.

liblit:
Well done, @msridhar! 💪🧠

msridhar:
Thanks for the quick review, @liblit!  The `Thread` issue was a bit nasty to track down but the rest was thankfully straightforward.

PR #1344, opened 2023-12-09
Use structured KDoc comments

liblit:
Also conform to KDoc conventions: initial capital letter and trailing period for the first sentence.


msridhar (approved):
👍

Issue #1278, opened 2023-06-23
Configuration cache failures when compiling on JDK 17


msridhar:
I've started to tinker with basic JDK 17 compatibility for WALA.  As a start, on the master branch I ran `./gradlew compileJava compileTestJava`, and got the following configuration cache warnings:

```
5 problems were found storing the configuration cache, 2 of which seem unique.
- Plugin 'com.diffplug.configuration-cache-for-platform-specific-build': external process started 'uname -a'
  See https://docs.gradle.org/8.1/userguide/configuration_cache.html#config_cache:requirements:external_processes
- Task `:cast:java:test:data:compileTestJava` of type `org.gradle.api.tasks.compile.JavaCompile`: execution of task ':cast:java:test:data:compileTestJava' caused invocation of 'Task.extensions' in other task at execution time which is unsupported.
  See https://docs.gradle.org/8.1/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution
```

The first one looks like a plugin issue, but the second one might be in our own scripts.  FYI @liblit

liblit:
I am unable to reproduce these problems using

* a pristine WALA build tree,
* Fedora Linux 38,
* JDK 17, and
* the following command: `env JAVA_HOME=/usr/lib/jvm/java-17 ./gradlew compileTestJava`.

Gradle's output reports no configuration cache problems:

```
[...]
0 problems were found storing the configuration cache.

See the complete report at file:///home/liblit/Documents/research/joust/WALA/build/reports/configuration-cache/ac73fxhx2rqo7rrnei59eef83/2uk93bd2cpoe4qhaqrtefziw/configuration-cache-report.html

Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

See https://docs.gradle.org/8.1/userguide/command_line_interface.html#sec:command_line_warnings

BUILD SUCCESSFUL in 3m 23s
47 actionable tasks: 47 executed
Configuration cache entry stored.
```

Rerunning the same build command likewise shows no problems:

```
Encryption of the configuration cache is enabled.
Reusing configuration cache.

BUILD SUCCESSFUL in 1s
37 actionable tasks: 37 up-to-date
Configuration cache entry reused.
```

@msridhar, can you provide more details, perhaps about any aspects of your system that differ from mine?

msridhar:
[...]

Issue #1230, opened 2023-02-06
Configuration cache fails for the `compileTestJava` task


msridhar:
I tried running `./gradlew --configuration-cache compileTestJava`, which I expected to succeed.  Instead, I got a failure:
```
* What went wrong:
Configuration cache problems found in this build.

1 problem was found storing the configuration cache.
- Build file 'build.gradle.kts': external process started 'uname -a'
  See https://docs.gradle.org/7.6/userguide/configuration_cache.html#config_cache:requirements:external_processes
```

In the report file, here is some of the failure stack:
```
org.gradle.api.InvalidUserCodeException: Starting an external process 'uname -a' during configuration time is unsupported.
	at org.gradle.configurationcache.initialization.DefaultConfigurationCacheProblemsListener.onExternalProcessStarted(ConfigurationCacheProblemsListener.kt:84)
	at org.gradle.configurationcache.InstrumentedInputAccessListener.externalProcessStarted(InstrumentedInputAccessListener.kt:110)
	at org.gradle.internal.classpath.Instrumented.externalProcessStarted(Instrumented.java:394)
	at org.gradle.internal.classpath.Instrumented.externalProcessStarted(Instrumented.java:398)
	at org.gradle.internal.classpath.Instrumented.exec(Instrumented.java:259)
	at com.diffplug.common.swt.os.OS.exec(OS.java:156)
	at com.diffplug.common.swt.os.OS.calculateNative(OS.java:133)
	at com.diffplug.common.swt.os.OS.detectPlatform(OS.java:113)
	at com.diffplug.common.swt.os.OS.detectPlatform(OS.java:120)
	at com.diffplug.common.swt.os.OS.getNative(OS.java:100)
	at com.diffplug.gradle.FileMisc.<clinit>(FileMisc.java:426)
	at com.diffplug.gradle.eclipse.MavenCentralMapping.createVersionMap(MavenCentralMapping.java:138)
	at com.diffplug.gradle.eclipse.MavenCentralMapping.groupIdArtifactIdToVersion(MavenCentralMapping.java:124)
	at com.diffplug.gradle.eclipse.MavenCentralExtension$ReleaseConfigurer.<init>(MavenCentralExtension.java:56)
	at com.diffplug.gradle.eclipse.MavenCentralExtension.release(MavenCentralExtension.java:47)
	at com.diffplug.gradle.eclipse.MavenCentralExtension.release(MavenCentralExtension.java:43)
```

[...]

PR #1348, opened 2023-12-09
Upgrade to Spotless 2.43.1 and Spotless Gradle 6.23.3

liblit:
In addition to the usual automated tests, I've also manually verified that `./gradlew compile{,Test}Java` succeeds and correctly uses the configuration cache on an M2 macOS machine using either Java 11 or 17. So these changes should not cause a regression of #1230 or #1278.


msridhar (approved):
Thanks for the careful testing!

PR #1353, opened 2023-12-29
Update R8 and enable Dalvik tests on JDK 21

### LINKED ISSUE #1349: Dalvik tests do not pass on JDK 21
Currently the Dalvik tests do not pass on JDK 21.  Running on JDK 21 requires an R8 update, and D8's behavior seems to have changed in some way that causes test failures.  I suspect these are issues specific to R8/D8 and how it behaves, rather than any issues in our Dalvik frontend.

msridhar:
Updating to the latest R8 version lets us run the Dalvik tests on JDK 21.  We disable one check due to https://issuetracker.google.com/issues/316744331; we'll have to update R8 again once that is fixed.  We also needed to tweak how we construct the test jar for use with Dalvik.  My best guess is some classes caused issues with D8 before, but they no longer do so.  Finally, we performed some cleanup in `JVMLDalvikComparisonTest` to make failures easier to understand.

Fixes #1349

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -312,10 +312,10 @@ artifacts.add(collectTestDataJar.name, collectTestData.map { it.destinationDirec
 
 ////////////////////////////////////////////////////////////////////////
 //
-//  collect "com.ibm.wala.core.testdata_1.0.0a.jar"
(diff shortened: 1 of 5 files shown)


liblit (approved):
Thanks, @msridhar!

PR #1355, opened 2024-01-04
analysisScope toJson function

aakgna:
Added a new function to serialize an `AnalysisScope` to JSON format.


core/src/main/java/com/ibm/wala/ipa/callgraph/AnalysisScope.java
```diff
@@ -345,6 +351,30 @@ public String toString() {
...
   }
 
+  public String toJson() {
+    HashMap<String, Object> res = new HashMap<>();
+    HashMap<String, ArrayList<String>> loaders = new HashMap<>();
+    for (ClassLoaderReference loader : loadersByName.values()) {
+      ArrayList<String> arr = new ArrayList<>();
+      for (Module m : getModules(loader)) {
+        arr.add(m.toString());
+      }
```
msridhar, line 361: Here I think you could use the `Iterator2Collection.toList` type to avoid writing this loop: https://github.com/wala/WALA/blob/675cee88cd6fbfe82e7ad20f5d98e97430acb574/util/src/main/java/com/ibm/wala/util/collections/Iterator2Collection.java#L39

core/src/test/java/com/ibm/wala/core/tests/callGraph/CallGraphTest.java
```diff
@@ -680,4 +680,4 @@ public boolean hasEdge(MethodReference src, MethodReference dst) {
       }
     };
   }
-}
+}
```
msridhar, line 683: remove this change

core/src/test/java/com/ibm/wala/core/tests/cha/AnalysisScopeTest.java
```diff
@@ -55,4 +55,4 @@ public void testBaseScope() throws IOException, ClassHierarchyException {
                 ClassLoaderReference.Application, "Ljava/awt/AlphaComposite")),
         "found unexpected class");
   }
-}
+}
```
msridhar, line 58: remove this change

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -5,4 +5,4 @@ distributionUrl=https\://services.gradle.org/distributions/gradle-8.5-all.zip
 networkTimeout=10000
 validateDistributionUrl=true
 zipStoreBase=GRADLE_USER_HOME
-zipStorePath=wrapper/dists
+zipStorePath=wrapper/dists
```
msridhar, line 8: remove this change

core/src/main/java/com/ibm/wala/ipa/callgraph/AnalysisScope.java
```diff
@@ -345,6 +351,30 @@ public String toString() {
...
+    }
+    res.put("Loaders", loaders);
+    String[] exclusions = getExclusions().toString().split("\\|");
+    ArrayList<String> arr2 = new ArrayList<>();
+    for(int i = 0; i < exclusions.length; i++){
+      String word = exclusions[i];
+      word = word.replace("(", "");
+      word = word.replace(")", "");
+      arr2.add(word);
+    }
```
msridhar, line 372: Cool, way to go the extra mile to parse the exclusions!

msridhar (commented):
Overall looking good!  I have a few comments.  Also, we need to write a test for this new code.  You can add it in the AnalysisScopeTest class: https://github.com/wala/WALA/blob/master/core/src/test/java/com/ibm/wala/core/tests/cha/AnalysisScopeTest.java#L21-L21 The test should create an AnalysisScope and then confirm the JSON generated by `toJSON()` matches what we expect

core/src/main/java/com/ibm/wala/ipa/callgraph/AnalysisScope.java
msridhar: This comment is still not addressed

core/src/test/java/com/ibm/wala/core/tests/callGraph/CallGraphTest.java
msridhar: Still not addressed

core/src/test/java/com/ibm/wala/core/tests/cha/AnalysisScopeTest.java
```diff
@@ -55,4 +56,16 @@ public void testBaseScope() throws IOException, ClassHierarchyException {
...
+
+  @Test
+    public void testToJson() throws IOException {
+        AnalysisScope scope =
+            AnalysisScopeReader.instance.readJavaScope(
+                TestConstants.WALA_TESTDATA,
+                // new FileProvider().getFile("J2SEClassHierarchyExclusions.txt"),
+                new FileProvider().getFile("GUIExclusions.txt"),
+                AnalysisScopeTest.class.getClassLoader());
+        String exp = "{\"Loaders\":{\"Primordial\":[\"JarFileModule:/opt/homebrew/Cellar/openjdk/21.0.1/libexec/openjdk.jdk/Contents/Home/jmods/java.base.jmod\",\"Nested Jar File:primordial.jar.model\"],\"Extension\":[],\"Application\":[\"JarFileModule:/Users/aakgna/Documents/WALA-Research/WALA/core/build/resources/test/com.ibm.wala.core.testdata_1.0.0.jar\"],\"Synthetic\":[]},\"Exclusions\":[\"java\\\\/awt\\\\/.*\",\"javax\\\\/swing\\\\/.*\",\"sun\\\\/awt\\\\/.*\",\"sun\\\\/swing\\\\/.*\"]}";
```
msridhar, line 68: This is looking better!  But I expect this test will not pass in general, since we have absolute paths in it.  So I guess a straightforward string comparison won't work.  Here's an alternate approach:
1. Call `scope.toJson()` to get the JSON string
2. Re-parse this String using Gson
3. Check that specific parts of the resulting `Map` match what we expect

Can you try that?  Let me know if it doesn't make sense

core/src/main/java/com/ibm/wala/ipa/callgraph/AnalysisScope.java
msridhar: We decided to just punt on this, as it was causing other issues

core/src/test/java/com/ibm/wala/core/tests/cha/AnalysisScopeTest.java
```diff
@@ -55,4 +62,75 @@ public void testBaseScope() throws IOException, ClassHierarchyException {
...
+            for(int i = 0; i < exclusions.length; i++){
+                String word = exclusions[i];
+                word = word.replace("(", "");
+                word = word.replace(")", "");
+                arr2.add(word);
+            }
+            assertEquals(arr2, map.get("Exclusions"));
+        }
+        Type type2 = new TypeToken<LinkedHashMap<String, ArrayList<String>>>(){}.getType();
+        LinkedHashMap<String, ArrayList<String>> loaders = gson.fromJson(gson.toJson(map.get("Loaders")), type2);
```
msridhar, line 90: Here I am confused, why do you need to serialize the data in the map back to JSON and then parse it again?  I think you can get rid of this and just use some downcasts to get the data you want out of the map directly

core/src/test/java/com/ibm/wala/core/tests/cha/AnalysisScopeTest.java
```diff
@@ -55,4 +62,75 @@ public void testBaseScope() throws IOException, ClassHierarchyException {
...
+                TestConstants.WALA_TESTDATA,
+                // new FileProvider().getFile("J2SEClassHierarchyExclusions.txt"),
+                new FileProvider().getFile("GUIExclusions.txt"),
+                AnalysisScopeTest.class.getClassLoader());
+        Gson gson = new Gson();
+        Type type = new TypeToken<LinkedHashMap<String, Object>>(){}.getType();
+        LinkedHashMap<String, Object> map = gson.fromJson(scope.toJson(), type);
+        System.out.println(map);
+        if(map.containsKey("Exclusions")) {
+            String[] exclusions = scope.getExclusions().toString().split("\\|");
```
msridhar, line 79: Here we know what is in `GUIExclusions.txt`.  So I think we can simplify this code.  We can just do something like `assertEquals(List.of("java\/awt\/.*", "javax\/swing\/.*", "sun\/awt\/.*", "sun\/swing\/.*"), map.get("Exclusions"))`.  We can then get rid of this whole loop that re-parses the exclusions string

core/src/test/java/com/ibm/wala/core/tests/cha/AnalysisScopeTest.java
```diff
@@ -55,4 +62,75 @@ public void testBaseScope() throws IOException, ClassHierarchyException {
...
+                String word = exclusions[i];
+                word = word.replace("(", "");
+                word = word.replace(")", "");
+                arr2.add(word);
+            }
+            assertEquals(arr2, map.get("Exclusions"));
+        }
+        Type type2 = new TypeToken<LinkedHashMap<String, ArrayList<String>>>(){}.getType();
+        LinkedHashMap<String, ArrayList<String>> loaders = gson.fromJson(gson.toJson(map.get("Loaders")), type2);
+        if(loaders.containsKey("Primordial")) {
```
msridhar, line 91: Again we can simplify this code.  We know the contents of `wala.testdata.txt`:
```
Primordial,Java,stdlib,base
Primordial,Java,jarFile,primordial.jar.model
Application,Java,jarFile,com.ibm.wala.core.testdata_1.0.0.jar
```
So I think we can write assertions like:
* The keys of `loaders` should be exactly `"Primordial","Application","Extension","Synthetic"`
* Extension and Synthetic should be mapped to empty lists
* Primordial should be mapped to a list of length two.  And one of the elements should contain `"primordial.jar.model"
* Application should be mapped to a list of length 1, and that element should contain `"com.ibm.wala.core.testdata_1.0.0.jar"`

msridhar (commented):
Almost there!  A couple more comments on the tests

core/src/main/java/com/ibm/wala/ipa/callgraph/AnalysisScope.java
```diff
@@ -345,6 +346,31 @@ public String toString() {
     return result.toString();
   }
 
+  public String toJson() {
+    LinkedHashMap<String, Object> res = new LinkedHashMap<>();
+    LinkedHashMap<String, ArrayList<String>> loaders = new LinkedHashMap<>();
+    for (ClassLoaderReference loader : loadersByName.values()) {
+      ArrayList<String> arr = new ArrayList<>();
+
```
msridhar, line 354: remove blank line

core/src/main/java/com/ibm/wala/ipa/callgraph/AnalysisScope.java
```diff
@@ -345,6 +346,31 @@ public String toString() {
     return result.toString();
   }
 
+  public String toJson() {
```
msridhar, line 349: We should add Javadoc on this method to indicate what it does

core/src/test/java/com/ibm/wala/core/tests/cha/AnalysisScopeTest.java
```diff
@@ -55,4 +70,79 @@ public void testBaseScope() throws IOException, ClassHierarchyException {
...
+
+  @Test
+  public void testToJsonCustom() throws IOException {
+    AnalysisScope scope;
+    scope = AnalysisScope.createJavaAnalysisScope();
+    AnalysisScope tempScope =
+        AnalysisScopeReader.instance.readJavaScope(
+            TestConstants.WALA_TESTDATA,
+            new FileProvider().getFile("GUIExclusions.txt"),
+            AnalysisScopeTest.class.getClassLoader());
```
msridhar, line 110: We don't need to create this temp scope just to test the exclusions.  Just skip testing the exclusions here

core/src/test/java/com/ibm/wala/core/tests/cha/AnalysisScopeTest.java
```diff
@@ -55,4 +70,79 @@ public void testBaseScope() throws IOException, ClassHierarchyException {
...
+    scope.setExclusions(tempScope.getExclusions());
+    String[] stdlibs = WalaProperties.getJ2SEJarFiles();
+    Arrays.sort(stdlibs);
+    int cnt = 0;
```
msridhar, line 114: Add a comment explaining why you are stopping at 5.  Also the value `5` should be stored in a local variable and then that variable should be used in multiple places

msridhar: But this code would be simpler if you just got rid of the `cnt` variable and added everything.  Is there any reason to add the limit?

core/src/test/java/com/ibm/wala/core/tests/cha/AnalysisScopeTest.java
```diff
@@ -55,4 +70,79 @@ public void testBaseScope() throws IOException, ClassHierarchyException {
...
+    @SuppressWarnings("unchecked")
+    Map<String, List<String>> loaders = (Map<String, List<String>>) map.get("Loaders");
+    Set<String> loaderKeys =
+        new HashSet<>(List.of("Primordial", "Extension", "Application", "Synthetic"));
+    assertEquals(loaders.keySet(), loaderKeys);
+    assertEquals(5, loaders.get("Primordial").size());
+    assertThat(
+        loaders.get("Primordial"),
+        hasItem(
+            "JarFileModule:/Users/aakgna/Library/Java/JavaVirtualMachines/corretto-11.0.15/Contents/Home/jmods/java.base.jmod"));
```
msridhar, line 139: This check will not pass on other machines.  You have the `stdlibs` array above.  You should use the strings there (which give the full paths to the jars where the tests are running) to check that the right strings are in the JSON.

msridhar (commented):
Some more comments

core/src/test/java/com/ibm/wala/core/tests/cha/AnalysisScopeTest.java
```diff
@@ -55,4 +71,71 @@ public void testBaseScope() throws IOException, ClassHierarchyException {
...
+    Set<String> loaderKeys =
+        new HashSet<>(List.of("Primordial", "Extension", "Application", "Synthetic"));
+    assertEquals(loaders.keySet(), loaderKeys);
+    assertEquals(stdlibs.length, loaders.get("Primordial").size());
+    assertTrue(
+        loaders
+            .get("Primordial")
+            .get(0)
+            .contains(
+                "Java/JavaVirtualMachines/corretto-11.0.15/Contents/Home/jmods/java.base.jmod"));
```
msridhar, line 130: This does not fully address my comment.  Please take the paths directly from the `stdlibs` array and check that they match.

core/src/test/java/com/ibm/wala/core/tests/cha/AnalysisScopeTest.java
```diff
@@ -55,4 +71,61 @@ public void testBaseScope() throws IOException, ClassHierarchyException {
...
+    LinkedHashMap<String, Object> map = gson.fromJson(scope.toJson(), type);
+    assertEquals(List.of(), map.get("Exclusions"));
+
+    @SuppressWarnings("unchecked")
+    Map<String, List<String>> loaders = (Map<String, List<String>>) map.get("Loaders");
+    Set<String> loaderKeys =
+        new HashSet<>(List.of("Primordial", "Extension", "Application", "Synthetic"));
+    assertEquals(loaders.keySet(), loaderKeys);
+    assertEquals(stdlibs.length, loaders.get("Primordial").size());
+    assertTrue(loaders.get("Primordial").get(0).contains("/Contents/Home/jmods/java.base.jmod"));
```
msridhar, line 125: This is still wrong, and tests are still failing.  You need to take the relevant strings from the `stdlibs` array for this comparison, rather than just doing `get(0)` and comparing to a constant string.

msridhar (approved):

PR #1357, opened 2024-01-14
Upgrade to Eclipse 4.30.0

### LINKED ISSUE #1354: Figure out how to update Eclipse version dependence
We currently depend on Eclipse 4.21.0 for the ECJ front end, which is quite old; the latest release is 4.30.0 (2023-12).  Older releases also have some kind of vulnerability (though I don't think this is a huge concern for us), fixed in 4.29.0:

https://gitlab.eclipse.org/security/cve-assignement/-/issues/8

The issue with just updating is that more recent Eclipse versions require JDK 17+ to run.  I don't think we want to force all of WALA to require JDK 17 to run.  But perhaps there is a way to only require JDK 17 if you want to run the ECJ front end.

liblit:
This newer Eclipse release requires Java 17 or better.  However, we still prefer to keep most of WALA compatible with Java 11.  Therefore, this Eclipse upgrade adds nontrivial build infrastructure to let us use a newer Java release only where Eclipse dependencies require it.

A new `com.ibm.wala.jdk-version` property in `gradle.properties` allows setting the default Java version used throughout WALA.  If this property is set to 17 or higher, then that same version is used both for Eclipse-dependent and Eclipse-independent WALA components.  However, if this property is set to 16 or earlier, then it is only used for WALA's Eclipse-independent components; the Eclipse-dependent components will override this and use Java 17 instead.

In a default configuration, most of WALA will be compiled with (and for) Java 11, with the Eclipse-dependent components using Java 17 instead.

Resolves #1354.


settings.gradle.kts
```diff
@@ -1,6 +1,9 @@
 buildscript { dependencies { classpath("com.diffplug.spotless:spotless-lib-extra:2.43.1") } }
 
-plugins { id("com.diffplug.configuration-cache-for-platform-specific-build") version "3.44.0" }
+plugins {
+  id("com.diffplug.configuration-cache-for-platform-specific-build") version "3.44.0"
+  id("org.gradle.toolchains.foojay-resolver-convention") version "0.7.0"
```
msridhar, line 5: I was a bit hesitant to add this in on another project, since it downloads JDKs and then runs them.  But I guess since it comes from Gradle, it's probably pretty trustworthy?  The alternative to using this would just be documenting which JDK versions need to be installed and installing them ourselves on CI.

msridhar (commented):
Thanks for this change!  I have one specific comment below.

Also, so I understand at a high level, with this change, the Java version that Gradle is running on becomes irrelevant in terms of what JDK is used to compile WALA code and run the tests; that is entirely controlled by the new property in `gradle.properties` with customization for Eclipse projects.  Is this correct?

liblit:
> Also, so I understand at a high level, with this change, the Java version that Gradle is running on becomes irrelevant in terms of what JDK is used to compile WALA code and run the tests; that is entirely controlled by the new property in `gradle.properties` with customization for Eclipse projects. Is this correct?

Yes, that's exactly right. Note also that this property can be set on the `./gradle` command line, which takes precedence over `gradle.properties`.

settings.gradle.kts
liblit: I think it's trustworthy. It's also widely used, so I'd expect any vulnerability to be discovered quickly. Certainly, this is far from being the only code we download from the Internet and run as part of building and testing WALA.

> The alternative to using this would just be documenting which JDK versions need to be installed and installing them ourselves on CI.

Agreed. But I prefer an automated mechanism over prose instructions that each new WALA developer must manually apply.

That being said, manual pre-installation of suitable JDKs is still an option. If someone is working on WALA in a restricted environment that would prevent [toolchain auto-provisioning](https://docs.gradle.org/current/userguide/toolchains.html#sec:provisioning) from working, that's fine. WALA's Eclipse components will build using whatever JDK 17 [Gradle can find already installed on the system](https://docs.gradle.org/current/userguide/toolchains.html#sec:auto_detection) or [custom-configured](https://docs.gradle.org/current/userguide/toolchains.html#sec:custom_loc).

msridhar:
Thanks a lot, @liblit, everything looks good here.  Right now I don't think we have any docs on what Java versions are required to run WALA.  Do you think this belongs in the main README.md?  I'm wondering if we could briefly document that in this PR.  Alternately, we could do it in a separate PR before cutting the next release.

liblit:
> Right now I don't think we have any docs on what Java versions are required to run WALA. Do you think this belongs in the main README.md?

Sure, no harm in letting people know what to expect. I'll add that to the main `README.md` and update this pull request accordingly. Please withhold your approval of this PR until I've done that.

msridhar (approved):
Thanks again for this!

liblit:
You're welcome! It was a fun challenge.

PR #1365, opened 2024-01-22
Add some generic type arguments

msridhar:
Everything seems to compile fine with these changes.  Let's us get rid of a rawtypes suppression.


liblit (approved):
Hooray for incremental progress toward cleaner code!

Issue #1378, opened 2024-03-21
`AbstractScriptEntity.toString()` should call `AbstractScriptEntity.getName()`


khatchad:
Currently, `com.ibm.wala.cast.ir.translator.AbstractScriptEntity.toString()` calls `java.io.File.getName()` to get the script name. However, it should use `com.ibm.wala.cast.ir.translator.AbstractScriptEntity.getName()`, which also calls `java.io.File.getName()` under certain conditions.

## Problems

1. There could be inconsistencies between the name of an entity and its string representation under certain conditions.
2. (The bigger problem for me) I need to override both methods if I want to change the name in a subclass. I should just override  `com.ibm.wala.cast.ir.translator.AbstractScriptEntity.getName()`.

https://github.com/wala/WALA/blob/45fa20ff0d31d72ff1c363ee153511a57da1b40e/cast/src/main/java/com/ibm/wala/cast/ir/translator/AbstractScriptEntity.java#L42-L53

msridhar:
This change sounds fine to me assuming regression tests pass.  Want to put up a PR @khatchad?

PR #1380, opened 2024-03-21
Write `AbstractScriptEntity.toString()`  in terms of `getName()`

khatchad:
Fixes https://github.com/wala/WALA/issues/1378.


msridhar (approved):
Thanks!

PR #1382, opened 2024-03-28
Refactor `EclipseSourceFileModule` to make it more extension-friendly

khatchad:
The removed null checks seem to happen further up the class hierarchy. Replace the static construction method with a public constructor. A lone private constructor basically makes it basically impossible to subclass.

### DIFF
--- a/ide/src/main/java/com/ibm/wala/ide/classloader/EclipseSourceFileModule.java
+++ b/ide/src/main/java/com/ibm/wala/ide/classloader/EclipseSourceFileModule.java
@@ -21,7 +21,8 @@ public class EclipseSourceFileModule extends SourceFileModule {
 
   public EclipseSourceFileModule(IFile f) {
     super(
-        // NOTE: Passing `null` here causes an exception to be thrown in the ctor of `com.ibm.wala.classLoader.FileModule`.
+        // NOTE: Passing `null` here causes an exception to be thrown in the ctor of
+        // `com.ibm.wala.classLoader.FileModule`.
         f == null ? null : new File(f.getFullPath().toOSString()),
         f == null ? null : f.getName(),
         null);
--- a/ide/src/main/java/com/ibm/wala/ide/classloader/EclipseSourceFileModule.java
+++ b/ide/src/main/java/com/ibm/wala/ide/classloader/EclipseSourceFileModule.java
@@ -20,7 +20,10 @@ public class EclipseSourceFileModule extends SourceFileModule {
   protected final IFile f;
 
   public EclipseSourceFileModule(IFile f) {
-    super(f == null ? null : new File(f.getFullPath().toOSString()), f == null ? null : f.getName(), null);
+    super(
+        f == null ? null : new File(f.getFullPath().toOSString()),
+        f == null ? null : f.getName(),
+        null);
     this.f = f;
   }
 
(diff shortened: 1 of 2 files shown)


msridhar (commented):
I understand making the constructor public, but not the removal of the static method and the null checks.  Why do we need the latter?

khatchad:
> I understand making the constructor public, but not the removal of the static method and the null checks. Why do we need the latter?

I believe that the reason the ctor was originally made private was so that the null check can be enforced. The exception throw couldn't be done in the ctor because `super()` needs to be the first statement. So, we have the static method invoking a private constructor.

If we make the ctor public, then you don't need the static method because you can simply just call the public ctor. So, I attempted to move the null checks to the ctor. I think this can be done as the null checking actually happens already further up the class hierarchy.

We could keep both the (now public) ctor and the static method, but I don't think anyone would use the latter if the former is available.

ide/src/main/java/com/ibm/wala/ide/classloader/EclipseSourceFileModule.java
```diff
@@ -17,17 +17,13 @@
...
-    if (f == null) {
-      throw new IllegalArgumentException("null f");
-    }
-    return new EclipseSourceFileModule(f);
-  }
-
-  private EclipseSourceFileModule(IFile f) {
-    super(new File(f.getFullPath().toOSString()), f.getName(), null);
+  public EclipseSourceFileModule(IFile f) {
+    super(
```
msridhar, line 23: Can you add a comment here indicating that passing `null` to `super` will cause a failure in the `FileModule` constructor?  We should really get NullAway running on all this code to get rid of this kind of stuff :-)

khatchad:
> > I understand making the constructor public, but not the removal of the static method and the null checks. Why do we need the latter?
> 
> I believe that the reason the ctor was originally made private was so that the null check can be enforced. The exception throw couldn't be done in the ctor because `super()` needs to be the first statement. So, we have the static method invoking a private constructor.
> 
> If we make the ctor public, then you don't need the static method because you can simply just call the public ctor. So, I attempted to move the null checks to the ctor. I think this can be done as the null checking actually happens already further up the class hierarchy.
> 
> We could keep both the (now public) ctor and the static method, but I don't think anyone would use the latter if the former is available.

Perhaps a different way to put it is that it looked like the static method is encapsulating the object construction. But that encapsulation is too strong if we want to subclass.

ide/src/main/java/com/ibm/wala/ide/classloader/EclipseSourceFileModule.java
khatchad: Does it cause a failure? We would hit this code:

https://github.com/wala/WALA/blob/3a2ca2e599e1cf72073ed64d5cf5839dea93bf71/core/src/main/java/com/ibm/wala/classLoader/FileModule.java#L30-L32

Which is what we had previously. The only thing that would really be different is that the filename string would be null.

khatchad: Or, is the thought that it was clear in this file that an exception would be thrown in this case, and now with the new code, we would need to go two levels up to see the same thing?

msridhar: My point was that it'd be good to write a comment indicating that we deliberately pass `null` _knowing_ that there is logic to fail elsewhere, so a reader does not get confused and think it is valid to pass `null` to the super constructor

msridhar (approved):

PR #1383, opened 2024-03-28
[WIP] Fix M1 Mac configuration issue

msridhar:
It seems the latest Gradle update broke configuration on M1 Macs.  I was attempting to fix but got stuck.  I added an M1 Mac to the CI config so we catch this in the future hopefully.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -15,6 +15,9 @@ jobs:
         include:
           - os: macos-latest
             java: 11
+          # This is for an M1 Mac
+          - os: macos-14
+            java: 11
           - os: ubuntu-latest
             java: 11
           - os: windows-latest
--- a/cast/smoke_main/build.gradle.kts
+++ b/cast/smoke_main/build.gradle.kts
@@ -39,6 +39,10 @@ fun createXlatorConfig(isOptimized: Boolean): Configuration =
           attributes {
             attribute(OPTIMIZED_ATTRIBUTE, isOptimized)
             attribute(USAGE_ATTRIBUTE, objects.named(Usage::class, NATIVE_RUNTIME))
+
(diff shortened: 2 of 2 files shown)


msridhar:
@liblit you can see the configuration error that occurs here:

https://github.com/wala/WALA/actions/runs/8470614425/job/23208719124?pr=1383#step:7:129

I tried to fix this by adding these lines around line 42 of `cast/smoke_main/build.gradle.kts`:

```
// Dynamically determine the architecture string
val architectureName = rootProject.extra["archName"] as String
attribute(Attribute.of("org.gradle.native.architecture", String::class.java), architectureName)
```

But this broke config on other platforms, and it led to a new issue on M1 Macs (I got some message about duplicate tasks named `checkSmokeMain`).  I'm wondering if you might have time to take a look at this?  I am blocked from basically doing anything with WALA on my local laptop until we fix this without downgrading Gradle locally or something like that.

msridhar:
Also note that I don't really care if I can build the native code stuff on an M1 Mac, but I do need configuration to succeed :-)

liblit:
> I'm wondering if you might have time to take a look at this?

Yes, I can dive in. Perhaps as soon as today, but perhaps as late as this coming weekend.

> I am blocked from basically doing anything with WALA on my local laptop until we fix this without downgrading Gradle locally or something like that.

Please feel free to revert that Gradle upgrade until I can get it working for M1 too. I don't want you to be blocked waiting for me to fix this.

msridhar:
Another note: `:cast:cast:linkRelease` and `:cast:cast:linkDebug` were already failing on my M1 Mac before the Gradle upgrade.  It's not critical to me that they succeed.  We could exclude those tasks on the M1 Mac CI job if needed.

liblit:
Solved by #1387.

PR #1384, opened 2024-03-28
Refactor `EclipseSourceDirectoryTreeModule` for more extensibility

khatchad:
When I subclassed `EclipseSourceDirectoryTreeModule`, I found myself needing to make a copy of `rootIPath` because it is private in the super class so that I could override a method. Overloading the `makeFile()` method now allows a bit more flexibility with the type of `FileModule` being created.


khatchad:
Looks like this build needs a restart, but I am unable to restart it.

msridhar:
@khatchad I restarted the build.

msridhar (approved):

PR #1392, opened 2024-04-16
ClassHierarchy toJson Function

aakgna:
Creates a JSON variable containing a map from the Root Class to its Subclasses.


core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +708,97 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
     }
   }
 
+  // Puts ClassHierarchy into a JSON readable variable that maps the class to a list of its
+  // subclasses
```
msridhar, line 712: Please write proper Javadoc

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +708,97 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
...
+  public Object toJson() {
+    // lines 713-715: initialize variables (subclass and dag) to store important values like hashmap
+    // of each iteration and a list of all maps of class to subclasses
```
msridhar, line 715: Generally, don't refer to line numbers in comments, as they will keep changing as the code evolves.

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +708,97 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
...
+    HashMap<Object, Object> subclass = new HashMap<>();
+    Iterator<Node> children = root.getChildren();
+    ArrayList<HashMap<Object, Object>> dag = new ArrayList<>();
+    // lines 717-721: Goes through all subclasses of the root class, and adds the subclass hashmap
+    // to the list of all subclass hashmaps
+    while (children.hasNext()) {
+      Node temp = children.next();
+      HashMap<Object, Object> fin = helper_toJson(temp, subclass);
```
msridhar, line 723: Use lower camel case when naming Java functions (no underscores) https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +708,97 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
...
+    // lines 732-742: inserting the root class to its subclasses into the main hashmap while
+    // removing unnecessary substrings from the class name
+    Iterator<Node> root_children = root.getChildren();
+    Set<String> final_root_children = new HashSet<>();
+    while (root_children.hasNext()) {
+      Node val = root_children.next();
+      String child_name = val.getJavaClass().toString();
+      child_name = child_name.replace("<Primordial,", "");
+      child_name = child_name.replace("<Application,", "");
+      child_name = child_name.replace(">", "");
```
msridhar, line 744: This code pattern is copy-pasted multiple times.  It should be placed in its own function and then the copy-pasting should be replaced with calls to that function

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +708,97 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
...
   }
 
+  // Puts ClassHierarchy into a JSON readable variable that maps the class to a list of its
+  // subclasses
+  public Object toJson() {
+    // lines 713-715: initialize variables (subclass and dag) to store important values like hashmap
+    // of each iteration and a list of all maps of class to subclasses
+    HashMap<Object, Object> subclass = new HashMap<>();
+    Iterator<Node> children = root.getChildren();
+    ArrayList<HashMap<Object, Object>> dag = new ArrayList<>();
```
msridhar, line 718: I don't understand the purpose of this list.  As far as I can tell, every element of the list will be the exact same map.  What am I missing?

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +708,97 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
     }
   }
 
+  // Puts ClassHierarchy into a JSON readable variable that maps the class to a list of its
+  // subclasses
+  public Object toJson() {
+    // lines 713-715: initialize variables (subclass and dag) to store important values like hashmap
+    // of each iteration and a list of all maps of class to subclasses
+    HashMap<Object, Object> subclass = new HashMap<>();
```
msridhar, line 716: Can this be declared as `Map<String,Set<String>>` instead?  That will make it a lot clearer what is going on; that this data structure maps a string representation of each class / interface to all of its immediate subtypes

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +708,97 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
...
+      child_name = child_name.replace(">", "");
+      final_root_children.add(child_name);
+    }
+    dag.get(0).put(key, final_root_children);
+    Gson gson = new Gson();
+    return gson.toJson(dag.get(0));
+  }
+
+  // helper function to toJson that performs recursion to go through all of the DAG
+  public HashMap<Object, Object> helper_toJson(Node n, HashMap<Object, Object> hash) {
```
msridhar, line 753: Since this function always returns the `hash` parameter that gets passed in, it can just return `void` (the caller will always have access to the map it passed in)

msridhar (commented):
Thanks for this!  I've added some initial comments.  I think this is on the right track but there is some significant cleanup possible.

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +707,111 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
...
 
+  /*
+   * Puts ClassHierarchy into a JSON readable variable that maps the class to a list of its
+   * subclasses
+   */
+  public Object toJson() {
+    /*
+     * initialize variables (subclass and dag) to store important values like hashmap of each
+     * iteration and a list of all maps of class to subclasses
+     */
```
msridhar, line 718: Within a function you should use single-line comments, e.g.:
```suggestion
     // initialize variables (subclass and dag) to store important values like hashmap of each
     // iteration and a list of all maps of class to subclasses
```
(This is not an absolute rule, but it holds for the comments you've written here)

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +707,111 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
...
+    }
+    /*
+     * Removes unnecesarry parts from name of the class
+     */
+    String key = nodeToString(root);
+    /*
+     * inserting the root class to its subclasses into the main hashmap while removing
+     * unnecessary substrings from the class name
+     */
+    Iterator<Node> root_children = root.getChildren();
```
msridhar, line 737: Note camel case again; this should be `rootChildren`.  Please update all local variable names appropriately

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +707,111 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
...
+        Set<String> root = new HashSet<>();
+        hash.put(key, root);
+      }
+    }
+  }
+
+  /*
+   * Removed unnecessary part of Node by turning it into a String
+   */
+  public String nodeToString(Node n) {
```
msridhar, line 807: this can be private

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +707,111 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
     }
   }
 
+  /*
+   * Puts ClassHierarchy into a JSON readable variable that maps the class to a list of its
+   * subclasses
+   */
+  public Object toJson() {
```
msridhar, line 714: This should return a `String` just like `AnalysisScope.toJson()` does

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +707,86 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
     }
   }
 
+  /*
```
msridhar, line 710: Javadoc comments start with `/**`

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +707,86 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
...
+   * Puts ClassHierarchy into a JSON readable variable that maps the class to a list of its
+   * subclasses
```
msridhar, line 712: Rephrase:
```suggestion
   * Converts ClassHierarchy to a JSON String, mapping each class name to a list of subclass names
```

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +707,86 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
...
+   */
+  public String toJson() {
+    // initialize variables (subclass and dag) to store important values like hashmap of each
+    // iteration and a list of all maps of class to subclasses
```
msridhar, line 716: Does this comment need to be updated?  I don't think `dag` exists anymore?

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +707,86 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
...
+    HashMap<String, Set<String>> subclass = new HashMap<>();
```
msridhar, line 717: rename to `classNameToSubclassNames`

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +707,86 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
...
+    Iterator<Node> children = root.getChildren();
+    // Goes through all subclasses of the root class, and adds the subclass hashmap to the
+    // list of all subclass hashmaps
```
msridhar, line 720: Comment out of date

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +707,86 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
...
+    while (children.hasNext()) {
+      helperToJson(children.next(), subclass);
+    }
+    // Removes unnecesarry parts from name of the class
```
msridhar, line 724:
```suggestion
    // Removes unnecessary parts from name of the class
```

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +707,86 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
...
+    Set<String> finalRootChildren = new HashSet<>();
+    while (rootChildren.hasNext()) {
+      finalRootChildren.add(nodeToString(rootChildren.next()));
+    }
+    subclass.put(key, finalRootChildren);
+    Gson gson = new Gson();
+    return gson.toJson(subclass);
+  }
+
+  /*
```
msridhar, line 738: `/**` for Javadoc

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +707,86 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
...
+   * helper function to toJson that performs recursion to go through all of the DAG
+   */
+  private void helperToJson(Node n, HashMap<String, Set<String>> hash) {
+    if (n.children.size() > 0) {
+      // while loop: defines iterator to go through the class's subclasses
```
msridhar, line 743: This comment is unnecessary; pretty obvious from the code

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +707,86 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
...
+          // class name --> adds that new value into varibale, root, containig all it's subclasses
+          Iterator<Node> val = temp.getChildren();
+          Set<String> root = new HashSet<>();
+          while (val.hasNext()) {
+            Node tempVal = val.next();
+            String newVal = nodeToString(tempVal);
+            root.add(newVal);
+          }
+          // insert into hashmap containing all class to subclass pairs
+          hash.put(key, root);
```
msridhar, line 765: This code pattern is duplicated also in the `toJSON` method.  Can we extract the code to a single method and call that method in both places?

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +707,86 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
...
+        }
+      }
+    } else {
+      // remove unnecessary substring from class name and map that to an empty
+      // list to add into the hashmap
```
msridhar, line 770: outdated comment

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +707,86 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
...
+    }
+    subclass.put(key, finalRootChildren);
+    Gson gson = new Gson();
+    return gson.toJson(subclass);
+  }
+
+  /*
+   * helper function to toJson that performs recursion to go through all of the DAG
+   */
+  private void helperToJson(Node n, HashMap<String, Set<String>> hash) {
```
msridhar, line 741: I think the code in this method is a bit buggy.  I think the guarantee we want is that when `helperToJSON` finishes, the entire subtree of the classhierarchy rooted at `n` is in the `classNameToSubclassName` map.  So, the steps should be:
1. Compute `key = nodeToString(n)`.  If `key` is already in the map, return.
2. Create a set (or list) `subclassNames`.  Iterate through the children, and for each `child`, recursively call `helperToJSON(child)` and also add `nodeToString(child)` to `subclassNames`. 
3. After the loop, do `classToSubclassNames.put(key,subclassNames)`.

If you do things this way, you can get rid of special-case code for the root node in `toJSON` and just call `helperToJson` on the root, and everything should work.

msridhar (commented):
Looking better, but still needs a bit of work

msridhar:
@aakgna please fix the CI builds and then I can look again

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +707,69 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
...
+    Set<String> subclassNames = new HashSet<>();
+    while (children.hasNext()) {
+      Node temp = children.next();
+      subclassNames.add(nodeToString(temp));
+      helperToJson(temp, classNameToSubclassNames);
+    }
+    // Removes unnecesarry parts from name of the class
+    String key = nodeToString(root);
+    // inserting the root class to its subclasses into the main hashmap
+    classNameToSubclassNames.put(key, subclassNames);
```
msridhar, line 727: Why do we need this code?  Can't you just call `helperToJSON(root, classNameToSubclassNames)` and that will do what you want?

core/src/test/java/com/ibm/wala/core/tests/callGraph/ClassConstantTest.java
```diff
@@ -79,4 +79,14 @@ public void testClassConstants()
...
   }
+
+  @Test
+  public void classHierarchyToJson() throws ClassHierarchyException, IOException {
+    AnalysisScope scope =
+        CallGraphTestUtil.makeJ2SEAnalysisScope(
+            TestConstants.WALA_TESTDATA, CallGraphTestUtil.REGRESSION_EXCLUSIONS);
+    ClassHierarchy cha = ClassHierarchyFactory.make(scope);
+    Object unused = cha.toJson();
+    // assertEquals("", json);
```
msridhar, line 90: Don't we need to write some actual assertions about what we expect to be in the final JSON string?

msridhar (commented):
couple more comments

core/src/main/java/com/ibm/wala/ipa/cha/ClassHierarchy.java
```diff
@@ -706,6 +708,56 @@ private void recursiveStringify(Node n, StringBuilder buffer) {
...
+    // If the node is already processed, return to avoid infinite recursion
+    if (hash.containsKey(key)) {
+      return;
+    }
+
+    // Initialize a set to store the names of the subclasses
+    Set<String> subclassNames = new HashSet<>();
+
+    // Process all children of the current node
+    if (n.children.size() > 0) {
```
msridhar, line 739: By printing out the JSON in the test and looking it over, I realized that it only handles subclassing, but not interface implementations.  E.g., for the interface `java.util.Map` we see `"java.util.Map":[]` in the output, missing all the classes that implement that interface.  We should update this logic so that for interfaces, we treat all implementing classes and sub-interfaces as subtypes in the outputted JSON.  In other words, the JSON should map each class or interface to all of its immediate subtypes, i.e., direct subclasses, direct sub-interfaces, or classes directly implementing an interface.

msridhar (changes requested):
Looking better!  I did notice a gap in the logic, see below

PR #1398, opened 2024-05-26
Fix JUnitEntryPoints annotation check

### LINKED ISSUE #1373: Found bug in JUnitEntryPoints
When using junit4 annotations, the static method `make(IClassHierarchy cha)` can not return any entrypoint.

I found that `isTestEntryPoint(TypeName typeName)` might be wrong:
```java
  private static boolean isTestEntryPoint(TypeName typeName) {
    // WALA uses $ to refers to inner classes. We have to replace "$" by "."
    // to make it a valid class name in Java source code.
    String javaName =
        StringStuff.jvmToReadableType(typeName.getPackage() + "." + typeName.getClassName());

    return TEST_ENTRY_POINT_ANNOTATION_NAMES.contains(javaName);
  }
```
For the class `Lorg/abc/Main`, `typeName.getPackage()` will return `org.abc` and delete the starting `L`. While `jvmToReadableType` depends on this `L` to parse the className. Without the starting `L`, `jvmToReadableType` will return an empty string.

So, I changed the method to the following:
```java
    private static boolean isTestEntryPoint(TypeName typeName) {
        // WALA uses $ to refers to inner classes. We have to replace "$" by "."
        // to make it a valid class name in Java source code.
        String javaName = StringStuff.jvmToReadableType(
                typeName.toString().replace('$', '.'));

        return TEST_ENTRY_POINT_ANNOTATION_NAMES.contains(javaName);
    }
```
And it works.

Note that in my case, I have no inner classes to test. So I just changed the code by the comment.

msridhar:
Fixes #1373

### DIFF
--- a/core/src/main/java/com/ibm/wala/core/util/scope/JUnitEntryPoints.java
+++ b/core/src/main/java/com/ibm/wala/core/util/scope/JUnitEntryPoints.java
@@ -132,8 +132,7 @@ public class JUnitEntryPoints {
   private static boolean isTestEntryPoint(TypeName typeName) {
     // WALA uses $ to refers to inner classes. We have to replace "$" by "."
     // to make it a valid class name in Java source code.
-    String javaName =
-        StringStuff.jvmToReadableType(typeName.getPackage() + "." + typeName.getClassName());
+    String javaName = StringStuff.jvmToReadableType(typeName.toString().replace('$', '.'));
 
     return TEST_ENTRY_POINT_ANNOTATION_NAMES.contains(javaName);
   }
--- a/core/src/test/java/com/ibm/wala/core/tests/cha/JunitEntrypointSupportTest.java
+++ b/core/src/test/java/com/ibm/wala/core/tests/cha/JunitEntrypointSupportTest.java
@@ -1,5 +1,7 @@
 package com.ibm.wala.core.tests.cha;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
 import com.ibm.wala.core.tests.callGraph.CallGraphTestUtil;
 import com.ibm.wala.core.tests.util.TestConstants;
 import com.ibm.wala.core.util.scope.JUnitEntryPoints;
@@ -11,6 +13,7 @@ import com.ibm.wala.ipa.cha.ClassHierarchyFactory;
 import com.ibm.wala.util.CancelException;
 import com.ibm.wala.util.collections.Iterator2Collection;
 import java.io.IOException;
+import java.util.List;
 import java.util.Set;
 import org.junit.jupiter.api.Test;
 
@@ -23,7 +26,14 @@ public class JunitEntrypointSupportTest {
         CallGraphTestUtil.makeJ2SEAnalysisScope(
             TestConstants.WALA_TESTDATA, CallGraphTestUtil.REGRESSION_EXCLUSIONS);
     ClassHierarchy cha = ClassHierarchyFactory.make(scope);
-    Set<Entrypoint> entrypoints = Iterator2Collection.toSet(JUnitEntryPoints.make(cha).iterator());
-    System.err.println(entrypoints);
+    List<Entrypoint> entrypoints =
+        Iterator2Collection.toList(JUnitEntryPoints.make(cha).iterator());
+    assertEquals(2, entrypoints.size());
+    assertEquals(
+        "< Application, Ljunit/JunitTests, test1()V >([<Application,Ljunit/JunitTests>])",
+        entrypoints.get(0).toString());
+    assertEquals(
+        "< Application, Ljunit/JunitTests, <init>()V >([<Application,Ljunit/JunitTests>])",
+        entrypoints.get(1).toString());
   }
 }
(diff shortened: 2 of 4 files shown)


msridhar:
@khatchad could you look this over?  You contributed the original JUnit entrypoint support.  Unfortunately we don't have tests and it's a bit of a pain to write one, so I'm not too sure about this change.

(@liblit: it's hard for me to write a test since I need allow the testSubjects code / source set in the core module to depend on junit, and I'm not too sure how to do that.  If you have any tips on that, that'd be welcome.)

liblit:
> I need allow the testSubjects code / source set in the core module to depend on junit

I'm going to assume that you want JUnit 5, for which we already have [dependency versions defined in `gradle/libs.versions.toml`](https://github.com/wala/WALA/blob/09a71511f0d499a758d6ce08ffb394d5e3e04020/gradle/libs.versions.toml#L33-L38). It would also be possible to depend on JUnit 4, but that would require adding some lines to that same `gradle/libs.versions.toml` file.

Going with JUnit 5, I'm also going to guess that the dependency you want is the `junit-jupiter-api` artifact. I believe that's the JAR that contains the various annotations and other symbols used when authoring tests.

If my assumptions above are correct, then you can add the desired dependency by augmenting [the `dependencies` block in `core/build.gradle.kts`](https://github.com/wala/WALA/blob/09a71511f0d499a758d6ce08ffb394d5e3e04020/core/build.gradle.kts#L55-L68) as follows:

```kotlin
dependencies {
  ...
  "testSubjectsImplementation"(libs.junit.jupiter.api)
  "testSubjectsImplementation"(platform(libs.junit.bom))
}
```

**Important:** remember the double quotes around both uses of `testSubjectsImplementation`!

msridhar:
Thanks a lot @liblit!  I'm now able to add a regression test and I have a lot more confidence in the fix here (in particular, the test fails before applying the fix).

core/src/test/java/com/ibm/wala/core/tests/cha/JunitEntrypointSupportTest.java
```diff
@@ -0,0 +1,38 @@
...
+import com.ibm.wala.util.collections.Iterator2Collection;
+import java.io.IOException;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+public class JunitEntrypointSupportTest {
+
+  @Test
+  public void basic()
+      throws ClassHierarchyException, IllegalArgumentException, CancelException, IOException {
```
liblit, line 23: I'm seeing a compilation failure reported on this line when using ECJ:

```
The declared exception CancelException is not actually thrown by the method basic() from type JunitEntrypointSupportTest
```

liblit (changes requested):

core/src/test/java/com/ibm/wala/core/tests/cha/JunitEntrypointSupportTest.java
msridhar: Yup, my bad 🙂 Should be fixed now.

liblit (approved):

khatchad:
> @khatchad could you look this over? You contributed the original JUnit entrypoint support. Unfortunately we don't have tests and it's a bit of a pain to write one, so I'm not too sure about this change.

I'm unsure where the 'L' is being removed at the reporter mentioned. [I am seeing it being added](https://github.com/wala/WALA/issues/1373#issuecomment-2135731893). But, if it is now being removed, these changes look fine to me.

core/src/main/java/com/ibm/wala/core/util/scope/JUnitEntryPoints.java
```diff
@@ -132,8 +132,7 @@ public static Iterable<Entrypoint> make(IClassHierarchy cha) {
   private static boolean isTestEntryPoint(TypeName typeName) {
     // WALA uses $ to refers to inner classes. We have to replace "$" by "."
     // to make it a valid class name in Java source code.
-    String javaName =
-        StringStuff.jvmToReadableType(typeName.getPackage() + "." + typeName.getClassName());
+    String javaName = StringStuff.jvmToReadableType(typeName.toString().replace('$', '.'));
```
khatchad, line 135: It would be good to have an inner class test as well.

khatchad (commented):
A test method in an inner class would be a good test.

core/src/main/java/com/ibm/wala/core/util/scope/JUnitEntryPoints.java
msridhar: @khatchad can you sketch out a test case that would let us test this?  I'm not familiar with how nested classes are used with JUnit 4 / 5

khatchad: I can have a look at this soon.

khatchad: I think we just need to make sure we are getting the entry point corresponding to a method declared in a nested type.

khatchad: https://github.com/wala/WALA/pull/1409

PR #1402, opened 2024-05-27
Add and use version catalog updater

liblit:
## Add a dependency catalog updater plugin

[This plugin](https://github.com/littlerobots/version-catalog-update-plugin) will help us clean and update the shared version catalog in `gradle/libs.versions.toml`.  It does not run automatically.  Instead, it provides a few helpful housekeeping tasks in the `version catalog update` group that we can run manually whenever convenient.

## Update many dependencies

These updates were all discovered and applied using the `versionCatalogUpdate` task provided by the `nl.littlerobots.version-catalog-update` plugin.

For now, I'm considering only those upgrades that cause no test regressions and that require no changes to WALA's Java code.  There are a few additional update opportunities that currently cause regressions. We can consider those separately, along with any Java code changes needed to make them work correctly.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -15,6 +15,7 @@ plugins {
   alias(libs.plugins.kotlin.jvm)
   alias(libs.plugins.shellcheck)
   alias(libs.plugins.task.tree)
+  alias(libs.plugins.version.catalog.update)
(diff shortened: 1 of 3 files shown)


msridhar (approved):
Nice!

PR #1403, opened 2024-05-28
Upgrade to newer ECJ release

liblit:
The `versionCatalogUpdate` task found this upgrade opportunity, but we also needed to suppress a resource-leak diagnostic that the newer ECJ produces for some of our `JarFile`-manipulating code.

### DIFF
--- a/core/src/main/java/com/ibm/wala/ipa/callgraph/AnalysisScope.java
+++ b/core/src/main/java/com/ibm/wala/ipa/callgraph/AnalysisScope.java
@@ -415,6 +415,7 @@ public class AnalysisScope {
   /**
    * @return the rt.jar (1.4), core.jar (1.5), java.core.jmod (13) file, or null if not found.
    */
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/ibm/wala/ipa/callgraph/AnalysisScope.java
```diff
@@ -415,6 +415,7 @@ public ArrayClassLoader getArrayClassLoader() {
   /**
    * @return the rt.jar (1.4), core.jar (1.5), java.core.jmod (13) file, or null if not found.
    */
+  @SuppressWarnings("resource")
```
msridhar, line 418: Someday I'd really like to get [the Resource Leak Checker](https://checkerframework.org/manual/#resource-leak-checker) running on WALA to see how well it works...

msridhar (approved):

PR #1409, opened 2024-06-21
Add inner class test for JUnit entrypoints

khatchad:
Additional test for https://github.com/wala/WALA/pull/1398.

### DIFF
--- a/core/src/test/java/com/ibm/wala/core/tests/cha/JunitEntrypointSupportTest.java
+++ b/core/src/test/java/com/ibm/wala/core/tests/cha/JunitEntrypointSupportTest.java
@@ -1,6 +1,8 @@
 package com.ibm.wala.core.tests.cha;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.hasToString;
 
 import com.ibm.wala.core.tests.callGraph.CallGraphTestUtil;
 import com.ibm.wala.core.tests.util.TestConstants;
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/ibm/wala/core/tests/cha/JunitEntrypointSupportTest.java
```diff
@@ -25,12 +25,18 @@ public void basic() throws ClassHierarchyException, IllegalArgumentException, IO
     ClassHierarchy cha = ClassHierarchyFactory.make(scope);
     List<Entrypoint> entrypoints =
         Iterator2Collection.toList(JUnitEntryPoints.make(cha).iterator());
-    assertEquals(2, entrypoints.size());
+    assertEquals(4, entrypoints.size());
     assertEquals(
-        "< Application, Ljunit/JunitTests, test1()V >([<Application,Ljunit/JunitTests>])",
+        "< Application, Ljunit/JunitTests$A, test2()V >([<Application,Ljunit/JunitTests$A>])",
```
msridhar, line 30: For readability can we switch these assertions to use the Hamcrest matcher `assertThat` and `contains` methods?   You can see an example [here](https://github.com/wala/WALA/blob/df5caa49a8c1ce5592ae98f996da2d609f278f8f/util/src/test/java/com/ibm/wala/util/test/BasicGraphTest.java).  `contains` is varargs and checks that the list contents exactly equal the arguments in order.  We can then also get rid of the check on `entrypoints.size()`

core/src/testSubjects/java/junit/JunitTests.java
```diff
@@ -6,4 +6,10 @@ public class JunitTests {
 
   @Test
   public void test1() {}
+
+  public static class A {
+
+    @Test
```
msridhar, line 12: Wow, will JUnit actually run this method?  I had no idea 🙂

msridhar (commented):
Awesome, thanks for this contribution!

core/src/testSubjects/java/junit/JunitTests.java
khatchad: I think it does. But, if it doesn't, I think that may be a different issue :). In fact, I'm unsure which version of JUnit is being supported here.

msridhar: I think it depends a bit on the JUnit version:

https://stackoverflow.com/questions/8758294/test-cases-in-inner-classes-with-junit

But in this case, since `A` is `static`, it will run the test method.

core/src/test/java/com/ibm/wala/core/tests/cha/JunitEntrypointSupportTest.java
khatchad: > We can then also get rid of the check on `entrypoints.size()`

Don't we want to know if there are other entry points? There could be erroneous ones.

msridhar (approved):
Thanks a lot!

PR #1419, opened 2024-07-20
Support building and running on JDK 22

### LINKED ISSUE #1414: Support running on JDK 22
Do a change similar to #1339 but for JDK 22.

msridhar:
We don't immediately crash on JDK 22 bytecodes, but new bytecode features have not been tested.

We modify the `TestList` test input to avoid a blowup when running `SlicerTest.testList`.  https://github.com/openjdk/jdk/commit/b62e774e6a531db934de04211724a2a8159d94db introduced calls to `String.format` from `Integer.parseInt`, which dramatically increased the number of reachable methods for the previous version of this test, which in turn led to very slow performance with the slicer.  Changing the test to use a `List<String>` and string constants removes this blowup.  (Whole-program analysis is fun!)

Fixes #1414

### DIFF
--- a/core/src/testSubjects/java/slice/TestList.java
+++ b/core/src/testSubjects/java/slice/TestList.java
@@ -8,10 +8,10 @@ public class TestList {
   static void doNothing(Object o) {}
 
   public static void main(String[] args) {
-    List<Integer> list = new ArrayList<>();
+    List<String> list = new ArrayList<>();
 
-    list.add(2);
-    list.add(3);
+    list.add("hi");
+    list.add("bye");
 
     doNothing(list.get(0));
   }
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -26,6 +26,8 @@ jobs:
             java: 17
           - os: ubuntu-latest
             java: 21
+          - os: ubuntu-latest
+            java: 22
       fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -362,9 +362,10 @@ tasks.named<Test>("test") {
   testLogging {
     exceptionFormat = TestExceptionFormat.FULL
     events("passed", "skipped", "failed", "started")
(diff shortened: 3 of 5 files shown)


core/src/testSubjects/java/slice/TestList.java
```diff
@@ -8,10 +8,10 @@ public class TestList {
   static void doNothing(Object o) {}
 
   public static void main(String[] args) {
-    List<Integer> list = new ArrayList<>();
+    List<String> list = new ArrayList<>();
 
-    list.add(2);
-    list.add(3);
+    list.add("hi");
+    list.add("bye");
```
liblit, line 14: What is the relevance of these changes to supporting JDK 22?

msridhar: This is what I was writing about in the PR summary:

> We modify the `TestList` test input to avoid a blowup when running `SlicerTest.testList`.  https://github.com/openjdk/jdk/commit/b62e774e6a531db934de04211724a2a8159d94db introduced calls to `String.format` from `Integer.parseInt`, which dramatically increased the number of reachable methods for the previous version of this test, which in turn led to very slow performance with the slicer.  Changing the test to use a `List<String>` and string constants removes this blowup.  (Whole-program analysis is fun!)

That OpenJDK commit is present in JDK 22 but not JDK 21.  (`Integer.parseInt` is relevant since it is invoked from the static initializer of `Integer$IntegerCache`, which is referenced from `Integer.valueOf`, which is implicitly invoked by the previous version of the test due to auto-boxing.)

liblit: Ah, right. Sorry I overlooked this clear explanation that you'd already provided in the PR summary.

msridhar: No worries!  Do you think I should add a comment somewhere in the test case about this?

liblit: I'd be satisfied with a similar explanation in the commit message. A comment in the code is fine too, if you prefer that.

msridhar: Ok, sure.  The PR summary will turn in to the commit message.

liblit (approved):

PR #1421, opened 2024-07-21
Add a Markdown linter to `check` validation tasks

### LINKED ISSUE #818: Add Markdown linter as part of `./gradlew check`
See https://github.com/wala/WALA/pull/816#issuecomment-688579557

liblit:
I would rather use Spotless for these checks, for consistency with other format checkers that we already use.  [Spotless supports Markdown via flexmark](https://github.com/diffplug/spotless/tree/main/plugin-gradle#flexmark), but unfortunately [Spotless does not provide access to any of flexmark's configuration options](https://github.com/diffplug/spotless/issues/2183).  That would leave us with a Markdown linter that is too lenient for my tastes.  For example, [Spotless's flexmark configuration does not even restrict maximum line length](https://github.com/diffplug/spotless/issues/2167).

Resolves #818.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,7 +26,7 @@ Beyond tests, other checks run as part of `./gradlew check` and
 
 1. Compilation with the Java compiler from [Eclipse JDT
    Core](https://www.eclipse.org/jdt/core/), which runs additional
-   lint checks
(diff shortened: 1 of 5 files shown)


msridhar (approved):
Nice!

Issue #521, opened 2019-06-07
switch to gradle-maven-publish plugin for cutting releases


msridhar:
https://github.com/vanniktech/gradle-maven-publish-plugin

Looks like a maintained version of the script we are using.  We should try it.

msridhar:
Would still be nice to do this, for this reason alone (from [the README](https://github.com/vanniktech/gradle-maven-publish-plugin?tab=readme-ov-file#advantages-over-maven-publish)):

> Maven central integration. The plugin makes it easy to configure publishing to Maven Central with dedicated APIs to set it up and configure everything that is required. It also avoids issues like having multiple staging repositories on Sonatype OSS and supports automatic releasing without requiring any interaction with the web interface.

@liblit how hard do you think it'd be to switch our scripts over to use this?

liblit:
Most WALA subprojects load the `publishing.gradle.kts` script, but that script itself does not export anything that importing subprojects use. The script‘s ”API“, if you want to think of it that way, is the empty set. So replacing this script with a different script should not disrupt other build logic.

Now, there are some unusual things about our publication logic. For example, we create and publish a Javadoc artifact that covers all of WALA. So we would need to figure out how to do any of these atypical things using the proposed replacement plugin. Having not looked at the latter yet, I cannot say how customizable it might be.

All in all, though, this change sounds pretty doable to me.

@msridhar, would you like me to take a swing at this change? Or are you already trying it yourself?

msridhar:
I am not working on this. I just noticed the pain point when cutting a release today. That said I would say it is low priority and it's certainly not blocking anything critical. If you think it'd be fun to try @liblit feel free to take a crack at it 🙂

PR #1427, opened 2024-07-27
Publish using gradle-maven-publish plugin (#521)

liblit:
Thoroughly testing these changes is delicate, since publishing to Maven Central involves an outside service, authentication, etc.  Instead, I've checked these changes in a limited manner:  I have verified that the `publishAllPublicationsToFakeRemoteRepository` task creates a local Maven repository containing exactly the same named files (in the same locations) as that task did before these changes.

### DIFF
--- a/build-logic/build.gradle.kts
+++ b/build-logic/build.gradle.kts
@@ -13,6 +13,7 @@ dependencies {
   implementation(libs.gradle.download.task)
   implementation(libs.gradle.errorprone.plugin)
   implementation(libs.gradle.goomph.plugin)
+  implementation(libs.gradle.maven.publish.plugin)
   implementation(libs.gradle.spotless.plugin)
 }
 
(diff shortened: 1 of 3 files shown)


msridhar (approved):
This is great!  I confirmed that I can upload a SNAPSHOT build after these changes using the instructions [here](https://vanniktech.github.io/gradle-maven-publish-plugin/central/#publishing-snapshots).  We should update our [release instructions](https://github.com/wala/WALA/wiki/Creating-Releases-and-Snapshots) when we cut the next WALA release.  In particular, we can get rid of the `---no-parallel` flag with this plugin as well, which should speed things up!

Thanks for this, @liblit!

liblit:
Thanks for that extra validation, @msridhar. I'm glad that this change will make future releases easier for us (well, you).

PR #1435, opened 2024-08-10
Simplify helper functions for native code

liblit:
The `cast/helpers.kt` script defines various functions that help us configure tasks that work with native code, such as natively compiled libraries.  I ported these from Groovy to Kotlin when I was still fairly new to the latter.  I've gotten better with Kotlin since then, and I now see how some of these functions could be simplified.

For example, judicious use of extension methods makes some functions easier to call, and also makes it easier for those functions' implementations to access properties of the receiver object.  In other cases, some function parameters can be computed from other parameters, thereby making the function easier to call correctly and harder to call incorrectly.

None of this constitutes a radical change to the suite of helper functions or how they are used to configure tasks.  It's just a bunch of small cleanups and simplifications from 2024-me, who knows Kotlin better than 2022-me.

---

**Reviewer tip:** ignore whitespace for a more semantic diff view.

### DIFF
--- a/build-logic/src/main/kotlin/com/ibm/wala/gradle/cast/helpers.kt
+++ b/build-logic/src/main/kotlin/com/ibm/wala/gradle/cast/helpers.kt
@@ -1,10 +1,12 @@
 package com.ibm.wala.gradle.cast
 
 import java.io.File
-import org.gradle.api.Project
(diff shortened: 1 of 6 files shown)


msridhar (approved):
Thanks for this cleanup!

PR #1442, opened 2024-09-19
Actually generate javadoc on latest JDK version

msridhar:
We need to pass `-Pcom.ibm.wala.jdk-version` in the `generate-latest-docs.sh` script.

Attempted to test locally, but the real test will be when this lands.


liblit (approved):

msridhar:
Looks like it worked!

https://wala.github.io/javadoc/

I confirmed with "View Source" that these docs are generated on JDK 23.

PR #1462, opened 2024-11-16
Iterate over Map entries to avoid extra lookups

liblit:


util/src/main/java/com/ibm/wala/dataflow/graph/DataflowSolver.java
```diff
@@ -119,9 +120,9 @@ private class UnionFind {
     private final Object[] allKeys;
 
     private int mapIt(int i, Object[] allVars, Map<Object, V> varMap) {
-      for (Object key : varMap.keySet()) {
-        allKeys[i] = key;
-        allVars[i++] = varMap.get(key);
+      for (Entry<Object, V> entry : varMap.entrySet()) {
+        allKeys[i] = entry.getKey();
+        allVars[i++] = entry.getValue();
```
msridhar, line 125: Is this really always better?  I often wonder if iterating this way causes a bunch of `Entry` objects to be allocated, even though the lookup is avoided.

liblit: The advice I'm seeing is that the following are ordered faster to slower:

1. `keySet` without `get` for each key
1. `entrySet`
1. `keySet` with `get` for each key

There's no guarantee about those `Entry` instances being allocated ahead of time versus on-demand. But even if the latter, the cost of `get` on every key exceeds the cost of those short-lived allocations.

msridhar (approved):

PR #1470, opened 2024-11-23
Update Google Java Format to 1.25.0

msridhar:
Just to stay up to date

### DIFF
--- a/build-logic/src/main/kotlin/com/ibm/wala/gradle/java.gradle.kts
+++ b/build-logic/src/main/kotlin/com/ibm/wala/gradle/java.gradle.kts
@@ -164,3 +164,8 @@ spotless {
             .toString())
   }
 }
+
+// Google Java Format versions 1.25.0 and higher require Java 17
+tasks.named("spotlessJava") {
+  onlyIf { JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_17) }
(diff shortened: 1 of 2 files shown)


liblit (approved):

liblit (changes requested):
Something ain't right in the automated CI jobs:

```
Caused by: java.lang.UnsupportedClassVersionError: com/google/googlejavaformat/java/JavaFormatterOptions$Style has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 55.0
```

msridhar:
> Something ain't right in the automated CI jobs:
> 
> ```
> Caused by: java.lang.UnsupportedClassVersionError: com/google/googlejavaformat/java/JavaFormatterOptions$Style has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 55.0
> ```

Yeah, GJF requires Java 17 now.  To update we'll have to only enable spotless tasks when the Gradle JVM is JDK 17+.  Does that sound reasonable, @liblit?  Shouldn't lead to any gaps on CI since we run jobs with the Gradle JVM at 17, 21, etc.

liblit (approved):

PR #1476, opened 2024-11-28
Add lambda support to CHACallGraph

### LINKED ISSUE #1459: Support lambdas in CHA call graph construction
This is a notable gap in WALA's call graph construction algorithms for Java bytecode.  I haven't worked out the full algorithm but I don't think this should be too hard to support.  Need to somehow pattern match on the IR or bytecodes and model the class that gets generated for the lambda.  Not sure if we want to also update the `ClassHierarchy` in the process.

msridhar:
Fixes #1459 

To support lambdas, we re-use logic from the `LambdaMethodTargetSelector` to handle calls that use `invokedynamic`.  `CHACallGraph` only includes edges for methods reachable from given entrypoints, and our handling of lambdas adds new synthetic classes to the class hierarchy (representing the classes generated for lambdas).  So to ensure a complete call graph, we do a second pass after adding these synthetic classes to recompute call targets.  This is probably not the most efficient way to do things, but I'm not sure it's the bottleneck; if we need to optimize in the future we should do so based on profiling.

Note that the handling of lambdas here is very imprecise; every call site of a functional interface method will have an edge to every lambda of that functional interface type.  Still, it could be practically useful, e.g., when combined with other heuristic filtering of irrelevant code.

We add new tests, and also fix a couple other minor bugs (e.g., there was no edge before to the synthetic method that invokes class initializers, so it wasn't being processed properly).


cast/src/main/java/com/ibm/wala/cast/ipa/cha/CrossLanguageClassHierarchy.java
```diff
@@ -276,5 +276,13 @@ public static CrossLanguageClassHierarchy make(AnalysisScope scope, ClassLoaderF
...
     return HashSetFactory.make();
   }
+
   /* END Custom change: unresolved classes */
+
+  @Override
+  public void clearCaches() {
+    for (IClassHierarchy cha : hierarchies.values()) {
+      cha.clearCaches();
+    }
```
liblit, line 286: I'd prefer a tacit style here:

```suggestion
    hierarchies.values().forEach(IClassHierarchy::clearCaches);
```

OK to ignore this suggestion if you prefer the explicit loop.

core/src/main/java/com/ibm/wala/ipa/callgraph/cha/CHACallGraph.java
```diff
@@ -138,6 +145,28 @@ public void init(Iterable<Entrypoint> entrypoints) throws CancelException {
...
+    // to update call targets to include lambdas, we clear all call target caches, iterate through
+    // all call sites, and re-compute the targets.
+    // TODO optimize if needed
+    targetCache.clear();
+    cha.clearCaches();
+    for (CGNode n : this) {
+      for (CallSiteReference site : Iterator2Iterable.make(n.iterateCallSites())) {
+        Iterator<IMethod> methods = getOrUpdatePossibleTargets(n, site);
+        while (methods.hasNext()) {
+          IMethod target = methods.next();
```
liblit, line 158:
```suggestion
        for (IMethod target : Iterator2Iterable.make(getOrUpdatePossibleTargets(n, site))) {
```

core/src/main/java/com/ibm/wala/ipa/callgraph/cha/CHACallGraph.java
```diff
@@ -138,6 +145,28 @@ public void init(Iterable<Entrypoint> entrypoints) throws CancelException {
...
+          if (isRelevantMethod(target)) {
+            CGNode callee = getNode(target, Everywhere.EVERYWHERE);
+            if (callee == null) {
+              throw new RuntimeException(
+                  "should have already created CGNode for " + target.toString());
```
liblit, line 163:
```suggestion
                  "should have already created CGNode for " + target);
```

core/src/main/java/com/ibm/wala/ipa/callgraph/cha/CHACallGraph.java
```diff
@@ -146,6 +175,12 @@ public IClassHierarchy getClassHierarchy() {
     return cha;
   }
 
+  /**
+   * Cache of possible targets for call sites.
+   *
+   * <p>In the future, this cache could be keyed on (MethodReference,isDispatch) pairs to save space
+   * and possibly time, where isDispatch indicates whether the call site is a virtual dispatch.
```
liblit, line 182:
```suggestion
   * <p>In the future, this cache could be keyed on ({@link com.ibm.wala.types.MethodReference},
   * {@code isDispatch}) pairs to save space and possibly time, where {@code isDispatch} indicates
   * whether the call site is a virtual dispatch.
```

core/src/main/java/com/ibm/wala/ipa/callgraph/cha/CHACallGraph.java
```diff
@@ -279,7 +347,7 @@ private void closure() throws CancelException {
     while (!newNodes.isEmpty()) {
       CGNode n = newNodes.pop();
       for (CallSiteReference site : Iterator2Iterable.make(n.iterateCallSites())) {
-        Iterator<IMethod> methods = getOrUpdatePossibleTargets(site);
+        Iterator<IMethod> methods = getOrUpdatePossibleTargets(n, site);
         while (methods.hasNext()) {
           IMethod target = methods.next();
```
liblit, line 352:
```suggestion
        for (IMethod target : Iterator2Iterable.make(getOrUpdatePossibleTargets(n, site))) {
```

core/src/main/java/com/ibm/wala/ipa/summaries/LambdaMethodTargetSelector.java
```diff
@@ -99,6 +99,26 @@ public IMethod getCalleeTarget(CGNode caller, CallSiteReference site, IClass rec
     return base.getCalleeTarget(caller, site, receiver);
   }
 
+  /**
+   * Get the summary class for a lambda factory, if it has already been created.
```
liblit, line 103:
```suggestion
   * Gets the summary class for a lambda factory, if it has already been created.
```

core/src/main/java/com/ibm/wala/ipa/cha/IClassHierarchy.java
```diff
@@ -186,4 +186,10 @@ public interface IClassHierarchy extends Iterable<IClass> {
    * @throws IllegalArgumentException if c2 is null
    */
   boolean isAssignableFrom(IClass c1, IClass c2);
+
+  /**
+   * clear internal caches that may be invalidated by addition of new classes, e.g., a cache of the
+   * results of {@link #getPossibleTargets(MethodReference)}
```
liblit, line 192:
```suggestion
   * Clear internal caches that may be invalidated by addition of new classes, e.g., a cache of the
   * results of {@link #getPossibleTargets(MethodReference)}.
```

core/src/main/java/com/ibm/wala/ipa/summaries/LambdaMethodTargetSelector.java
```diff
@@ -99,6 +99,26 @@ public IMethod getCalleeTarget(CGNode caller, CallSiteReference site, IClass rec
...
+   *
+   * @param caller the caller node
+   * @param site the call site reference
+   * @return the summary class for the lambda factory, or null if it has not been created
```
liblit, line 107:
```suggestion
   * @return the summary class for the lambda factory, or {@code null} if it has not been created
```

liblit (changes requested):
Looks good. I suggest a few minor tweaks to improve code and Javadoc style.

msridhar:
> Looks good. I suggest a few minor tweaks to improve code and Javadoc style.

Thanks for the detailed review!  All suggestions applied.

liblit (approved):

PR #1477, opened 2024-12-24
Fix a couple of bugs in parsing array type arguments in type signatures

msridhar:
We forgot to increment a variable, and we didn't handle all the primitive types.


liblit (approved):
Thanks for adding tests!

PR #1485, opened 2025-03-06
refactor: remove getTestName from IRTests.java so tests do not rely on method name

jkhaliqi:
Cleaned up `ECJJava15IRTest.java` by refactoring the tests to be parameterized to not rely on method name for running the tests. Also removed the `getTestName` from `IRTests` so the tests do not rely on that method to get the name of the test it should run. With this we pass in the name and tests can now be parameterized as well instead of relying on naming tests differently

### DIFF
--- a/cast/java/ecj/src/test/java/com/ibm/wala/cast/java/test/ECJJava15IRTest.java
+++ b/cast/java/ecj/src/test/java/com/ibm/wala/cast/java/test/ECJJava15IRTest.java
@@ -14,7 +14,9 @@ import com.ibm.wala.util.CancelException;
 import java.io.IOException;
 import java.util.Collection;
 import java.util.List;
-import org.junit.jupiter.api.Test;
+import java.util.stream.Stream;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
 
 public class ECJJava15IRTest extends IRTests {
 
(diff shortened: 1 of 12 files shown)


jkhaliqi:
@msridhar @juliandolby Hey whenever you get a chance could I get review on this, thank you!

msridhar:
@jkhaliqi I love it!  Do you have time to take this further?  This is the key function that pulls out the name of the currently-running test:

https://github.com/wala/WALA/blob/90350c797ed8cb23ea165bb7b50da86110d895e4/cast/java/src/testFixtures/java/com/ibm/wala/cast/java/test/IRTests.java#L357-L366

I'd love it if we could push this cleanup further so that the function above could just be deleted and everywhere we rely on it we instead use parameterized tests.  I think it would involve a few more changes similar to what you've already done.  What do you think?

jkhaliqi:
Hey sounds good, and thank you! I'll give that a try and see how far I get

cast/java/ecj/src/test/java/com/ibm/wala/cast/java/test/ECJJava15IRTest.java
```diff
@@ -39,241 +41,39 @@ protected Iterable<Entrypoint> makeDefaultEntrypoints(IClassHierarchy cha) {
...
+        "Annotations",
+        "Cocovariant",
+        "ExplicitBoxingTest",
+        "Wildcards",
+        "BasicsGenerics",
+        "CustomGenericsAndFields",
+        "TypeInferencePrimAndStringOp",
+        "SimpleEnums2",
+        "GenericMemberClasses",
+        "MoreOverriddenGenerics");
```
liblit, line 66: If the order of these items does not matter, then please sort them alphabetically. That will make this list easier to scan when adding new items, easier to avoid adding duplicates, etc..

liblit (changes requested):

msridhar:
@jkhaliqi is this ready for review?

jkhaliqi:
@msridhar Yup, I believe this is ready for review, whenever you get a chance, thank you!

jkhaliqi:
I just changed the JavaIRTests to be alphabetical as well, but all should be good now!

msridhar (commented):
Looking great!  A couple of comments particularly on `JavaIRTests`

cast/java/src/testFixtures/java/com/ibm/wala/cast/java/test/JavaIRTests.java
```diff
@@ -80,295 +84,271 @@ public JavaIRTests() {
...
-                assertTrue("Expected class to be either " + typeStr + " or " + "Bar", cls.getName().getClassName().toString()
-                    .equals(typeStr)
-                    || cls.getName().getClassName().toString().equals("Bar"));
-              }
-              */
-            }),
-        true,
-        null);
-  }
+  static List<? extends IRAssertion> caForInterfaceTest1 =
```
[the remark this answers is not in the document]
jkhaliqi: I was exactly sure what `ca` also meant I got this variable name from the fourth parameter inside of `runTests`. But i'm assuming it has something to do with the `callGraph` and just changed it to be `callAssertionsForXXX`. Let me know if theres a better name.

cast/java/src/testFixtures/java/com/ibm/wala/cast/java/test/JavaIRTests.java
```diff
@@ -80,295 +84,271 @@ public JavaIRTests() {
...
-        null);
+        simpleTestEntryPoint(java17IRTestName),
+        ca,
+        assertReachable,
+        exclusionsFile);
   }
 
   @Test
-  public void testArrayLiteral2() throws IllegalArgumentException, CancelException, IOException {
+  public void testTwoClasses() throws IllegalArgumentException, CancelException, IOException {
```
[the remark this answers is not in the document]
jkhaliqi: Some of the remaining tests call runTests and store its output in a variable and then do other stuff with it so I just left those alone.

msridhar:
@jkhaliqi can you just add commits to your branch from now on instead of force pushing?  That will make it easier to see what is changing.  I will squash the commits before merging.

jkhaliqi:
@msridhar Sounds good, sorry about that!

cast/java/src/testFixtures/java/com/ibm/wala/cast/java/test/JavaIRTests.java
```diff
@@ -80,371 +84,411 @@ public JavaIRTests() {
...
   public void testInnerClassA() throws IllegalArgumentException, CancelException, IOException {
     Pair<CallGraph, CallGraphBuilder<? super InstanceKey>> x =
-        runTest(singleTestSrc(), rtJar, simpleTestEntryPoint(), new ArrayList<>(), true, null);
+        runTest(
+            singleTestSrc("InnerClassA"),
+            rtJar,
+            simpleTestEntryPoint("InnerClassA"),
+            new ArrayList<>(),
+            true,
+            null);
```
msridhar, line 491: There are many call sites now where we pass exactly these arguments, except that the test name is different.  Could we possibly add a new wrapper method for `runTest` that takes a single `String` argument and then calls `runTest` appropriately, and then replace all these other calls with a call to the wrapper method?

msridhar (commented):
Thanks again so much for this cleanup!  One more comment / question

cast/java/src/testFixtures/java/com/ibm/wala/cast/java/test/JavaIRTests.java
jkhaliqi: Added in the wrapper, thank you!

msridhar (dismissed):

msridhar:
Thanks again for the contribution!

msridhar:
@liblit I think you will need to approve again before this can land

liblit (changes requested):
According to [this bot comment](https://github.com/wala/WALA/pull/1485#issuecomment-2704954733), this change results in a net increase of 101 tests. Is that expected?

msridhar:
> @liblit I believe so, is there a way I can double check all the test cases the bot only shows a couple before showing the ...

Yeah, this is pretty hard to check based on the [bot comment](https://github.com/wala/WALA/pull/1485#issuecomment-2704954733) which cuts off the output.  Any ideas @liblit?  I can see various artifacts related to the executed tests at the bottom of the [page for the action run](https://github.com/wala/WALA/actions/runs/13841881532?pr=1485) but not clear how to get the full differences from that.

liblit:
You could run the tests on your local machine. Test before and after your proposed changes. Presumably each of those will create _some_ kind of artifact you can compare without truncation.

msridhar:
> You could run the tests on your local machine. Test before and after your proposed changes. Presumably each of those will create _some_ kind of artifact you can compare without truncation.

Sure, I guess I was wondering if there's an easy way to run whatever code generates the comments on the pull requests, though I guess one could just run `diff` on the relevant files.

msridhar:
In [0db3fc9](https://github.com/wala/WALA/pull/1485/commits/0db3fc988f7a12420c6ac15bdd9ad4778f494e27) I tried bumping up the limit on how many test changes are listed in the comment, using the `test_changes_limit` setting, but it didn't seem to work 😐

jkhaliqi:
@liblit Can you explain more on the test and artifact comparison I wasn't able to follow or find the artifacts created after running the test. But I did just run and compare and everything is the same as far as these tests with first number being current change and second one before changes. There's only an increase in one for the tests because the @parameterized part will run but everything else ran the same...
java15irtest -> 22 -> 21
java17 -> 7 -> 6
testcomments-> same
Issue666Test.java -> same
Issue667Test -> same
jlextest -> same 
javairtest -> 37->  37 
syncduplicator -> same

liblit (approved):

msridhar:
Thanks again for the contribution @jkhaliqi!

PR #1493, opened 2025-04-11
JS ES6 Support and Template Literals Support

apple5775:
For javascript, add ES6 Support for rhino 1.7.15, and support Template Literals as a concatenation/addition of strings. Also includes a few tests for Template Literals in "es6.js" file. Tagged Template Literals are not yet supported.


cast/js/rhino/src/main/java/com/ibm/wala/cast/js/translator/RhinoToAstTranslator.java
```diff
@@ -1596,6 +1600,40 @@ public CAstNode visitSymbol(Symbol node, WalkContext arg) {
       return null;
     }
 
+    @Override
+    public CAstNode visitTemplateCharacters(TemplateCharacters node, WalkContext arg) {
+      return Ast.makeConstant(node.getValue());
+    }
+
+    @Override
+    public CAstNode visitTemplateLiteral(TemplateLiteral node, WalkContext arg) {
```
msridhar, line 1609: I can't really understand the logic here; I think we need some comments to explain.  Is the code in this method handling just the string concatenation?

cast/js/rhino/src/main/java/com/ibm/wala/cast/js/translator/RhinoToAstTranslator.java
```diff
@@ -1596,6 +1600,40 @@ public CAstNode visitSymbol(Symbol node, WalkContext arg) {
...
+                lastBinaryNode);
+      }
+      return lastBinaryNode;
+    }
+
+    @Override
+    public CAstNode visitTaggedTemplateLiteral(TaggedTemplateLiteral node, WalkContext arg) {
+      // TODO Auto-generated method stub
+      return null;
+    }
```
msridhar, line 1635: I'm concerned this will cause a crash if we get an input with a tagged template literal.  If we don't handle this case yet, maybe we should not add this method and just skip visiting the nodes?

cast/js/rhino/src/main/java/com/ibm/wala/cast/js/translator/TypedNodeVisitor.java
```diff
@@ -158,6 +161,12 @@ public R visit(AstNode node, A arg) {
       return visitSwitchCase((SwitchCase) node, arg);
     } else if (node instanceof SwitchStatement) {
       return visitSwitchStatement((SwitchStatement) node, arg);
+    } else if (node instanceof TaggedTemplateLiteral) {
+      return visitTaggedTemplateLiteral((TaggedTemplateLiteral) node, arg);
```
msridhar, line 165: See comment above about not visiting these nodes if we don't yet handle them; or is the issue that we'll get an unexpected node type?

In general, what happens right now when we try to pass an ES6 input to WALA, given that we haven't set the language version?  Do we just get a parse error in Rhino?

cast/js/src/testFixtures/java/com/ibm/wala/cast/js/test/TestSimpleCallGraphShape.java
```diff
@@ -1071,4 +1071,38 @@ public void testComplexFinally()
 
     CAstCallGraphUtil.dumpCG(B.getCFAContextInterpreter(), B.getPointerAnalysis(), CG);
   }
+
+  private static final Object[][] assertionsES6Args =
```
msridhar, line 1075: For the test cases where the template substitution has a function call, can we assert that the appropriate call edge(s) are present?

msridhar (commented):
Thank you for this contribution!  I have a few comments and questions.

liblit:
There has been no new discussion about this PR since April 27, 2025. @apple5775 and @msridhar, do you still expect to reach consensus that this PR is good and ready to merge? Or has this PR been abandoned, in which case we should just close it?

msridhar (commented):
Did another pass and added a couple more comments.  Also, some of my previous comments have not yet been addressed.  Thanks!

liblit:
@apple5775, do you still intend to address @msridhar's comments and submit an updated PR for review? Or should we consider this PR to be abandoned and therefore close it with no further action?

PR #1499, opened 2025-04-26
Correct assertEquals argument order

liblit:
JUnit 5's `assertEquals` method should be called with the _expected_ value _first_ and the _actual_ value _second_.


msridhar (approved):
👍

PR #1504, opened 2025-05-05
Standardize on AssertJ

liblit:
Previously, test assertions used a mix of JUnit 5, Hamcrest, and AssertJ.  Now we standardize on AssertJ, including a few WALA-specific AssertJ assertions and conditions.  JUnit 5 is still our underlying test engine, and Hamcrest is no longer used.

---

# Advice to Reviewer

This is a _big_ set of changes. To keep your place, you may find it useful to [mark individual files as reviewed](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request#about-reviewing-pull-requests) when you've looked at them.

If you see something that you think could have been done better, there's a good chance that the same not-quite-ideal pattern appears elsewhere. Please bring any nits to my attention as you find them, so that I can systematically improve all similar instances and update this pull request. That way you won't waste your time spotting similar nits over and over again.

The two ideas above work well together. If you have marked a file as reviewed, and then I update the pull request, your "already reviewed" marks will be _retained_ for any files that didn't change.

### DIFF
--- a/cast/java/src/testFixtures/java/com/ibm/wala/cast/java/test/IRTests.java
+++ b/cast/java/src/testFixtures/java/com/ibm/wala/cast/java/test/IRTests.java
@@ -431,9 +431,7 @@ public abstract class IRTests {
     }
 
     if (assertReachable) {
-      assertThat(unreachable)
-          .withFailMessage(() -> "unreachable methods: " + unreachable)
-          .isEmpty();
+      assertThat(unreachable).isEmpty();
     }
   }
 
@@ -490,12 +488,12 @@ public abstract class IRTests {
         }
       }
     }
-    assertThat(foundLib).withFailMessage(() -> "couldn't find library file from " + libs).isTrue();
+    assertThat(foundLib).isTrue();
 
     for (String srcFilePath : sources) {
       String srcFileName = srcFilePath.substring(srcFilePath.lastIndexOf(File.separator) + 1);
       File f = new File(srcFilePath);
-      assertThat(f).withFailMessage(() -> "couldn't find " + srcFilePath).exists();
+      assertThat(f).exists();
       if (f.isDirectory()) {
         engine.addSourceModule(new SourceDirectoryTreeModule(f));
       } else {
--- a/cast/java/src/testFixtures/java/com/ibm/wala/cast/java/test/JavaIRTests.java
+++ b/cast/java/src/testFixtures/java/com/ibm/wala/cast/java/test/JavaIRTests.java
@@ -90,11 +90,9 @@ public abstract class JavaIRTests extends IRTests {
                 findOrCreateTypeReference("Source", typeStr, cg.getClassHierarchy());
 
             final IClass iClass = cg.getClassHierarchy().lookupClass(type);
-            assertThat(iClass).withFailMessage(() -> "Could not find class " + typeStr).isNotNull();
+            assertThat(iClass).isNotNull();
 
-            assertThat(iClass)
-                .withFailMessage("Expected IFoo to be an interface.")
-                .matches(IClass::isInterface);
+            assertThat(iClass).matches(IClass::isInterface);
           },
           cg -> {
             final String typeStr = "FooIT1";
@@ -103,14 +101,13 @@ public abstract class JavaIRTests extends IRTests {
                 findOrCreateTypeReference("Source", typeStr, cg.getClassHierarchy());
 
             final IClass iClass = cg.getClassHierarchy().lookupClass(type);
-            assertThat(iClass).withFailMessage(() -> "Could not find class " + typeStr).isNotNull();
+            assertThat(iClass).isNotNull();
 
             final Collection<? extends IClass> interfaces = iClass.getDirectInterfaces();
 
-            assertThat(interfaces).withFailMessage("Expected one single interface.").hasSize(1);
+            assertThat(interfaces).hasSize(1);
 
             Assertions.<IClass>assertThatCollection(interfaces)
-                .withFailMessage("Expected Foo to implement IFoo")
                 .contains(
                     cg.getClassHierarchy()
                         .lookupClass(
@@ -126,9 +123,7 @@ public abstract class JavaIRTests extends IRTests {
                 findOrCreateTypeReference("Source", typeStr, cg.getClassHierarchy());
 
             final IClass derivedClass = cg.getClassHierarchy().lookupClass(type);
-            assertThat(derivedClass)
-                .withFailMessage(() -> "Could not find class " + typeStr)
-                .isNotNull();
+            assertThat(derivedClass).isNotNull();
 
             final TypeReference baseType =
                 findOrCreateTypeReference("Source", "Base", cg.getClassHierarchy());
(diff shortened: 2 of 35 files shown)


cast/java/src/testFixtures/java/com/ibm/wala/cast/java/test/IRTests.java
```diff
@@ -333,7 +335,7 @@ public void check(CallGraph cg) {
               }
             }
 
-            assertFalse(false, "cannot find " + at);
+            Assertions.fail("cannot find " + at);
```
msridhar, line 338: Just `fail(...)` due to static import?

cast/java/src/testFixtures/java/com/ibm/wala/cast/java/test/IRTests.java
```diff
@@ -234,9 +233,12 @@ public void check(CallGraph cg) {
       MethodReference mref = descriptorToMethodRef(method, cg.getClassHierarchy());
 
       for (CGNode cgNode : cg.getNodes(mref)) {
-        assertTrue(
-            this.check(cgNode.getMethod(), cgNode.getIR()),
-            "failed for " + this.variableName + " in " + cgNode + "\n" + cgNode.getIR());
+        assertThat(this)
+            .withFailMessage(
```
msridhar, line 237: is there a reason to call `withFailMessage` before doing the actual assertion?  I find it a bit hard to read

msridhar (commented):
couple of quick initial comments.  the overall review may take some time.

cast/java/src/testFixtures/java/com/ibm/wala/cast/java/test/IRTests.java
liblit: Good catch. I'll change that in the next update to this PR. I looked for other instances of `Assertions.fail` but found none.

cast/java/src/testFixtures/java/com/ibm/wala/cast/java/test/IRTests.java
liblit: If the actual assertion precedes `withFailMessage`, then a failure of that assertion will raise an exception before `withFailMessage` has a chance to customize the message. Thus, [`withFailMessage` must be called _before_ an assertion to have any effect](https://assertj.github.io/doc/#calling-withfailmessageoverridingerrormessage-after-the-assertion).

Truth be told, I have rarely used `withFailMessage` at all in other projects. I generally find the AssertJ-generated failure messages to be fine, especially when I use specialized assertions like `hasSize` or `contains`. I've occasionally [used `as` to improve the description of the assertion itself or of the value being checked](https://assertj.github.io/doc/#assertj-core-assertion-description). But I've rarely felt a need for `withFailMessage`.

So why do I have so many of them in this PR? Because the pre-existing code made heavy use of the last `String` argument to various JUnit 5 assertion methods. That last argument serves the same role: it customizes the message used in case of failure. I didn't want to throw those all away, because apparently _someone_ thought them useful enough to add in the first place.

(Incidentally, [`as` must precede any actual assertions too](https://assertj.github.io/doc/#calling-as-after-the-assertion).)

cast/java/src/testFixtures/java/com/ibm/wala/cast/java/test/IRTests.java
```diff
@@ -492,12 +495,12 @@ public static void populateScope(
...
       }
     }
-    assert foundLib : "couldn't find library file from " + libs;
+    assertThat(foundLib).withFailMessage(() -> "couldn't find library file from " + libs).isTrue();
 
     for (String srcFilePath : sources) {
       String srcFileName = srcFilePath.substring(srcFilePath.lastIndexOf(File.separator) + 1);
       File f = new File(srcFilePath);
-      assertTrue(f.exists(), "couldn't find " + srcFilePath);
+      assertThat(f).withFailMessage(() -> "couldn't find " + srcFilePath).exists();
```
msridhar, line 503: I'm really finding the readability to be reduced with all these `withFailMessage()` calls.  They kind of defeat the whole readability advantage of the fluent interface.  I'm leaning towards just getting rid of them and adding back custom messages where we need them, or possibly splitting tests later if they are asserting too many things.

msridhar (commented):
@liblit see comment below about the `withFailMessage` calls; what do you think about just getting rid of them?

liblit:
> @liblit see comment below about the `withFailMessage` calls; what do you think about just getting rid of them?

Sorry for leaving that comment hanging. I thought I replied, but I don't see my reply anywhere.

I kept the `withFailMessage` calls because the code I was replacing often provided a custom failure message. _Someone_ apparently thought those were useful enough to create them in the first place, and I didn't want to throw away that prior investment of effort.

That being said, AssertJ will do a better job than JUnit of explaining assertion failures, simply because it has a richer assertion structure and vocabulary to work with. So I can believe that most of these custom failure messages are no longer needed. I am happy to remove all `withFailMessage` calls, per @msridhar's request.

I'll do that in a separate commit that will be included in this existing PR. That way we have a clear distinction between the work done to move to AssertJ and the work done to remove those custom failure messages. It may take me a few days to get around to that, but I'll get it done when I can.

msridhar:
Sorry I was confusing. You didn't leave anything hanging! My comment was misleading. 

In agreement with your plan to remove the `withFailMessage` calls in a separate commit on this PR. No hurry.

cast/java/src/testFixtures/java/com/ibm/wala/cast/java/test/IRTests.java
liblit: d333b13108d4f1ca7bd256d4f97bbab67dcec8ba removes all `withFailMessage` calls.

core/src/test/java/com/ibm/wala/core/tests/ir/CFGTest.java
```diff
@@ -151,16 +156,19 @@ public void testSync3() {
...
 
   public static void testCFG(SSACFG cfg, int[][] assertions) {
-    for (int i = 0; i < assertions.length; i++) {
-      SSACFG.BasicBlock bb = cfg.getNode(i);
-      assertEquals(assertions[i].length, cfg.getSuccNodeCount(bb), "basic block " + i);
-      for (int j = 0; j < assertions[i].length; j++) {
-        assertTrue(cfg.hasEdge(bb, cfg.getNode(assertions[i][j])));
-      }
-    }
+    IntStream.range(0, assertions.length)
```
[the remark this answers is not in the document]
liblit: That change allowed each successive value of `i` to be effectively final, which thereby allowed me to use it in a `withFailMessage` lambda. Now that we've removed `withFailMessage`, the change away from a `for` loop is no longer needed. I'll revert to the `for` loop.

Thanks for spotting this change and asking about it!

cast/java/src/testFixtures/java/com/ibm/wala/cast/java/test/IRTests.java
```diff
@@ -234,9 +232,7 @@ public void check(CallGraph cg) {
       MethodReference mref = descriptorToMethodRef(method, cg.getClassHierarchy());
 
       for (CGNode cgNode : cg.getNodes(mref)) {
-        assertTrue(
-            this.check(cgNode.getMethod(), cgNode.getIR()),
-            "failed for " + this.variableName + " in " + cgNode + "\n" + cgNode.getIR());
+        assertThat(this).matches(tests -> tests.check(cgNode.getMethod(), cgNode.getIR()));
```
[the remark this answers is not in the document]
liblit: I thought that we'd get a better failure message by separating the value under test from the test applied to it. But actually, neither particularly good. Using `match` we get:

```
java.lang.AssertionError: 
Expecting actual:
  com.ibm.wala.cast.java.test.IRTests$SourceMapAssertion@6ceb710
to match given predicate.
```

Using `isTrue` we get:

```
org.opentest4j.AssertionFailedError: 
Expecting value to be true but was false
```

If `this.toString()` were more informative, then we'd see a bigger benefit to using `assertThat(this).matches(…)`. As is, it probably makes no difference. I'll change to the simpler `isTrue` style.

msridhar (approved):
Thanks for the fixes and for this great change!

liblit:
Thanks for your effort to review this jumbo-sized change, @msridhar!

PR #1509, opened 2025-05-15
Add aggregated code coverage report task

msridhar:
With this change, we can run `./gradlew testCodeCoverageReport` to get an aggregated test coverage report across all the WALA projects.  Once this lands, we can look into using [Codecov](https://about.codecov.io) to show differences in coverage for PRs.  Based on the sample [here](https://docs.gradle.org/current/samples/sample_jvm_multi_project_with_code_coverage_standalone.html).


code-coverage-report/build.gradle.kts
```diff
@@ -0,0 +1,26 @@
+plugins {
+  id("com.ibm.wala.gradle.eclipse-maven-central")
```
msridhar, line 2: We seem to need to pull in all the Eclipse dependencies or otherwise this doesn't work with Eclipse-related projects

code-coverage-report/build.gradle.kts
```diff
@@ -0,0 +1,26 @@
...
+  id("com.ibm.wala.gradle.java")
+  id("com.ibm.wala.gradle.maven-eclipse-jsdt")
+  id("jacoco-report-aggregation")
+}
+
+dependencies {
+  implementation(projects.cast.java)
```
msridhar, line 9: @liblit do you know a better way to do this?  I want to add an `implementation` dependence on all projects (including nested subprojects) that use the Jacoco plugin.  But I couldn't figure out how to do that without listing them all out.

liblit: I don't immediately know a better way. But I _might_ be able to find one. I propose that we proceed with this pull request in its current form. In addition, please create an issue requesting a better way, and assign that issue to me as a reminder.

build-logic/src/main/kotlin/com/ibm/wala/gradle/java.gradle.kts
```diff
@@ -171,3 +174,4 @@ spotless {
 tasks.named("spotlessJava") {
   onlyIf { JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_17) }
 }
+
```
liblit, line 177: Spurrious?

code-coverage-report/build.gradle.kts
liblit: By the way, [this documentation](https://docs.gradle.org/current/userguide/jacoco_report_aggregation_plugin.html#sec:dependency_management) and [this example](https://docs.gradle.org/current/samples/sample_jvm_multi_project_with_code_coverage_standalone.html) refer to using a `jacocoAggregation` dependency configuration defined by the `jacoco-report-aggregation` plugin, not the `java` plugin's `implementation` dependency configuration as you've done here.

msridhar: I think this example is more like what I was looking for!  I might try a rewrite to be more like this one

msridhar: Re-worked to use `jacocoAggregation` which is much better.  Thanks for the suggestion!

liblit (approved):
Nice infrastructure improvement, @msridhar! Thanks for working on this change.

It’s a remarkably small change, too. That `jacoco-report-aggregation` plugin is doing quite a bit of heavy lifting for us. Sweet!

msridhar:
> Nice infrastructure improvement, @msridhar! Thanks for working on this change.
> 
> It’s a remarkably small change, too. That `jacoco-report-aggregation` plugin is doing quite a bit of heavy lifting for us. Sweet!

Thanks!  And yeah, I was surprised how easy this is now.  I've done it on other projects and before `jacoco-report-aggregation` it was a lot more painful.

code-coverage-report/build.gradle.kts
liblit: I wrote:

> I don't immediately know a better way. But I _might_ be able to find one.

Better way found in #1513.

PR #1513, opened 2025-05-19
Simplify JaCoCo report aggregation

liblit:
When aggregating, automatically include every subproject that uses the `jacoco` plugin.  The logic for doing this works out quite nicely thanks to `PluginContainer::withId`.  That method executes an action immediately if a project has already loaded the plugin of interest, _or_ registers the action to run later if the project loads the plugin of interest later.  Thus, everything works as intended regardless of the order in which subprojects are configured.

### DIFF
--- a/code-coverage-report/build.gradle.kts
+++ b/code-coverage-report/build.gradle.kts
@@ -6,21 +6,5 @@ plugins {
 }
 
 dependencies {
-  jacocoAggregation(projects.cast.java)
-  jacocoAggregation(projects.cast.java.ecj)
-  jacocoAggregation(projects.cast.js)
-  jacocoAggregation(projects.cast.js.html.nuValidator)
-  jacocoAggregation(projects.cast.js.nodejs)
-  jacocoAggregation(projects.cast.js.rhino)
-  jacocoAggregation(projects.core)
-  jacocoAggregation(projects.dalvik)
-  jacocoAggregation(projects.ide)
-  jacocoAggregation(projects.ide.jdt)
-  jacocoAggregation(projects.ide.jdt.test)
-  jacocoAggregation(projects.ide.jsdt)
-  jacocoAggregation(projects.ide.jsdt.tests)
-  jacocoAggregation(projects.ide.tests)
-  jacocoAggregation(projects.scandroid)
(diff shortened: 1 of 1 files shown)


code-coverage-report/build.gradle.kts
```diff
@@ -6,21 +6,5 @@ plugins {
...
-  jacocoAggregation(projects.ide)
-  jacocoAggregation(projects.ide.jdt)
-  jacocoAggregation(projects.ide.jdt.test)
-  jacocoAggregation(projects.ide.jsdt)
-  jacocoAggregation(projects.ide.jsdt.tests)
-  jacocoAggregation(projects.ide.tests)
-  jacocoAggregation(projects.scandroid)
-  jacocoAggregation(projects.shrike)
-  jacocoAggregation(projects.util)
+  rootProject.subprojects { plugins.withId("jacoco") { jacocoAggregation(this@subprojects) } }
```
msridhar, line 9: Coincidentally I think I got the Codecov integration working earlier today!  Its results look ok, but did you happen to check if `subprojects` covers projects nested at arbitrary depths, like `cast/js/rhino`?

liblit: > Coincidentally I think I got the Codecov integration working earlier today!

W00t! 👏

> did you happen to check if `subprojects` covers projects nested at arbitrary depths, like `cast/js/rhino`?

Not specifically. However, I generated the aggregated coverage report before and after this change, and found them to be identical.

msridhar (approved):
Great cleanup!  FWIW ChatGPT utterly failed to come up with something this nice when I asked it to help me try to do this 🙂

code-coverage-report/build.gradle.kts
liblit: Hmm, let's just place it safe: I'll replace `rootProject.subprojects` with `rootProject.allprojects`. Then there should be no doubt as to whether we got everything.

liblit:
> Great cleanup!

Thanks!

> FWIW ChatGPT utterly failed to come up with something this nice when I asked it to help me try to do this 🙂

That's both disappointing and reassuring. 🤖 😄

PR #1514, opened 2025-05-19
Aggregate coverage from nested subprojects

liblit:
Previously we might only have been aggregating coverage from the root project's immediate subprojects. Now it should be clear that we're considering all projects regardless of nesting depth.

### DIFF
--- a/code-coverage-report/build.gradle.kts
+++ b/code-coverage-report/build.gradle.kts
@@ -6,5 +6,5 @@ plugins {
 }
 
 dependencies {
-  rootProject.allprojects { plugins.withId("jacoco") { jacocoAggregation(this@subprojects) } }
+  rootProject.allprojects { plugins.withId("jacoco") { jacocoAggregation(this@allprojects) } }
 }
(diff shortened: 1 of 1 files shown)


msridhar:
@liblit FWIW with your previous approach I ran `./gradlew testCodeCoverageReport --dry-run` and I still see the deeply nested projects listed, e.g., I see `:cast:js:html:nu_validator:test SKIPPED`.  So maybe that's decent evidence that using `subprojects` is ok?

liblit:
> @liblit FWIW with your previous approach I ran `./gradlew testCodeCoverageReport --dry-run` and I still see the deeply nested projects listed, e.g., I see `:cast:js:html:nu_validator:test SKIPPED`. So maybe that's decent evidence that using `subprojects` is ok?

Maybe. Thanks for checking! But I don't like that I don't know _why_ that works. I'm going to proceed with changing to `allprojects` so that it will be _obviously_ correct.

msridhar (approved):

PR #1518, opened 2025-06-11
Update JavaCompileUsingEcj.kt

Obielicious:
Builds on Windows assuming there is a C++ compiler

Thank you for contributing to WALA!  Please see the contribution guidelines at https://github.com/wala/WALA/blob/master/CONTRIBUTING.md for information on code style and general guidelines for pull requests.

### DIFF
--- a/build-logic/src/main/kotlin/com/ibm/wala/gradle/JavaCompileUsingEcj.kt
+++ b/build-logic/src/main/kotlin/com/ibm/wala/gradle/JavaCompileUsingEcj.kt
@@ -54,7 +54,7 @@ abstract class JavaCompileUsingEcj : JavaCompile() {
             "-properties",
             jdtPrefs.toString(),
             "-classpath",
-            this@JavaCompileUsingEcj.classpath.joinToString(":"),
+            this@JavaCompileUsingEcj.classpath.joinToString(File.pathSeparator),
             "-d",
             destinationDirectory.get().toString())
       }
@@ -64,10 +64,18 @@ abstract class JavaCompileUsingEcj : JavaCompile() {
 
   @TaskAction
   fun compile() {
+    val testArgs = options.allCompilerArgs
+    val argFile = createTempFile(prefix = "kotlinTemp", suffix = ".tmp")
+    val writer = java.io.PrintWriter(argFile)
+    for (testArg in testArgs) {
+      writer.print(testArg + " ")
+    }
+    writer.close()
+
     execOperations.javaexec {
       classpath(ecjJar.absolutePath)
       executable(javaLauncherPath.get())
-      args(options.allCompilerArgs)
+      args("@" + argFile)
     }
   }
 


build-logic/src/main/kotlin/com/ibm/wala/gradle/JavaCompileUsingEcj.kt
```diff
@@ -55,7 +55,7 @@ abstract class JavaCompileUsingEcj : JavaCompile() {
             "-properties",
             jdtPrefs.toString(),
             "-classpath",
-            this@JavaCompileUsingEcj.classpath.joinToString(":"),
+            this@JavaCompileUsingEcj.classpath.joinToString(File.pathSeparator),
```
liblit, line 58: Attention @msridhar: this pull request implies that ECJ builds on Windows were not previously working, but I think we didn't know that. Should we be enabling ECJ builds on Windows so that we will be aware of any future regressions to that capability? Should we enable ECJ builds on macOS too?

Or conversely, do we neither care about nor support ECJ builds on Windows, in which case this whole pull request should be discarded?

build-logic/src/main/kotlin/com/ibm/wala/gradle/JavaCompileUsingEcj.kt
```diff
@@ -65,10 +65,18 @@ abstract class JavaCompileUsingEcj : JavaCompile() {
 
   @TaskAction
   protected override fun compile(inputs: InputChanges) {
+    val testArgs = options.allCompilerArgs
+    val argFile = createTempFile(prefix = "kotlinTemp", suffix = ".tmp")
```
liblit, line 69: `createTempFile` is deprecated. Use `kotlin.io.path.createTempFile` instead.

build-logic/src/main/kotlin/com/ibm/wala/gradle/JavaCompileUsingEcj.kt
```diff
@@ -65,10 +65,18 @@ abstract class JavaCompileUsingEcj : JavaCompile() {
...
+    val writer = java.io.PrintWriter(argFile)
```
liblit, line 70: Use Kotlin's `File.printWriter` extension method instead of calling `PrintWriter` constructor directly.

build-logic/src/main/kotlin/com/ibm/wala/gradle/JavaCompileUsingEcj.kt
```diff
@@ -65,10 +65,18 @@ abstract class JavaCompileUsingEcj : JavaCompile() {
...
+    for (testArg in testArgs) {
+      writer.print(testArg + " ")
```
liblit, line 72: In my opinion, this file would be more readable (e.g., when debugging) if arguments were separated by line breaks instead of spaces. If we change `print` to `println` and remove the `+ " "`, does that still work correctly on all supported platforms? Or does something require that we use spaces instead of newlines?

build-logic/src/main/kotlin/com/ibm/wala/gradle/JavaCompileUsingEcj.kt
```diff
@@ -65,10 +65,18 @@ abstract class JavaCompileUsingEcj : JavaCompile() {
...
+    }
+    writer.close()
```
liblit, line 74: Use Kotlin's `use { ... }` extension on the `PrintWriter` instance instead of an explicit `close()`.

build-logic/src/main/kotlin/com/ibm/wala/gradle/JavaCompileUsingEcj.kt
liblit, line 69: Will something delete this temporary file when compilation is done? Does that happen implicitly? Do we need to register something that will delete this file when the JVM that is running Gradle terminates?

liblit (changes requested):
Thanks for proposing these changes! I have a few refinements to suggest, most of which involve making your improvements more Kotlin-idiomatic.

build-logic/src/main/kotlin/com/ibm/wala/gradle/JavaCompileUsingEcj.kt
juliandolby: We actually need Windows builds here at IBM, since we have product usage of WALA on Windows (_not_ mine!). So please can we keep support for building it all on Windows?

build-logic/src/main/kotlin/com/ibm/wala/gradle/JavaCompileUsingEcj.kt
juliandolby: Can we use that call with the same arguments as `createTempFile`.  I thought this was a bad thing to use, but the documentation I found online was not too helpful.

juliandolby: hmm...  i guess i assumed that temp files get deleted on exit,  am pretty sure they do in Java.

build-logic/src/main/kotlin/com/ibm/wala/gradle/JavaCompileUsingEcj.kt
liblit: OK, we're all agreed: CI will include ECJ builds on all supported platforms. @juliandolby : it's OK with me if you want to leave this alone for now. Once the current pull request is merged, I'll create a separate pull request to enable ECJ builds everywhere.

build-logic/src/main/kotlin/com/ibm/wala/gradle/JavaCompileUsingEcj.kt
liblit: Yes, you can use the same arguments. So what's the difference?

* `kotlin.io.createTempFile` is implemented using `java.io.File.createTempFile`, which is sloppy about permissions on the temporary file.
* `kotlin.io.path.createTempFile` is implemented using `java.nio.file.Files.createTempFile`, which is more careful about permissions on the temporary file.

liblit: Neither Kotlin nor Java deletes these temporary files on JVM exit.

`kotlin.io.createTempFile` uses `java.io.File.createTempFile`, whose [documentation](https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/io/File.html#createTempFile(java.lang.String,java.lang.String,java.io.File)) states:

> This method provides only part of a temporary-file facility. To arrange for a file created by this method to be deleted automatically, use the [`deleteOnExit()`](https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/io/File.html#deleteOnExit()) method.

`kotlin.io.path.createTempFile` uses `java.nio.file.Files.createTempFile`, whose [documentation](https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/nio/file/Files.html#createTempFile(java.nio.file.Path,java.lang.String,java.lang.String,java.nio.file.attribute.FileAttribute...)) states:

> As with the `File.createTempFil`e methods, this method is only part of a temporary-file facility. Where used as a work file, the resulting file may be opened using the [`DELETE_ON_CLOSE`](https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/nio/file/StandardOpenOption.html#DELETE_ON_CLOSE) option so that the file is deleted when the appropriate close method is invoked. Alternatively, a [shutdown-hook](https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/lang/Runtime.html#addShutdownHook(java.lang.Thread)), or the [`File.deleteOnExit()`](https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/io/File.html#deleteOnExit()) mechanism may be used to delete the file automatically.

build-logic/src/main/kotlin/com/ibm/wala/gradle/JavaCompileUsingEcj.kt
juliandolby: i tried a test, and it seems to want the arguments on a single line

liblit: OK, thanks for giving that a try. I'm satisfied with this code change as originally proposed.

build-logic/src/main/kotlin/com/ibm/wala/gradle/JavaCompileUsingEcj.kt
liblit: The revised code's use of `use { ... }` looks good. Thanks!

build-logic/src/main/kotlin/com/ibm/wala/gradle/JavaCompileUsingEcj.kt
```diff
@@ -69,10 +70,19 @@ abstract class JavaCompileUsingEcj : JavaCompile() {
...
+      for (testArg in testArgs) {
+        writer.print(testArg + " ")
+      }
+    }
+
     execOperations.javaexec {
       classpath(ecjJar)
       executable(javaLauncherPath.get())
-      args(options.allCompilerArgs)
+      args("@" + f)
```
liblit, line 85: Please consider the following further improvements to

* use a better name than `f` and
* prefer string templates over concatenations.

```suggestion
      args("@$moreArgsFile")
```

liblit (approved):
I think I can improve this code further, but let's not let perfect be the enemy of good. I'll approve this PR in its current form. If I think I can make it even nicer, then I'll propose those changes separately.

PR #1532, opened 2025-06-21
Require that Gradle run in a Java 17+ JVM

### LINKED ISSUE #1530: Require that Gradle use a Java 17+ JVM
Gradle 9 must run in a Java 17+ JVM. So let's make that an explicit requirement in anticipation of a future Gradle 9 upgrade.

Note that we will still be able to publish bytecode that is compatible with Java 11.

liblit:
We still publish bytecode that is compatible with Java 11. But Gradle 9 must run in a Java 17+ JVM. So let's make that an explicit requirement in anticipation of a future Gradle 9 upgrade.

Fixes #1530.


msridhar (approved):
@liblit are you able to update the required CI job names?

liblit:
> @liblit are you able to update the required CI job names?

Done.

PR #1536, opened 2025-06-23
Treat all ad-hoc downloads as dependencies

liblit:
For each ad-hoc download, we now create a single-dependency configuration and declare an exclusive-content Ivy repository to resolve that dependency.  This strategy moves downloads out of the local build tree and instead retains them in the user's Gradle cache under `~/.gradle`.

This approach has pros and cons:

* **Pros**

   * Downloads do not change from build to build, and rarely change even over the span of months or years.  By retaining these downloads in the user's Gradle cache under `~/.gradle` cache, we amortize the download time across all of the user's WALA trees.

   * Our automated GitHub actions [cache downloaded dependencies](https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#which-content-is-cached). If that works well, then it should amortize the download time across multiple GitHub actions runs.  This _might_ also make our actions more resilient in the face of transient network outages on GitHub action runners.  We've been hitting those outages a _lot_ lately, and it has been a real drag on development.

* **Cons**

   * Gradle downloads most dependencies that the task graph will need before starting the tasks themselves:  even the very first task must wait until most dependencies are downloaded.  Additionally, while downloads can be concurrent, I see less concurrency than that we achieve when evaluating the task graph.  In practice, many of the largest downloads are happening with no other concurrent activity.  That leaves a lot of potential parallelism on the table when these downloads are not already in the user's Gradle cache.

I think this change is worth giving a try, to see whether the advantages outweigh the disadvantages in practice.


msridhar (approved):

msridhar:
@liblit I'm seeing a large number of seemingly serialized downloads just when loading WALA into IntelliJ now, I think at configuration time?  Is this expected?  It seems to be more than just the ad-hoc downloads, but possibly it's happening that way because I'm on a slower connection.  It's a bit of a bummer if all the downloads need to happen the first time a user loads WALA into the IDE / does any configuration, though perhaps it's better overall.

liblit:
> @liblit I'm seeing a large number of seemingly serialized downloads just when loading WALA into IntelliJ now

When IntelliJ IDEA imports a Gradle project, it uses Gradle APIs to build a fairly rich model of the build structure. It is expected that IntelliJ IDEA triggers more project configuration activity than you would get simply by running a command line `./gradlew tasks`.

That being said, the way we write the build logic can influence how much work happens on import. There _might_ be places where dependency resolution is happening earlier than necessary, and there _might_ be things we can do to mitigate that. **I am looking into some ideas for this right now.** However, any tweaks I have in mind would affect the ad-hoc downloads only, not downloads arising from standard dependencies and plugins.

As for the fact that these downloads are _sequential_, I've noticed that too. It seems to be specific to IntelliJ IDEA: we get much better download parallelism when running tasks from the command like, such as `./gradlew build`. I think that this must have something to do with Gradle’s tooling API or how IntelliJ IDEA uses that API. I have not been able to find any existing bug reports describing this one-download-at-a-time behavior, though.

liblit:
I wrote:
> I am looking into some ideas for [download procrastination] right now.

#1541 implements my ideas, avoiding 101 MB of downloads during initial IntelliJ IDEA import.

PR #1539, opened 2025-06-27
Remove :ide:jdt:test

### LINKED ISSUE #1538: Remove or reclassify test sources in `ide/jdt/test/src`
The Java sources under [`ide/jdt/test/src`](https://github.com/wala/WALA/tree/master/ide/jdt/test/src) define five classes:

1. `com.ibm.wala.cast.java.jdt.test.Activator`
1. `com.ibm.wala.cast.java.test.JDTJavaTest`
1. `com.ibm.wala.cast.java.test.TestPlugin`
1. `com.ibm.wala.cast.java.test.TypeInferenceAssertion`
1. `com.ibm.wala.demandpa.driver.DemandCastChecker`

I cannot find any other code in the WALA repository that references any of these five classes. Furthermore, none of these file classes defines any `@Test`-annotated tests. Thus, this all seems like dead code to me, which would make `:ide:jdt:test` a dead Gradle subproject.

Are these classes used by something else? Did I miss a use of these classes elsewhere inside the main WALA project? Are they used by some other WALA-adjacent project?

Gradle 9 will complain about `test` source sets that do not actually define any tests. Thus, in order to upgrade to Gradle 9, we need to to _something_ about this code. Maybe it should be reclassified as test-subject or test-fixture code. Or maybe it should be removed entirely.

liblit:
The sources in this subproject define no tests and are not used by any other code in this main WALA repository.  As far as I can tell, that code is all dead code, making this a dead subproject.

Fixes #1538.


ide/jdt/test/src/test/java/com/ibm/wala/cast/java/jdt/test/Activator.java
```diff
@@ -1,85 +0,0 @@
...
- * IS". REGENTS HAS NO OBLIGATION TO PROVIDE MAINTENANCE, SUPPORT,
- * UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
- */
-package com.ibm.wala.cast.java.jdt.test;
-
-import org.eclipse.core.runtime.Plugin;
-import org.osgi.framework.BundleContext;
-
-/** The activator class controls the plug-in life cycle */
-public class Activator extends Plugin {
```
msridhar, line 44: This might be related to running something as an Eclipse plug-in, but I thought we got rid of that functionality?  @khatchad do you rely on this being here at all?

ide/jdt/test/src/test/java/com/ibm/wala/cast/java/test/JDTJavaTest.java
```diff
@@ -1,82 +0,0 @@
...
-import com.ibm.wala.ipa.callgraph.impl.Util;
-import com.ibm.wala.ipa.callgraph.propagation.InstanceKey;
-import com.ibm.wala.ipa.cha.IClassHierarchy;
-import com.ibm.wala.util.io.TemporaryFile;
-import java.io.File;
-import java.io.IOException;
-import java.util.Collection;
-import java.util.List;
-
-public abstract class JDTJavaTest extends IRTests {
```
msridhar, line 31: I think this used to be used when we were actually running Eclipse Plugin tests by loading them under Maven with Tycho.  But we don't do that anymore, so I think it's ok to delete this one.

ide/jdt/test/src/test/java/com/ibm/wala/demandpa/driver/DemandCastChecker.java
```diff
@@ -1,301 +0,0 @@
...
-import java.util.List;
-import java.util.Properties;
-import java.util.function.Predicate;
-
-/**
- * Uses a demand-driven points-to analysis to check the safety of downcasts.
- *
- * @author Manu Sridharan
- */
-public class DemandCastChecker {
```
msridhar, line 94: This code I would like to keep around as it's a useful example.  Could it be moved to `core/src/main/java/com/ibm/wala/examples/drivers`?

msridhar (commented):
Thanks for this cleanup, @liblit!  Couple comments but mostly I'm fine with this

ide/jdt/test/src/test/java/com/ibm/wala/demandpa/driver/DemandCastChecker.java
liblit: Retained and moved in the latest version (c359fb1ef6a541e2373967607cf6abf064da86e7) of this pull request.

liblit:
The `codecov/project` check is failing due to a slight reduction (-0.08%) in test coverage. However, I believe that this change is benign. It does not actually reflect a loss of test coverage in any code that was previously covered.

This change appears to be due to moving `DemandCastChecker.java` into `core/src/main/java/com/ibm/wala/examples/drivers`, per @msridhar's [request](https://github.com/wala/WALA/pull/1539#discussion_r2173071798). [That file contains 120 tracked lines](https://app.codecov.io/gh/wala/WALA/pull/1539/tree/core/src/main/java/com/ibm/wala/examples/drivers?dropdown=coverage), i.e., 120 lines that contain code. However, none of these lines are covered by any tests. Previously, that didn't affect our coverage stats because previously this file was itself assumed to be test code. So these 120 lines have gone from not being counted at all to being counted as not-covered.

msridhar (approved):

PR #1541, opened 2025-06-28
Delay some downloads until actually needed

liblit:
Before this change, several project activities could trigger downloads of `ecj`, `java-cup`, and `ocamljava`.  For example, this would happen when running `./gradlew tasks` or when importing the WALA project into IntelliJ IDEA.  Those three downloads pulled in 101 MB of cacheable data, 95 MB of which was for `ocamljava` alone.  As [discussed elsewhere](https://github.com/wala/WALA/pull/1536#issuecomment-3014794567), it's undesirable to download anything more than what we actually need.

With this change applied, we no longer download `ecj` or any of the ad-hoc downloads (including `java-cup` and `ocamljava`) during `./gradle tasks` or during IntelliJ IDEA project import.  Thus, those actions now download 101 MB less than they did before.  Of course, these downloads will still happen _eventually_ if we actually execute tasks that require them.

### DIFF
--- a/build-logic/src/main/kotlin/com/ibm/wala/gradle/JavaCompileUsingEcj.kt
+++ b/build-logic/src/main/kotlin/com/ibm/wala/gradle/JavaCompileUsingEcj.kt
@@ -1,8 +1,8 @@
 package com.ibm.wala.gradle
 
-import java.io.File
 import javax.inject.Inject
 import org.gradle.api.Project
+import org.gradle.api.artifacts.Configuration
 import org.gradle.api.file.RegularFile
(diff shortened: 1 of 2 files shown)


msridhar (approved):
Awesome! I've seen these jars getting downloaded unexpectedly in other contexts so this is a great improvement.

liblit:
> Awesome! I've seen these jars getting downloaded unexpectedly in other contexts so this is a great improvement.

I'm glad that you like it! Please let me know if you notice other scenarios where large downloads might be happening unnecessarily.

PR #1544, opened 2025-06-30
Postpone some work until after task configuration

liblit:
The more work we do during task _configuration_, the more likely we are to trigger something (e.g., a large dependency download) that won't actually be needed.  We should prefer to postpone work until task _execution_ time, since at that point we know that there is truly something to do.


msridhar (approved):
👍

PR #1546, opened 2025-07-03
Enable Gradle configuration-on-demand

liblit:
[Configuration-on-demand](https://docs.gradle.org/current/userguide/multi_project_configuration_and_execution.html) lets Gradle skip configuring subprojects that have no tasks to run.  For example, if you tell Gradle to build the `:util` subproject, then there is no need to configure the `:dalvik` subproject or even to compile the `:dalvik` subproject's build script.


msridhar (approved):
👍

PR #1554, opened 2025-07-11
Decouple OS detection from Gradle root project

liblit:
We are refining Gradle subprojects to be less intertwined.  In this specific case, a subproject no longer looks to the root project to identify the current operating system.  Instead, any subproject that requires this information should import a small project plugin that provides this information.

### DIFF
--- a/build-logic/src/main/kotlin/com/ibm/wala/gradle/operating-system.gradle.kts
+++ b/build-logic/src/main/kotlin/com/ibm/wala/gradle/operating-system.gradle.kts
@@ -1,0 +1,23 @@
+package com.ibm.wala.gradle
+
+/**
+ * Utility script for operating system detection in Gradle builds.
+ *
+ * This script provides variables to identify the current operating system, which can be used to
+ * configure platform-specific build settings.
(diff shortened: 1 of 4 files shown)


build-logic/src/main/kotlin/com/ibm/wala/gradle/operating-system.gradle.kts
```diff
@@ -0,0 +1,23 @@
...
+ * [the `os.name` system property](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/System.html#getProperties()).
+ */
+val osName: String by extra(System.getProperty("os.name"))
+
+/**
+ * Indicates whether the current operating system is Windows.
+ *
+ * This is determined by checking if the operating system name starts with `"Windows "`.
+ */
+@Suppress("unused") val isWindows by extra(osName.startsWith("Windows "))
```
msridhar, line 23: (not critical) curious why we need this suppression; isn't this property "public"?

msridhar (approved):
LGTM

PR #1558, opened 2025-07-12
Report coverage using just-tested Java version

### LINKED ISSUE #1555: Collecting coverage reports in CI is re-running some tests
See e.g. [this run](https://github.com/wala/WALA/actions/runs/16210343455/job/45769083835?pr=1507).  Tests should _not_ need to be re-executed when running the `testCodeCoverageReport` task as they must have just passed as part of the regular build.

@liblit any idea what could be going on here?  Could the "Upload test results" job somehow be invalidating the local build results / cache?

liblit:
Previously we were running tests under Java 17, 21, and 24.  We were then reporting coverage from the Java 17 tests.  However, when aggregating that coverage report, we were actually asking for the coverage of tests run using Java 11 (the default value of `com.ibm.wala.jdk-version`).  That led to us needing to rerun nearly all tests for a second time.

Now we explicitly set the Java version when aggregating test coverage reports in the same way that we set it when running the tests the first time.  So when we aggregate coverage under Java 17, we're still requesting coverage data under Java 17, not Java 11.  That test data should already be available, and up-to-date, so we won't need to rerun the tests for a second time.

Fixes #1555.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -64,7 +64,7 @@ jobs:
           path: '**/build/test-results/test/TEST-*.xml'
       - name: Aggregate coverage
         id: jacoco_report
-        run: ./gradlew testCodeCoverageReport
+        run: ./gradlew testCodeCoverageReport "-Pcom.ibm.wala.jdk-version=${{ matrix.java }}"
         if: matrix.coverage && github.repository == 'wala/WALA'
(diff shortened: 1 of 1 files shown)


liblit:
> Thanks so much for tracking this down!

You're welcome! And thank _you_ for noticing the problem. This is a nice CI latency boost for a one-line change. Go us! 🧠

PR #1573, opened 2025-07-19
Remove some unused fields in a test class

liblit:
`IRTests.projectName` was only ever set to `null` and was never subsequently used.  `IRTests.javaHomePath` was neither set nor used. Both can be removed.  Removing the pointless initialization of `IRTests.projectName`, in turn, makes many constructors trivial and therefore removable.


msridhar:
@juliandolby are the fields removed in this PR needed for some non-open-source subclass of `IRTests` that you know of?  Or is this cleanup ok?

msridhar:
@juliandolby another ping on this one

msridhar (approved):

PR #1575, opened 2025-07-19
Use implicit `super()` to simplify constructors

liblit:
If the first thing that a constructor does is call `super()` with no arguments, then we can let the Java compiler add that call for us.

If the _only_ thing that the _only_ constructor does is call `super()` with no arguments, and if we don't mind that constructor being `public`, then we can let the Java compiler generate the entire constructor for us.


msridhar (approved):
Nice!

PR #1586, opened 2025-07-24
Be cautious with single-use `Iterable`s

liblit:
The `Iterable` returned by `Iterator2Iterable.make` is single-use. If you try to iterate through it multiple times, you get an empty sequence of elements the second and subsequent times.

That's unhelpful when reporting assertion failures.  We may need to iterate through the elements once to check them, and a second time to describe the elements in a diagnostic message if the assertion fails.

A single-use `Iterable` is also dangerous when used to produce an `Iterable` value returned by a method.  Callers may be expecting an `Iterable` that can be iterated multiple times, as that is the standard `Iterable` contract.

There are two good ways around this problem.  The generic solution is to build an `Iterable` from an uncalled reference to a nullary function that returns the `Iterator`.  In the specific case of `AssertJ` assertions, we also have the option to use the `toIterable` method. Internally, the latter consumes the `Iterator` just once, stashing its elements in a `List` that can be replayed later as often as needed.

### DIFF
--- a/core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/PointerAnalysisImpl.java
+++ b/core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/PointerAnalysisImpl.java
@@ -571,7 +571,7 @@ public class PointerAnalysisImpl extends AbstractPointerAnalysis {
    */
   @Override
   public Iterable<PointerKey> getPointerKeys() {
-    return Iterator2Iterable.make(pointsToMap.iterateKeys());
+    return pointsToMap::iterateKeys;
   }
 
   @Override
--- a/core/src/test/java/com/ibm/wala/core/tests/basic/PrimitivesTest.java
+++ b/core/src/test/java/com/ibm/wala/core/tests/basic/PrimitivesTest.java
@@ -770,7 +770,8 @@ public class PrimitivesTest extends WalaTestCase {
     Dominators<Object> D = Dominators.make(G, nodes[10]);
 
     // Assert.assertions
-    assertThat(Iterator2Iterable.make(D.dominators(nodes[4])))
+    assertThat(D.dominators(nodes[4]))
+        .toIterable()
         .containsExactly(nodes[4], nodes[7], nodes[8], nodes[5], nodes[10]);
 
     int j = 0;


core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/PointerAnalysisImpl.java
```diff
@@ -571,7 +571,7 @@ public IClassHierarchy getClassHierarchy() {
    */
   @Override
   public Iterable<PointerKey> getPointerKeys() {
-    return Iterator2Iterable.make(pointsToMap.iterateKeys());
+    return pointsToMap::iterateKeys;
```
msridhar, line 574: I wonder if this issue crops up elsewhere...

msridhar (approved):

core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/PointerAnalysisImpl.java
liblit: WALA contains 259 calls to `Iterator2Iterable.make` other than the two affected by this PR. All 259 of these use the result of the call in the `Iterable` role of a `for`-each statement, as in:

```java
for (... : Iterator2Iterable.make(...)) {
  ...
}
```

These are all safe. So no, this issue doesn't _currently_ crop up elsewhere.

---

That said, I consider `Iterator2Iterable.make` to be a dangerous API. I'd love to replace it with something that returned a proper, replayable `Iterable`. Unfortunately, all of the ways that I've considered doing that impose either a performance penalty or a more verbose syntax. So I think we're stuck with those current 259 uses for now.

It would also be nice if we could annotate `Iterator2Iterable.make` to tell linters that the returned value is single-use. Unfortunately I know of no such annotation and checker. Some linters will check for unsafe multiple use of `Iterator`s and `Stream`s, but I don't know of any that let you extend this to selected values of other types.

liblit: See #1588 for a more detailed record of what I tried as `Iterator2Iterable.make` replacements, and why I didn't consider any of the alternatives viable.

PR #1604, opened 2025-08-01
Improve readability of an exclusion list

liblit:
These exclusions are eventually represented as a stream of bytes, but semantically they are an unordered sequence of slashed package names. Let's factor out the uniform transformations that are applied to each such name.

### DIFF
--- a/core/src/test/java/com/ibm/wala/examples/analysis/dataflow/DataflowTest.java
+++ b/core/src/test/java/com/ibm/wala/examples/analysis/dataflow/DataflowTest.java
@@ -52,7 +52,9 @@ import com.ibm.wala.util.intset.IntIterator;
 import com.ibm.wala.util.intset.IntSet;
 import java.io.ByteArrayInputStream;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
@@ -66,20 +68,20 @@ public class DataflowTest extends WalaTestCase {
 
   // more aggressive exclusions to avoid library blowup
   // in interprocedural tests
-  private static final String EXCLUSIONS =
-      "java\\/awt\\/.*\n"
-          + "javax\\/swing\\/.*\n"
-          + "sun\\/awt\\/.*\n"
-          + "sun\\/swing\\/.*\n"
-          + "com\\/sun\\/.*\n"
-          + "sun\\/.*\n"
-          + "org\\/netbeans\\/.*\n"
-          + "org\\/openide\\/.*\n"
-          + "com\\/ibm\\/crypto\\/.*\n"
-          + "com\\/ibm\\/security\\/.*\n"
-          + "org\\/apache\\/xerces\\/.*\n"
-          + "java\\/security\\/.*\n"
-          + "";
+  private static final String[] EXCLUSIONS = {
+    "com/ibm/crypto",
+    "com/ibm/security",
+    "com/sun",
+    "java/awt",
+    "java/security",
+    "javax/swing",
+    "org/apache/xerces",
+    "org/netbeans",
+    "org/openide",
+    "sun",
+    "sun/awt",
+    "sun/swing",
+  };
 
   @BeforeAll
   public static void beforeClass() throws Exception {
@@ -88,7 +90,13 @@ public class DataflowTest extends WalaTestCase {
         AnalysisScopeReader.instance.readJavaScope(
             TestConstants.WALA_TESTDATA, null, DataflowTest.class.getClassLoader());
 
-    scope.setExclusions(new FileOfClasses(new ByteArrayInputStream(EXCLUSIONS.getBytes("UTF-8"))));
+    scope.setExclusions(
+        new FileOfClasses(
+            new ByteArrayInputStream(
+                Arrays.stream(EXCLUSIONS)
+                    .map(exclusion -> (exclusion + "/.*\n").replace("/", "\\/"))
+                    .collect(Collectors.joining())
+                    .getBytes("UTF-8"))));
     try {
       cha = ClassHierarchyFactory.make(scope);
     } catch (ClassHierarchyException e) {


core/src/test/java/com/ibm/wala/examples/analysis/dataflow/DataflowTest.java
```diff
@@ -88,7 +90,13 @@ public static void beforeClass() throws Exception {
         AnalysisScopeReader.instance.readJavaScope(
             TestConstants.WALA_TESTDATA, null, DataflowTest.class.getClassLoader());
 
-    scope.setExclusions(new FileOfClasses(new ByteArrayInputStream(EXCLUSIONS.getBytes("UTF-8"))));
+    scope.setExclusions(
+        new FileOfClasses(
+            new ByteArrayInputStream(
+                Arrays.stream(EXCLUSIONS)
+                    .map(exclusion -> (exclusion + "/.*\n").replace("/", "\\/"))
```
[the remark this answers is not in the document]
liblit: On further inspection, it turns out that backslash-escaping each slash is useless, because "`/`" has no special meaning in Java regular expressions. Probably the slashes used to be periods, for which backslash-escaping is relevant.

I think I may be able to improve this whole API. There's no reason we should have to simulate a file just to collect and manage a disjunction of regexp patterns. Let's see if I can clean this all up a bit.

util/build.gradle.kts
```diff
@@ -7,6 +7,7 @@ plugins {
 eclipse.project.natures("org.eclipse.pde.PluginNature")
 
 dependencies {
+  api(libs.jetbrains.annotations)
```
msridhar, line 10: How important is this dependence?  And does it do backward-incompatible changes across versions?  In my experience adding dependencies can cause surprising issues for downstream users who may be dependent on an older version so just want to check this carries its weight.

util/src/main/java/com/ibm/wala/util/config/PatternsFilter.java
```diff
@@ -0,0 +1,197 @@
...
+        Pattern.compile(
+                regexps.stream()
+                    .map(regexp -> String.format("(?:%s)", regexp))
+                    .collect(Collectors.joining("|")))
+            .asMatchPredicate();
+    return isMatch;
+  }
+
+  /**
+   * Creates a {@code SetOfClassPatterns} with the given regular expressions, one per line in the
```
msridhar, line 119: `SetOfClassPatterns` is old?

msridhar (commented):
Looks like a really nice improvement!  A couple of questions:

1. I don't fully follow how an exclusions file like https://github.com/wala/WALA/blob/master/core/src/main/resources/Java60RegressionExclusions.txt now gets processed.  That file has lots of `\/` character sequences and I'm not sure what they are for; is that just valid in the regex?  Or does it get processed away at some point?
2. Could you summarize the impact for downstream WALA users in terms of APIs?  I feel like if you're just using an exclusions file there should be no impact, is that right?

util/build.gradle.kts
liblit: This dependency provides the `@Language` annotation that we use [here](https://github.com/wala/WALA/pull/1604/files#diff-99cd86e6b771cff5923602e03658b686f8a3dae193377a7b1cdec135331a5352R144). That annotation tells IntelliJ IDEA that the annotated `String` parameter represents a regular expression. Given that hint, IntelliJ IDEA activates some useful additional features. For example, it warns about the unnecessary backslashes in `java\\/lang`. It also provides an embedded regular-expression editor, including the ability to test the expression against a probe live in the IDE. See [documentation from JetBrains](https://www.jetbrains.com/help/idea/using-language-injections.html) for more information about language injections.

There are not very many JetBrains annotations, and I've never run into backward-compatibility problems with them. That being said, I appreciate what you're saying about downstream users. The very act of adding a dependency is itself backward-incompatible. So should we discard this annotation and the corresponding dependency? Or is it worth adding? I welcome your opinion.

liblit:
> 1. I don't fully follow how an exclusions file like https://github.com/wala/WALA/blob/master/core/src/main/resources/Java60RegressionExclusions.txt now gets processed.  That file has lots of `\/` character sequences and I'm not sure what they are for; is that just valid in the regex?  Or does it get processed away at some point?

The only special processing is to split at newlines and to skip lines that start with `#`, both of which were already being done in the old code. Neither the old code nor the new code do any special processing for `\/`. Thus, if `\/` appears in some exclusion file, then `\/` is what the corresponding `Pattern` instance will see.

You may be aware, though, that `/` has no special meaning in `Pattern`'s regular expression syntax. So `\/` in a `Pattern` is the same as `/`: it matches a literal slash. Thus, all of these `\/` you're seeing are being backslash-escaped for no good reason whatsoever. That was true in the old code and it remains true now. If I had to guess, I'd say that there was probably an older version of these exclusion files that used backslash-escaped dotted names: `org\.apache\.xerces` instead of `org/apache/xerces`. Someone probably bulk-replaced the dots with slashes and neglected to remove the no-longer-necessary backslashes.

After the current PR is merged, I'll propose a follow-on PR that replaces `\/` with `/` wherever these are equivalent. I've already verified that all tests continue to pass after that change.

>     2. Could you summarize the impact for downstream WALA users in terms of APIs?  I feel like if you're just using an exclusions file there should be no impact, is that right?

Correct: the exclusions file syntax has not changed. Additionally, the deprecated `FileOfClasses` and `SetOfClasses` classes have been retrofitted to implement the `StringFilter` interface. Thus, code like the following will continue to work:

```java
setExclusions(new FileOfClasses(…));  // formal parameter type is now `StringFilter`
```

Thus, we have backward compatibility for Java _sources_ that use the WALA API. However, we don't have backward compatibility for Java _bytecode_ that uses the WALA API. Bytecode that has already been compiled will expect to find an `AnalysisScope.setExclusions(SetOfClasses)` method, which I've replaced with a `AnalysisScope.setExclusions(StringFilter)` method. Is that OK? Or should we restore a `AnalysisScope.setExclusions(SetOfClasses)` method, mark it as deprecated, and have it simply call the new `AnalysisScope.setExclusions(StringFilter)` method? That would give us backward compatibility at the _bytecode_ level, not just at the source level.

util/build.gradle.kts
liblit: Actually, we can get the best of both worlds here. The `@Language` annotation is retained in the `.class` file, but is not actively _used_ at run time. Therefore, the JetBrains annotations dependency can be a `compileOnly` dependency instead of an `api` dependency. That should keep things perfectly backward-compatible for downstream users: they won't know about this dependency at all, since it will only have been used when compiling WALA itself.

PR #1609, opened 2025-08-05
Do not backslash-escape slashes in `Pattern`s

liblit:
`/` has no special meaning in `Pattern` regular expressions, so there's no reason to backslash-escape them.

This change is fully backward-compatible.  `\/` in an exclusion file was equivalent to `/` before, and that's still true now.


msridhar (approved):
Great cleanup!

PR #1613, opened 2025-08-06
Remove two useless calls to getter methods

liblit:
The call to `NewInstruction.getType()` does not do anything useful. `NewInstruction` is a `final` class, so we know exactly which implementation of `getType()` will be called.  That implementation has no side effects.  So if we don't need the returned value, then there's no reason to call it.

The call to `super.isAssigned()` _appears_ to be useless as well.  We know exactly which implementation of `isAssigned()` will be called. That implementation has no side effects.  So if we don't need the returned value, then there's no reason to call it.  **However,** I do wonder whether this was supposed to be a call to `super.setAssigned()` instead.  Manu, any thoughts on that?

### DIFF
--- a/core/src/main/java/com/ibm/wala/core/util/ssa/ParameterAccessor.java
+++ b/core/src/main/java/com/ibm/wala/core/util/ssa/ParameterAccessor.java
@@ -242,7 +242,6 @@ public class ParameterAccessor {
 
       this.disp = disp;
       this.descriptorOffset = descriptorOffset;
-      super.isAssigned();
     }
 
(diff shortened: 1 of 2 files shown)


msridhar:
Need to think about this one a bit, will do so when I can

liblit:
> Need to think about this one a bit, will do so when I can

Thanks. No rush!

Meta-comment: if we cannot easily tell whether `super.isAssigned()` should be `super.setAssigned()`, then we're probably missing an automated test that would be useful to add.

liblit:
> Thanks for catching these!

You're welcome. And thank _you_ for reviewing these changes with the care that they deserve.

PR #1615, opened 2025-08-07
Remove many redundant local variables

liblit:
"Redundant" can be subjective, but JetBrains offers an inspection that uses a reasonable definition:

> Reports unnecessary local variables that add nothing to the comprehensibility of a method, including:
>
> * Local variables that are immediately returned.
>
> * Local variables that are immediately assigned to another variable and then not used.
>
> * Local variables that always have the same value as another local variable or parameter.

### DIFF
--- a/cast/js/src/main/java/com/ibm/wala/cast/js/util/JSCallGraphBuilderUtil.java
+++ b/cast/js/src/main/java/com/ibm/wala/cast/js/util/JSCallGraphBuilderUtil.java
@@ -143,14 +143,12 @@ public class JSCallGraphBuilderUtil extends com.ibm.wala.cast.js.ipa.callgraph.J
   public static Module[] makeSourceModules(String dir, String name, ClassLoader loader)
       throws IOException {
     URL script = getURLforFile(dir, name, loader);
-    Module[] modules =
-        new Module[] {
-          (script.openConnection() instanceof JarURLConnection)
-              ? new SourceURLModule(script)
-              : makeSourceModule(script, dir, name),
-          getPrologueFile("prologue.js")
-        };
-    return modules;
+    return new Module[] {
+      (script.openConnection() instanceof JarURLConnection)
+          ? new SourceURLModule(script)
+          : makeSourceModule(script, dir, name),
+      getPrologueFile("prologue.js")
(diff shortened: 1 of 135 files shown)


cast/js/src/main/java/com/ibm/wala/cast/js/util/JSCallGraphBuilderUtil.java
```diff
@@ -194,9 +192,8 @@ public static CallGraph makeScriptCG(
       String dir, String name, CGBuilderType builderType, ClassLoader loader)
       throws IOException, IllegalArgumentException, CancelException, WalaException {
     PropagationCallGraphBuilder b = makeScriptCGBuilder(dir, name, builderType, loader);
-    CallGraph CG = b.makeCallGraph(b.getOptions());
     // dumpCG(b.getPointerAnalysis(), CG);
```
msridhar, line 195: Probably in this case the local was retained so that the next line could easily be uncommented.  Can we retain this one somehow?

cast/js/src/main/java/com/ibm/wala/cast/js/util/JSCallGraphBuilderUtil.java
```diff
@@ -208,9 +205,8 @@ public static CallGraph makeScriptCG(
             : null;
     PropagationCallGraphBuilder b =
         makeCGBuilder(makeLoaders(preprocessor), scripts, builderType, irFactory);
-    CallGraph CG = b.makeCallGraph(b.getOptions());
```
msridhar, line 211: same comment as above

msridhar (commented):
great cleanup!

cast/js/src/main/java/com/ibm/wala/cast/js/util/JSCallGraphBuilderUtil.java
liblit: Sure, [easily done](https://github.com/wala/WALA/pull/1615/files#diff-cd424bb52b98730f78610bc25c6016af3e50283c867e1dc7678b8dac9250f1a2R195-R198). Good call.

PR #1618, opened 2025-08-07
Use more structured loop control flow

liblit:
IMHO, the control flow of these loops is easier to follow if we reduce the use of `if`-guarded `break`, `continue`, and `return` statements.

### DIFF
--- a/shrike/src/main/java/com/ibm/wala/shrike/shrikeBT/analysis/Analyzer.java
+++ b/shrike/src/main/java/com/ibm/wala/shrike/shrikeBT/analysis/Analyzer.java
@@ -226,9 +226,7 @@ public class Analyzer {
       return topType;
     }
 
-    String x = ClassHierarchy.findCommonSupertype(hierarchy, patchType(t1), patchType(t2));
-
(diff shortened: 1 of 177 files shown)


shrike/src/main/java/com/ibm/wala/shrike/shrikeBT/analysis/Analyzer.java
```diff
@@ -268,12 +268,9 @@ private void getReachableRecursive(
...
     }
-    while (true) {
-      // stop if we've already visited this instruction or if we've gone outside
-      // the mask
-      if (reachable.get(from) || (mask != null && !mask.get(from))) {
-        return;
-      }
+
+    // stop if we've already visited this instruction or if we've gone outside the mask
+    for (; !reachable.get(from) && (mask == null || mask.get(from)); ++from) {
```
msridhar, line 273: Would you be ok with a `while` loop here?  I find I have to think slightly harder about `for` loops without an initializer 🙂

msridhar (approved):

PR #1619, opened 2025-08-07
Remove redundant conditions

liblit:
Each removed Boolean condition is redundant, as it is covered by some other subsequent condition.  For example, the `null` check is redundant in "`x != null && x instanceof C`", because "`null instanceof C`" always evaluates to `false`.


msridhar (approved):
Nice!

PR #1621, opened 2025-08-07
Simplify some arithmetic expressions

liblit:
In `PruneArrayOutOfBoundExceptionEdge`, two `static final int` fields are initialized to "`0 + …`".  We want to keep them in this form.  A comment for each of those fields makes it clear that the value is "#[array access] + #[other]", and we would lose that structure if we simplified.  So we suppress IntelliJ IDEA's warnings for these fields.

In the other cases, though, the simplification seems natural and useful.


msridhar (approved):
👍

PR #1623, opened 2025-08-07
Remove some dead assignments; allow others

liblit:
For many of these cases, the dead assignment is the result of incrementing some `int` cursor when there will be no subsequent uses of that new cursor value.  I'm allowing most of those to remain, since there is some readability value in knowing that the cursor steps forward in a systematic, uniform manner, even up to the very last use.  For example, if one were to add further uses of the cursor in the future, then it will be good to have had that last increment in place.

For the remaining cases, though, the assignment being removed seems to provide little or no value for future code readers or maintainers.


msridhar (approved):
Nice!

PR #1625, opened 2025-08-11
Try to use a Mirror for Kawa download

msridhar:
CI jobs are failing due to `ftp.gnu.org` being down (from where we grab our Kawa download).  Using the mirror gateway for the download should be more robust.


msridhar:
@liblit any ideas on what is going on here?  I wonder if the re-direct is incompatible with how we do ad-hoc downloads?

PR #1668, opened 2025-09-06
Refactor Gradle CI/CD command across platforms

liblit:
Previously we ran Gradle using either of two distinct `run` steps: one for Linux and one for non-Linux.  However, these two steps had much in common.  The only two differences were that Linux ran Gradle under Xvfb and that Linux used a few extra Gradle build tasks.

Now we've factored out these differences more cleanly:

For using Xvfb only on Linux, we now rely on a helpful GitHub action. This `GabrielBB/xvfb-action` action runs a command under Xvfb on Linux, or without Xvfb on non-Linux.  Perfect!

For adding some extra Gradle build tasks, we conditionally set an `OS_SPECIFIC_GRADLE_TASKS` environment variable to those extra tasks, if any.  Later, we insert the value of that variable at the appropriate place in the Gradle command line.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -46,14 +46,13 @@ jobs:
           distribution: 'zulu'
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v4
-      - name: Build and test using Gradle with ECJ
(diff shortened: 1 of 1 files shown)


msridhar (approved):
Nice cleanup!

PR #1673, opened 2025-09-14
Update for JDK 25

msridhar:


msridhar:
Looks like I spoke too soon.  Everything passes on my Mac but something is up on Linux...

liblit:
FYI, I have an upgrade for Gradle 9.1.0 in the works. Should be ready for you to review later today.

msridhar:
@liblit had to do some fixes here, so needs another review

msridhar:
> Note that the repository still thinks that "JDK 24 on ubuntu-latest" is required before merging to the `master` branch.

Fixed!  We now require the JDK 25 job to pass.

PR #1675, opened 2025-09-16
Revamp management of OCaml-Java artifacts

liblit:
Some WALA tests use bytecode created by the OCaml-Java compiler.  That compiler is a sizeable download.  For faster builds, we previously kept a copy of the generated `hello_hash.jar` file in Git, and only tried to rebuild it if it was missing.

However, code that never runs is code that eventually stops working. The `unpackOcamlJava` task that prepared the OCaml-Java distribution for use was incompatible with Gradle's configuration cache.  Even worse, the `generateHelloHashJar` task that actually ran the OCaml-Java compiler was not running it correctly, and hasn't actually been able to recreate `hello_hash.jar` for a long time.

This commit fixes `unpackOcamlJava` so that it is compatible with Gradle's configuration cache.  It also overhauls `generateHelloHashJar` to run the OCaml-Java compiler correctly. Lastly, it removes `hello_hash.jar` from Git so that we will always download OCaml-Java and use it when needed to create this test input. Yes, it's a sizeable download.  But our downloads make much better use of Gradle's dependency cache than they used to.  This should generally be a one-time download that gets cached and reused transparently for future builds.

Unfortunately I have been unable to get OCaml-Java to work on Windows. Something's broken with how its linker finds the bytecode implementation of the OCaml standard library.  So for now, I've disabled the OCaml-Java test on Windows.  We're still testing all of this on Linux and macOS.  I think that losing the Windows test is better than being unable to recreate `hello_hash.jar` entirely.


msridhar (approved):
Thanks for digging into this and fixing things up to work correctly!  Agreed that this change is worthwhile though we need to skip testing on Windows

PR #1678, opened 2025-09-16
Postpone turning archives into `FileTree`s

liblit:
By fiddling with parentheses and braces a bit, we slightly postpone calls to `tarTree` and `zipTree`, potentially avoiding them entirely if the enclosing task is configured but not executed.  In general, Gradle performance is helped by delaying as much computation as possible.

### DIFF
--- a/cast/js/build.gradle.kts
+++ b/cast/js/build.gradle.kts
@@ -48,7 +48,7 @@ val downloadAjaxslt =
 
 val unpackAjaxslt by
     tasks.registering(Sync::class) {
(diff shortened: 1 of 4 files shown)


msridhar (approved):
Would this change potentially lead to these files not getting downloaded at all in certain cases where they are not needed?

liblit:
> Would this change potentially lead to these files not getting downloaded at all in certain cases where they are not needed?

Yes, I think that could be a consequence of this change.

PR #1689, opened 2025-09-17
Improve task configuration avoidance

liblit:
Some Gradle task-related APIs eagerly force tasks to be created.  Others are lazy, merely installing closures that may _eventually_ be called _if_ a suitable task is created.  We prefer the latter.  Unfortunately, it's hard to remember which APIs fall into which category.  For example, some `withType` methods of some classes are eager, while others with the same name (even within the same class) are lazy.  Oof!  I thought that we were already being quite lazy, but I missed a few cases, including some that are heavily used.

I assessed the impact of this change using `./gradlew build -PexcludeSlowTests --no-daemon --rerun-tasks --scan` in a clean build tree.  After these changes, 127 fewer tasks are created during Gradle's configuration phase.  85 of those are created during task graph calculation, i.e., 85 of these were still needed eventually in order to run the build.  The remaining 42, however, were not created at all.  So those were tasks that were previously being created too eagerly even though they were never actually used for this build.

### DIFF
--- a/build-logic/src/main/kotlin/com/ibm/wala/gradle/NullAway.gradle.kts
+++ b/build-logic/src/main/kotlin/com/ibm/wala/gradle/NullAway.gradle.kts
@@ -13,7 +13,7 @@ dependencies {
   compileOnly(versionCatalogs.named("libs").findLibrary("nullaway-annotations").get())
 }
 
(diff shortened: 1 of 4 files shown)


msridhar (approved):
Great finds!

PR #1692, opened 2025-09-18
Close a file output stream when done with it

liblit:
The garbage collector and operating system would clean this up for us eventually, but it's good practice to close file streams promptly as soon as we are done writing to them.


liblit:
Oops, this change won't work on Windows. It tries to close the `FileOutputStream` too early: before the underlying subprocess has even run! macOS and Linux are quite forgiving of this, in the same way that they let you delete an open file. Windows doesn't allow that, so the task hangs.

I'll need to rethink my strategy here.

liblit:
I've revised the approach to be best-effort: we now close the stream if the task that used this helper completes successfully. Unfortunately the stream is not closed promptly after a task failure. [I've asked the Gradle community for any ideas on how we can do better.](https://discuss.gradle.org/t/close-fileoutputstream-after-javaexec-task/51592/1) If someone comes up with something I missed, then we can incorporate that improvement later.

msridhar (approved):

PR #1700, opened 2025-09-21
Make foojay versions consistent

msridhar:
This version is now consistent with what is used in the root `settings.gradle.kts` file.  This older version caused an issue when I was testing WALA in a Docker image running Gradle on JDK 25.

@liblit is there some way to keep this version number in only one place?


liblit:
> @liblit is there some way to keep this version number in only one place?

Unfortunately, we cannot keep it in `libs.versions.toml`, as that is not yet loaded when `settings.gradle.kts` files are being processed.

We could put it in some external file that both `settings.gradle.kts` files read:

```kotlin
plugins {
  ...
  id("org.gradle.toolchains.foojay-resolver-convention") version
          File("foojay-version.txt").readText().trim()
  ...
}
```

liblit:
> We could put it in some external file that both `settings.gradle.kts` files read:

A tidy variant of this could be to put the version in `gradle.properties`. I haven't actually tried that, though. @msridhar, if you want to try this but need help with the details, let me know and I can take a swing at it.

msridhar:
> > We could put it in some external file that both `settings.gradle.kts` files read:
> 
> A tidy variant of this could be to put the version in `gradle.properties`. I haven't actually tried that, though. @msridhar, if you want to try this but need help with the details, let me know and I can take a swing at it.

If you have time, since you are far more of a Gradle expert than me and you could take a swing at it, that would be great!  Nothing at all urgent  here, I just opened a PR since I ran into the problem and didn't want to forget.  I believe you should be able to push to this PR if you like, or you can just open a new one if that's easier.

liblit:
> I believe you should be able to push to this PR branch if you like

Done.

PR #1716, opened 2025-10-01
Upgrade to Spotless Gradle plugin 8.0.0

liblit:
This plugin version requires Java 17 or newer.  Specifically, the JVM that is running Gradle itself must be Java 17+.  However, that's not a new requirement for us:  we committed to requiring Java 17+ to run Gradle back in revision 292991c802bb2317beef8cc613e6b9c77cf9ee0c, in late June of this year.

That being said, when compiling the Kotlin code in `build-logic`, we were still targeting Java 11.  Some of that code configures the Spotless Gradle plugin, so we need to move `build-logic` to Java 17+ in order for it to talk to the Spotless Gradle plugin's Java 17+ code. I'm not too concerned about that change:  as I said, we've required that Gradle use Java 17+ for a few months now.

### DIFF
--- a/build-logic/build.gradle.kts
+++ b/build-logic/build.gradle.kts
@@ -16,7 +16,7 @@ dependencies {
   implementation(libs.gradle.spotless.plugin)
 }
 
-kotlin.jvmToolchain(11)
+kotlin.jvmToolchain(17)
 
 spotless {
   val ktfmtVersion = libs.versions.ktfmt.get()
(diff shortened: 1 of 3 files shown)


msridhar (approved):
👍

I guess we no longer run WALA tests on JDK 11, even though we still support it.  That's a bit disturbing, but something that should be handled in a separate issue / PR.

PR #1719, opened 2025-10-07
Run tests on JDK 11 on CI

### LINKED ISSUE #1718: Investigate whether we test on JDK 11 in CI
@liblit wrote in detail [here](https://github.com/wala/WALA/pull/1716#issuecomment-3363495263) how by default, WALA unit tests will run on JDK 11.  However, in our CI config, we always pass the `com.ibm.wala.jdk-version` Gradle property, and it is never set to 11 as far as I can tell:

https://github.com/wala/WALA/blob/4dad931c8e08fe8a49cf159a21d57a80b25ef2dd/.github/workflows/continuous-integration.yml#L60

So, it seems to me we don't end up testing on JDK 11 on CI.

@liblit did I miss something?

msridhar:
We add an appropriate entry in our CI matrix, and tweak our use of `setup-java` to just install all the JDK versions we might need.

Fixes #1718

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -42,7 +42,7 @@ jobs:
     steps:
       - name: Check out WALA sources
         uses: actions/checkout@v4
-      - name: 'Set up JDK ${{ matrix.java }}'
+      - name: 'Set up JDKs'
         uses: actions/setup-java@v5
         with:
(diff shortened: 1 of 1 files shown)


liblit (approved):
After this change is merged to `master`, please remember to add the new job configuration to the required checks for that branch.

PR #1726, opened 2025-10-14
Correct a test that was succeeding vacuously

liblit:
Previously `TestSimpleCallGraphShape.testFunctionDotCall` succeeded vacuously, because the call graph contains no node `n` such that `n.getMethod().getName().toString().equals("call4")` is true.  However, there is a node whose method name is `"$$ call_4"`.  I suspect that's the node that these assertions are intended to examine.

Correcting that name check reveals that the assertion on the string representation of this node's successor list also fails, apparently due to those two successors being in an unexpected order.  I've corrected that assertion as well.  However, are successor nodes guaranteed to be in any consistent order?  If not, then I would prefer to replace this assertion with one that is order-independent.  I'm happy to do that in a future commit if reviewers think that would be appropriate.


msridhar (approved):
Thanks for finding this!  And yes, ideally we would do an assertion here that does checks the succs in an order-independent way

liblit:
> ideally we would do an assertion here that does checks the succs in an order-independent way

Will do, in an upcoming pull request! Thanks for this guidance.

PR #1730, opened 2025-10-15
Defend test against vacuous success and order changes

liblit:
An earlier version of this test passed vacuously because it was looking for a node named `"call4"` instead of `"$$ call_4"`.  That mistake has already been corrected.  However, we can make this test even more robust by ensuring that the sought-after node appears exactly once.

Furthermore, when validating the contents of a list of successors, we should not assume that those successors are given in any particular order.

See #1726 for further discussion on recent corrections and improvements to this test.

### DIFF
--- a/cast/js/src/testFixtures/java/com/ibm/wala/cast/js/test/TestSimpleCallGraphShape.java
+++ b/cast/js/src/testFixtures/java/com/ibm/wala/cast/js/test/TestSimpleCallGraphShape.java
@@ -28,9 +28,8 @@ import com.ibm.wala.util.CancelException;
 import com.ibm.wala.util.MonitorUtil.IProgressMonitor;
 import com.ibm.wala.util.NullProgressMonitor;
 import com.ibm.wala.util.WalaException;
-import com.ibm.wala.util.collections.Iterator2Collection;
 import java.io.IOException;
-import java.util.List;
+import org.assertj.core.api.InstanceOfAssertFactories;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


cast/js/src/testFixtures/java/com/ibm/wala/cast/js/test/TestSimpleCallGraphShape.java
```diff
@@ -493,16 +492,15 @@ public void testExtJSSwitch()
...
-        List<CGNode> succs = Iterator2Collection.toList(cg.getSuccNodes(n));
-        assertThat(succs)
-            .hasToString(
-                "[Node: <Code body of function Lfunction_call.js/bar> Context: Everywhere, Node: <Code body of function Lfunction_call.js/foo> Context: Everywhere]");
-      }
-    }
+    assertThat(cg)
+        .filteredOn(n -> n.getMethod().getName().toString().equals("$$ call_4"))
+        .singleElement()
+        .extracting(cg::getSuccNodes, InstanceOfAssertFactories.iterator(CGNode.class))
```
msridhar: Why do we need the `InstanceOfAssertFactories.iterator()` call here?

liblit: With that argument, we call [this overload of `extracting`](https://javadoc.io/static/org.assertj/assertj-core/3.27.6/org/assertj/core/api/AbstractObjectAssert.html#extracting(java.util.function.Function,org.assertj.core.api.InstanceOfAssertFactory)), which returns an [`IteratorAssert`](https://javadoc.io/static/org.assertj/assertj-core/3.27.6/org/assertj/core/api/IteratorAssert.html), which has the [`toIterable()`](https://javadoc.io/static/org.assertj/assertj-core/3.27.6/org/assertj/core/api/AbstractIteratorAssert.html#toIterable()) method that we call on the next line.

Without that argument, we call [this overload of `extracting`](https://javadoc.io/static/org.assertj/assertj-core/3.27.6/org/assertj/core/api/AbstractObjectAssert.html#extracting(java.util.function.Function)), which returns an [`AbstractObjectAssert`](https://javadoc.io/static/org.assertj/assertj-core/3.27.6/org/assertj/core/api/AbstractObjectAssert.html), which has no ` toIterable()` method for us to call on the next line.

[More examples here.](https://assertj.github.io/doc/#assertj-core-3.14.0-extracting-function)

msridhar (approved):

PR #1731, opened 2025-10-15
Correct and refine a vacuous test

liblit:
This test was passing vacuously, because its assertions were never actually executing.  The revised version corrects the node-matching criteria so that assertions have a chance to run.  The revised version also strengthens the test so that failing to match any nodes at all is considered a failure instead of a vacuous success.


msridhar (approved):
Nice find!

liblit:
> Nice find!

Thanks! I'm now considering options for using coverage data to identify all `assertThat` calls that never execute across all tests.

Issue #1733, opened 2025-10-16
The return value of `com.ibm.wala.ipa.callgraph.propagation.AllocationSiteInNode.toString()` doesn't include its `com.ibm.wala.classLoader.NewSiteReference`


khatchad:
https://github.com/wala/WALA/blob/02e47dcc59728b4327dcd27ef2a5c5ae27ce05cb/core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/AllocationSiteInNode.java#L36-L45

The return value of `com.ibm.wala.ipa.callgraph.propagation.AllocationSiteInNode.toString()` doesn't include its `com.ibm.wala.classLoader.NewSiteReference` instance field. This is quite problematic during debugging because `com.ibm.wala.ipa.callgraph.propagation.NormalAllocationInNode.equals(Object)` *does* consider this field. Often times, I use the `String` representation to understand whether the objects are equal, especially when sifting through large logs. Furthermore, site the site is a field of the class, why not include it in `toString()`?

The field may be found at: https://github.com/wala/WALA/blob/02e47dcc59728b4327dcd27ef2a5c5ae27ce05cb/core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/AllocationSiteInNode.java#L23

PR #1735, opened 2025-10-17
Add `NewSiteReference` to `toString()` in `AllocationSiteInNode`

khatchad:
The return value of `com.ibm.wala.ipa.callgraph.propagation.AllocationSiteInNode.toString()` now includes its `com.ibm.wala.classLoader.NewSiteReference`.  I used the `toString()` method of `AllocationSite` as a guide.

Fixes https://github.com/wala/WALA/issues/1733.


liblit (changes requested):
Thanks, @khatchad!

When this change is approved and squashed into `master`, the description of this PR will become the commit message. The current description, "Fixes #1733", will be completely opaque if, for example, we were ever to migrate off of GitHub. @khatchad, would you please expand that description to explain the purpose of the change even if the #1733 link is not available? Feel free to copy and paste anything from the original #1733 bug report. And feel free to keep the #1733 reference as _part_ of the description. It just shouldn't be the _entire_ description.

PR #1741, opened 2025-10-19
Install a suitable Node.js for Markdown linting

liblit:
Previously we assumed that the build machine already had a reasonable version of the `npx` tool from Node.js.  Now we download and install a version that we know works.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -89,12 +89,19 @@ shellcheck {
   shellcheckBinary = "shellcheck"
   sourceFiles =
       fileTree(".") {
+        exclude("/.gradle/")
         exclude("**/build")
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -112,7 +119,10 @@ val lintMarkdown by
 
 tasks.named("check") { dependsOn("buildHealth", lintMarkdown) }
 
-tasks.named("shellcheck") { group = "verification" }
+tasks.named("shellcheck") {
+  group = "verification"
+  dependsOn("nodeSetup")
```
msridhar: How is `shellcheck` related to the node.js installation?  I thought `shellcheck` was implemented in Haskell?

PR #1776, opened 2025-11-19
Leverage return value of AssertJ `fail`

liblit:
`fail` will never actually return.  However, it pretends to return any type that is needed.

### DIFF
--- a/cast/java/src/testFixtures/java/com/ibm/wala/cast/java/test/IRTests.java
+++ b/cast/java/src/testFixtures/java/com/ibm/wala/cast/java/test/IRTests.java
@@ -501,8 +501,7 @@ public abstract class IRTests {
         return loader.getReference();
       }
     }
-    fail("This code should be unreachable");
-    return null;
+    return fail("This code should be unreachable");
   }
 
   public static void populateScope(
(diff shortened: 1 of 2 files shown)


liblit:
This `fail` method is annotated with `@CanIgnoreReturnValue`. That annotation _should_ make IntelliJ IDEA silently allow ignoring the returned value. Unfortunately, that doesn't work for some reason. Each `fail` call leads to an IntelliJ IDEA "Result of method call ignored" warning.

We can suppress that warning at individual call sites by adding a comment on the line before:

```java
//noinspection ResultOfMethodCallIgnored
```

In cases where the `fail` appeared immediately before a `return <expr>`, I thought it simpler and cleaner to combine the `return` with the `fail` call instead of adding a suppression comment.

This style felt natural and familiar to me from Kotlin. In Kotlin, `throw <expr>` is an _expression_ (not just a statement). From a static type checking standpoint, the value computed by `throw` has a `Nothing` type that can be used anywhere. In reality, that value will never actually be provided and used at run time, because the thrown exception will interrupt the normal control flow. `Nothing` is also commonly used as the return type of library- or user-defined functions that never actually return. Call to such functions can be used anywhere an expression of any type is expected.

If you prefer the suppression comment over `return fail(…)`, let me know and I'll revise this PR accordingly.

msridhar:
SGTM and makes sense

PR #1777, opened 2025-11-19
Allow a bit more covariance in maps and sets

liblit:


liblit:
The motivation is to allow the `Hash{Map,Set}Factory.make` methods to be used in a few ways that are safe but not currently allowed. For example:

```java
Set<String> names = …;
Set<Object> things = HashSetFactory.make(names);
```

The above code fails to type check if the formal parameter to `HashSetFactory.make` is declared as `Collection<T>`, which is how things stand before this PR. After this PR, that formal parameter is declared as `Collection<? extends T>`, and the above passes the type checker.

msridhar (approved):
I've never had a need for this extra flexibility personally but why not 🙂

liblit:
An actual need for this change arose during some other `Hash{Map,Set}Factory` work that I'm still developing. I cannot find the exact spot right now, though. If I come across it again, I'll point it out as a non-contrived example of this change being useful.

PR #1794, opened 2025-12-03
Initialize some variables in one assignment each, not two

liblit:


liblit:
My preferences for both of these are strongly influenced by my IDE:

* IntelliJ IDEA underlines non-`final` variables. So when I'm editing code, if I _don't_ see an underline, then I know there's one and only one place I need to consider regarding where that variable got its value.

* IntelliJ IDEA adds an inlay hint giving the inferred type for each `var` declaration. So using `var` doesn't reduce readability for me. Conversely, I like `var` because I like type inference in general, and because letting the compiler infer more can make future evolution of an API easier by reducing the number of places where an explicit type must be updated.

That said, these are merely soft preferences. I can revise this PR to use explicit types and omit `final` if you wish.

msridhar (dismissed):

PR #1803, opened 2025-12-11
Record catch types as Set, not array

liblit:
Examining AstTranslator.CodeEntityContext.setCatchType, it is clear that the catch types for a given SSA instruction are conceptually a duplicate-free set, not a random-access array.  Let's use `Set` to manage those types concisely and efficiently.

This change has cascading effects across quite a few files, including some `public` APIs.  I'm not sure it's wise to change those APIs without at least a minor version bump.  Ideally we would keep the old APIs around too for some time, but mark them as deprecated.  Unfortunately, that's not straightforward:  the old and new methods' signatures are identical after erasure of generic type parameters.  In some cases, the signatures would be identical even without erasure.  So we cannot simply have both old and new overloads

We _could_ add new APIs with different names, though.  For example, `AstTranslator.WalkContext` could have:

```java
@Deprecated Map<IBasicBlock<SSAInstruction>, TypeReference[]> getCatchTypes();
Map<IBasicBlock<SSAInstruction>, Set<TypeReference>> getCatchTypesSets();
```

Is that worth doing?

### DIFF
--- a/cast/src/main/java/com/ibm/wala/cast/ir/translator/AstTranslator.java
+++ b/cast/src/main/java/com/ibm/wala/cast/ir/translator/AstTranslator.java
@@ -10,7 +10,8 @@
  */
 package com.ibm.wala.cast.ir.translator;
 
-import com.google.common.collect.Sets;
+import static java.util.Objects.requireNonNullElseGet;
+
 import com.ibm.wala.cast.ir.ssa.AssignInstruction;
 import com.ibm.wala.cast.ir.ssa.AstAssertInstruction;
 import com.ibm.wala.cast.ir.ssa.AstEchoInstruction;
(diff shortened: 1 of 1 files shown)


msridhar (commented):
I agree using a `Set` abstraction here makes sense, but I'm not sure about using `HashSet`s as the implementation.  My guess is that these sets are small, and that primarily they are iterated over rather than used for checking membership of some exception type.  If that's the case, maybe we should use [`ArraySet`](https://github.com/wala/WALA/blob/a1f25d74d79dd0a851830c7e0611c2b42f534e56/util/src/main/java/com/ibm/wala/util/collections/ArraySet.java) as the implementation instead?

liblit:
> maybe we should use [`ArraySet`](https://github.com/wala/WALA/blob/a1f25d74d79dd0a851830c7e0611c2b42f534e56/util/src/main/java/com/ibm/wala/util/collections/ArraySet.java) as the implementation instead?

Agreed. I'll revise this PR accordingly. That should be a tiny change, thanks to the abstraction magic of the `Set` interface.

msridhar (approved):
Looks great!

PR #1820, opened 2025-12-22
Avoid redundant map lookups in `cast`

liblit:
Remove some redundant `Map` lookups, such as calling `containsKey` immediately before `get`.  That's redundant as long as we never map to `null`, which appears to be the case for the maps we're modifying here.

Also codify the existing `null`-avoidance policy for these maps by adding JSpecify annotations.  It's not clear how thoroughly current tools check this annotation when used on a `Map`'s value type.  But even if current tools don't check them, it's worth adding these annotations as documentation and (hopefully) as hints for future checkers.

### DIFF
--- a/cast/src/main/java/com/ibm/wala/cast/ipa/callgraph/CrossLanguageCallGraph.java
+++ b/cast/src/main/java/com/ibm/wala/cast/ipa/callgraph/CrossLanguageCallGraph.java
@@ -68,8 +68,8 @@ public class CrossLanguageCallGraph extends AstCallGraph {
   public AbstractRootMethod getLanguageRoot(Atom language) {
     return languageRoots.computeIfAbsent(
         language,
-        absent -> {
-          AbstractRootMethod languageRoot = roots.get(absent, this);
+        newLanguage -> {
+          AbstractRootMethod languageRoot = roots.get(newLanguage, this);
 
           CGNode languageRootNode = null;
           try {


cast/src/main/java/com/ibm/wala/cast/ipa/callgraph/CrossLanguageCallGraph.java
```diff
@@ -62,37 +63,37 @@ public CrossLanguageCallGraph(
...
 
-  private final Map<Atom, IMethod> languageRoots = HashMapFactory.make();
+  private final Map<Atom, @NonNull AbstractRootMethod> languageRoots = HashMapFactory.make();
 
   public AbstractRootMethod getLanguageRoot(Atom language) {
-    if (!languageRoots.containsKey(language)) {
-      AbstractRootMethod languageRoot = roots.get(language, this);
+    return languageRoots.computeIfAbsent(
+        language,
+        absent -> {
```
msridhar: should the parameter be named `language`?  It is actually used

liblit: We cannot name this lambda's parameter `language`, because that would shadow the existing `language` parameter to the `getLanguageRoot` method itself. Java doesn't allow that kind of shadowing.

I suppose we could call it `absentLanguage` or `key` or `newLanguage` if you prefer any of those. Or we could rename the `language` parameter to `getLanguageRoot`.

liblit: Easily done! Thanks for raising an eyebrow when you saw confusing code.

msridhar (approved):

PR #1822, opened 2025-12-23
Avoid redundant map lookups in `util`

liblit:
Remove some redundant `Map` lookups, such as calling `containsKey` immediately before `get`.  That's redundant as long as we never map to `null`, which is the case for the maps we're modifying here.


msridhar (commented):
Since all the packages in `util` are now `@NullMarked`, we should only need explicit `@NonNull` annotations inside code that is explicitly `@NullUnmarked`, correct?

liblit:
> Thanks for all these improvements across the WALA code base!

You are welcome! And thank _you_ for shepherding them along with careful reviews and high standards.

PR #1877, opened 2026-04-09
Convert `assert` statements to AssertJ

liblit:
The rest of WALA uses AssertJ for testing, so we should do that here too.  AssertJ's higher-level abstractions provide much richer diagnostics when checks fail.

The version of `json-unit-assertj` that we are using here isn't the newest release, but it is the last release compatible with Java 11. Fortunately, that release already had everything we need.


msridhar (approved):
👍

PR #1878, opened 2026-04-11
Update Error Prone and NullAway to latest

msridhar:
Fix a number of calls to use `StandardCharsets.UTF_8` to fix a new Error Prone warning


cast/js/src/main/java/com/ibm/wala/cast/js/ipa/callgraph/JSSSAPropagationCallGraphBuilder.java
```diff
@@ -206,16 +206,10 @@ protected AbstractFieldPointerKey fieldKeyForUnknownWrites(AbstractFieldPointerK
...
 
   static {
-    FieldReference x = null;
-    try {
-      byte[] utf8 = "__proto__".getBytes("UTF-8");
-      x =
-          FieldReference.findOrCreate(
-              JavaScriptTypes.Root, Atom.findOrCreate(utf8, 0, utf8.length), JavaScriptTypes.Root);
-    } catch (UnsupportedEncodingException e) {
-      assert false;
```
liblit, line 216: I've recently come to see `assert false` as an anti-pattern: it changes the program's control flow depending on whether the JVM is in `assert`-enforcement mode or not. I'm glad that this change eliminates one such statement and makes the code simpler too!

liblit (approved):
Nice simplifications regarding character-set encodings!

PR #1879, opened 2026-04-13
Use `JrtModule` consistently to handle standard library on modern JDKs

### LINKED ISSUE #1876: Remove redundancy between JrtModule and JDK module support
I notice now that we have some redundant support for `jrt` modules built into the JDK.  See also:

https://github.com/wala/WALA/blob/master/core/src/main/java/com/ibm/wala/ipa/callgraph/AnalysisScope.java#L208-L208

And also [here](https://github.com/wala/WALA/wiki/Analysis-Scope#:~:text=The%20above%20lines%20will%20include%20all%20modules,desktop%0APrimordial%2CJava%2CjarFile%2Cprimordial.jar.model) in the docs.

The `JrtModule` support might be more robust than the `jdkModule` support.  My feeling is the `Java9AnalysisScopeReader` logic should just be merged into `AnalysisScopeReader`.  

_Originally posted by @msridhar in https://github.com/wala/WALA/pull/1873#discussion_r3052946708_

msridhar:
Fixes #1876 

Before this change, we would load standard library `jmod` files for recent JDKs from a well-known place in the filesystem.  But, certain JDK distributions like Temurin no longer distribute such files directly.  This change allows WALA to run on such JDKs, by loading the standard library files via the pre-existing `JrtModule` support, which loads the files via a special virtual filesystem.  This will be a nice benefit for WALA users, since WALA will be able to load the running JVM's standard libraries out of the box on recent versions of JDKs like Temurin that don't package jmod files.  To ensure this support doesn't regress, we now install Temurin JDKs by default on CI, and we no longer specifically request Zulu JDKs (which include jmod files) in our Java toolchain configuration.

There are some API-incompatible changes in this PR.  In particular, `WalaProperties.getJ2SEJarFiles` now can throw a `NoJDKLibraryFilesFoundException` if there are no standard library files distributed with the JDK, in which case client code should fall back on calling `PlatformUtil.getJDKModuleNames` and then loading the modules via `JrtModule`.  It would be nice to have a single clean code path for handling all these cases, but it's not straightforward, given we still want to support loading standard libraries specified in scope files.

There are a couple of cases where tests passed without even loading the standard library; so for now, I just left TODO comments on those to do more detailed handling in the future.

Other related changes:

* Moved the functionality from `Java9AnalysisScopeReader` into `AnalysisScopeReader`, and deprecated the former.
* This PR exposed another bug which is that Shrike did not yet support `CONSTANT_Dynamic` constant pool entries (exposed since these changes caused all library modules to be loaded on modern JDKs for the first time); add basic support for those.

### DIFF
--- a/cast/java/ecj/src/main/java/com/ibm/wala/cast/java/ecj/util/SourceDirCallGraph.java
+++ b/cast/java/ecj/src/main/java/com/ibm/wala/cast/java/ecj/util/SourceDirCallGraph.java
@@ -26,6 +26,7 @@ import com.ibm.wala.properties.WalaProperties;
 import com.ibm.wala.ssa.SymbolTable;
 import com.ibm.wala.types.ClassLoaderReference;
 import com.ibm.wala.util.PlatformUtil;
+import com.ibm.wala.util.PlatformUtil.NoJDKModulesFoundException;
 import com.ibm.wala.util.io.CommandLine;
 import java.io.File;
 import java.io.IOException;
@@ -81,12 +82,12 @@ public class SourceDirCallGraph {
     String mainClass = p.getProperty("mainClass");
     AnalysisScope scope = new JavaSourceAnalysisScope();
     // add standard libraries to scope
-    String[] stdlibs = WalaProperties.getJ2SEJarFiles();
-    if (stdlibs.length > 0) {
+    try {
+      String[] stdlibs = WalaProperties.getJ2SEJarFiles();
       for (String stdlib : stdlibs) {
         scope.addToScope(ClassLoaderReference.Primordial, new JarFile(stdlib));
       }
-    } else {
+    } catch (NoJDKModulesFoundException e) {
       for (String moduleName : PlatformUtil.getJDKModuleNames(false)) {
         scope.addJDKModuleToScope(ClassLoaderReference.Primordial, moduleName);
       }
--- a/cast/java/ecj/src/main/java/com/ibm/wala/cast/java/ecj/util/SourceDirCallGraph.java
+++ b/cast/java/ecj/src/main/java/com/ibm/wala/cast/java/ecj/util/SourceDirCallGraph.java
@@ -22,16 +22,12 @@ import com.ibm.wala.ipa.callgraph.impl.Util;
 import com.ibm.wala.ipa.cha.ClassHierarchyException;
 import com.ibm.wala.ipa.cha.ClassHierarchyFactory;
 import com.ibm.wala.ipa.cha.IClassHierarchy;
-import com.ibm.wala.properties.WalaProperties;
 import com.ibm.wala.ssa.SymbolTable;
 import com.ibm.wala.types.ClassLoaderReference;
-import com.ibm.wala.util.PlatformUtil;
-import com.ibm.wala.util.PlatformUtil.NoJDKLibraryFilesFoundException;
 import com.ibm.wala.util.io.CommandLine;
 import java.io.File;
 import java.io.IOException;
 import java.util.Properties;
-import java.util.jar.JarFile;
 
 /**
  * Driver that constructs a call graph for an application specified as a directory of source code.
@@ -82,16 +78,7 @@ public class SourceDirCallGraph {
     String mainClass = p.getProperty("mainClass");
     AnalysisScope scope = new JavaSourceAnalysisScope();
     // add standard libraries to scope
-    try {
-      String[] stdlibs = WalaProperties.getJ2SEJarFiles();
-      for (String stdlib : stdlibs) {
-        scope.addToScope(ClassLoaderReference.Primordial, new JarFile(stdlib));
-      }
-    } catch (NoJDKLibraryFilesFoundException e) {
-      for (String moduleName : PlatformUtil.getJDKModuleNames(false)) {
-        scope.addJDKModuleToScope(ClassLoaderReference.Primordial, moduleName);
-      }
-    }
+    scope.addStdLibs(false, ClassLoaderReference.Primordial);
     // add the source directory
     File root = new File(sourceDir);
     if (root.isDirectory()) {
--- a/cast/java/ecj/src/main/java/com/ibm/wala/cast/java/ecj/util/SourceDirCallGraph.java
+++ b/cast/java/ecj/src/main/java/com/ibm/wala/cast/java/ecj/util/SourceDirCallGraph.java
@@ -26,7 +26,7 @@ import com.ibm.wala.properties.WalaProperties;
 import com.ibm.wala.ssa.SymbolTable;
 import com.ibm.wala.types.ClassLoaderReference;
 import com.ibm.wala.util.PlatformUtil;
-import com.ibm.wala.util.PlatformUtil.NoJDKModulesFoundException;
+import com.ibm.wala.util.PlatformUtil.NoJDKLibraryFilesFoundException;
 import com.ibm.wala.util.io.CommandLine;
 import java.io.File;
 import java.io.IOException;
@@ -87,7 +87,7 @@ public class SourceDirCallGraph {
       for (String stdlib : stdlibs) {
         scope.addToScope(ClassLoaderReference.Primordial, new JarFile(stdlib));
       }
-    } catch (NoJDKModulesFoundException e) {
+    } catch (NoJDKLibraryFilesFoundException e) {
       for (String moduleName : PlatformUtil.getJDKModuleNames(false)) {
         scope.addJDKModuleToScope(ClassLoaderReference.Primordial, moduleName);
       }
--- a/cast/java/ecj/src/main/java/com/ibm/wala/cast/java/ecj/util/SourceDirCallGraph.java
+++ b/cast/java/ecj/src/main/java/com/ibm/wala/cast/java/ecj/util/SourceDirCallGraph.java
@@ -25,6 +25,7 @@ import com.ibm.wala.ipa.cha.IClassHierarchy;
 import com.ibm.wala.properties.WalaProperties;
 import com.ibm.wala.ssa.SymbolTable;
 import com.ibm.wala.types.ClassLoaderReference;
+import com.ibm.wala.util.PlatformUtil;
 import com.ibm.wala.util.io.CommandLine;
 import java.io.File;
 import java.io.IOException;
@@ -81,8 +82,14 @@ public class SourceDirCallGraph {
     AnalysisScope scope = new JavaSourceAnalysisScope();
     // add standard libraries to scope
     String[] stdlibs = WalaProperties.getJ2SEJarFiles();
-    for (String stdlib : stdlibs) {
-      scope.addToScope(ClassLoaderReference.Primordial, new JarFile(stdlib));
+    if (stdlibs.length > 0) {
+      for (String stdlib : stdlibs) {
+        scope.addToScope(ClassLoaderReference.Primordial, new JarFile(stdlib));
+      }
+    } else {
+      for (String moduleName : PlatformUtil.getJDKModuleNames(false)) {
+        scope.addJDKModuleToScope(ClassLoaderReference.Primordial, moduleName);
+      }
     }
     // add the source directory
     File root = new File(sourceDir);
--- a/cast/java/ecj/src/main/java/com/ibm/wala/cast/java/translator/jdt/ecj/ECJSourceModuleTranslator.java
+++ b/cast/java/ecj/src/main/java/com/ibm/wala/cast/java/translator/jdt/ecj/ECJSourceModuleTranslator.java
@@ -89,17 +89,8 @@ public class ECJSourceModuleTranslator implements SourceModuleTranslator {
    * while {@code includeRunningVMBootclasspath} is enabled when the scope is backed by {@link
    * JrtModule} entries instead of jar or jmod paths.
    */
-  private static final class ClassPath {
-    private final String[] sources;
-    private final String[] libs;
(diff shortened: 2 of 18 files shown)


msridhar:
@julian-certora @juliandolby FYI in case you see any issues

msridhar:
Also, reported test coverage of this patch is misleadingly low.  The issue is that some code paths are only exercised on JDK 25, but we don't upload coverage results from JDK 25 CI jobs to codecov (we only upload from JDK 17).  I wonder if we could upload coverage files from multiple CI jobs and it would work?  But I think we should explore that separately.

core/src/main/java/com/ibm/wala/core/util/config/AnalysisScopeReader.java
```diff
@@ -229,12 +231,18 @@ public void processScopeDefLine(AnalysisScope scope, ClassLoader javaLoader, Str
...
-        scope.addToScope(walaLoader, new JarFile(stdlib, false));
+      try {
+        String[] stdlibs = WalaProperties.getJDKLibraryFiles(justBase);
+        for (String stdlib : stdlibs) {
+          scope.addToScope(walaLoader, new JarFile(stdlib, false));
+        }
+      } catch (NoJDKLibraryFilesFoundException e) {
+        for (String moduleName : PlatformUtil.getJDKModuleNames(justBase)) {
+          scope.addJDKModuleToScope(walaLoader, moduleName);
+        }
```
liblit: This code looks similar to pending changes in `SourceDirCallGraph.doit`. Is there an opportunity for factoring out common functionality? Or are these better kept separate?

cast/java/ecj/src/main/java/com/ibm/wala/cast/java/ecj/util/SourceDirCallGraph.java
```diff
@@ -80,9 +82,15 @@ public void doit(String[] args, Processor processor)
...
-      scope.addToScope(ClassLoaderReference.Primordial, new JarFile(stdlib));
+    try {
+      String[] stdlibs = WalaProperties.getJ2SEJarFiles();
+      for (String stdlib : stdlibs) {
+        scope.addToScope(ClassLoaderReference.Primordial, new JarFile(stdlib));
+      }
+    } catch (NoJDKLibraryFilesFoundException e) {
+      for (String moduleName : PlatformUtil.getJDKModuleNames(false)) {
+        scope.addJDKModuleToScope(ClassLoaderReference.Primordial, moduleName);
+      }
```
liblit: This code looks similar to pending changes in `AnalysisScopeReader.processScopeDefLine`. Is there an opportunity for factoring out common functionality? Or are these better kept separate?

core/src/testFixtures/java/com/ibm/wala/core/tests/shrike/DynamicCallGraphTestBase.java
```diff
@@ -71,13 +72,16 @@ protected void instrument(String testJarLocation)
...
 
       String rtJar = null;
-      for (String jar : WalaProperties.getJ2SEJarFiles()) {
-        if (jar.endsWith(File.separator + "rt.jar")
-            || jar.endsWith(File.separator + "classes.jar")) {
-          rtJar = jar;
+      try {
+        for (String jar : WalaProperties.getJ2SEJarFiles()) {
+          if (jar.endsWith(File.separator + "rt.jar")
+              || jar.endsWith(File.separator + "classes.jar")) {
```
liblit, line 78: These checks make it clear that the value returned by `WalaProperties.getJ2SEJarFIles` is semantically a `Path`, not just an arbitrary `String`. Should we make that official by changing the return types of `WalaProperties.getJ2SEJarFIles` and `WalaProperties.getJDKLibraryFiles`? That's a non-backward-compatible change, but this PR are already proposes non-backward-compatible changes to those two methods by changing their `throws` declarations.

shrike/src/main/java/com/ibm/wala/shrike/shrikeBT/Constants.java
```diff
@@ -503,6 +503,8 @@ public interface Constants {
 
   byte CONSTANT_MethodType = 16;
 
+  byte CONSTANT_Dynamic = 17;
```
liblit, line 506: It cannot be a coincidence that `Constants.CONSTANT_Dynamic` and `ClassConstants.CONSTANT_Dynamic` have identical values. Should one be initialized from the other? (Of course, that same question would apply to many of these constants, so perhaps that’s best considered outside of this PR.)

shrike/src/main/java/com/ibm/wala/shrike/shrikeCT/ClassConstants.java
```diff
@@ -43,6 +43,8 @@ public interface ClassConstants {
 
   byte CONSTANT_MethodType = 16;
 
+  byte CONSTANT_Dynamic = 17;
```
liblit, line 46: It cannot be a coincidence that `Constants.CONSTANT_Dynamic` and `ClassConstants.CONSTANT_Dynamic` have identical values. Should one be initialized from the other? (Of course, that same question would apply to many of these constants, so perhaps that’s best considered outside of this PR.)

util/src/main/java/com/ibm/wala/util/PlatformUtil.java
```diff
@@ -45,24 +53,42 @@ public static boolean onIKVM() {
...
+  }
+
   /**
    * Gets the standard JDK modules shipped with the running JDK
    *
    * @param justBase if {@code true}, only include the file corresponding to the {@code java.base}
    *     module
-   * @return array of {@code .jmod} module files
-   * @throws IllegalStateException if modules cannot be found
+   * @return array of {@code .jmod} module files, or an empty array if the files cannot be loaded
```
liblit: Do we need to distinguish "files cannot be loaded" (i.e., there was a problem) from "all zero files were correctly loaded with no problems"?

util/src/main/java/com/ibm/wala/util/PlatformUtil.java
```diff
@@ -45,24 +53,42 @@ public static boolean onIKVM() {
...
+   * @throws IllegalStateException if the running JDK does not include jmod files
    */
-  public static String[] getJDKModules(boolean justBase) {
+  public static String[] getJDKModules(boolean justBase) throws NoJDKLibraryFilesFoundException {
```
liblit: Given that we are breaking backward compatibility anyway (due to the `throws` change), should we chnage this return type to `List<String>` or even `List<Path>`?

util/src/main/java/com/ibm/wala/util/PlatformUtil.java
```diff
@@ -76,13 +102,26 @@ public static String[] getJDKModules(boolean justBase) {
...
+   * Gets the standard JDK module names exposed by the running JDK image.
    *
-   * @param moduleName name of the module, e.g., {@code "java.sql"}
-   * @return path to the module
+   * @param justBase if {@code true}, only include {@code java.base}
+   * @return array of module names from the {@code jrt:/modules} filesystem
    */
-  public static Path getPathForJDKModule(String moduleName) {
-    return Paths.get(System.getProperty("java.home"), "jmods", moduleName + ".jmod");
+  public static String[] getJDKModuleNames(boolean justBase) {
```
liblit, line 61: Given that we are breaking backward compatibility anyway (due to the `throws` change), should we chnage this return type to `List<String>` or even `List<Path>`?

cast/java/ecj/src/main/java/com/ibm/wala/cast/java/translator/jdt/ecj/ECJSourceModuleTranslator.java
```diff
@@ -82,6 +82,25 @@
...
 // remove me comment: Jdt little-case = not OK, upper case = OK
 public class ECJSourceModuleTranslator implements SourceModuleTranslator {
+  /**
+   * Classpath inputs derived from an {@link AnalysisScope} for ECJ parser configuration.
+   *
+   * <p>{@code libs} and {@code sources} are passed directly to {@link ASTParser#setEnvironment},
+   * while {@code includeRunningVMBootclasspath} is enabled when the scope is backed by {@link
+   * JrtModule} entries instead of jar or jmod paths.
+   */
+  private static final class ClassPath {
```
liblit: This `class` could be a `record`. Note that this subproject already uses Java 17 for compatibility with Eclipse dependencies, so `record`s _are_ available.

liblit (changes requested):

core/src/main/java/com/ibm/wala/core/util/config/AnalysisScopeReader.java
msridhar: This is now refactored into a common routine `AnalysisScope#addStdLibs`

cast/java/ecj/src/main/java/com/ibm/wala/cast/java/ecj/util/SourceDirCallGraph.java
msridhar: See above

util/src/main/java/com/ibm/wala/util/PlatformUtil.java
msridhar: I ended up deleting this whole method, and standardizing on joading stdlib jrt files via the new virtual filesystem approach, so we have less code to maintain

util/src/main/java/com/ibm/wala/util/PlatformUtil.java
msridhar: See above, this method is now deleted

cast/java/ecj/src/main/java/com/ibm/wala/cast/java/translator/jdt/ecj/ECJSourceModuleTranslator.java
msridhar: Done!

shrike/src/main/java/com/ibm/wala/shrike/shrikeBT/Constants.java
msridhar: I agree this redundancy should be removed, but I think it's outside the scope of this PR.  I've opened #1881 to track the issue.

shrike/src/main/java/com/ibm/wala/shrike/shrikeCT/ClassConstants.java
msridhar: See above.

core/src/testFixtures/java/com/ibm/wala/core/tests/shrike/DynamicCallGraphTestBase.java
msridhar: Is it ok to address this one also in a follow-up?  I think it's a good cleanup and I'm fine with breaking backward compatibility, but it's going to involve a bunch of code churn in tests and I'd rather do it in a separate PR.

util/src/main/java/com/ibm/wala/util/PlatformUtil.java
msridhar: See above, method is deleted now

msridhar:
@liblit thanks for the careful review!  I think this is ready for another look.  Note that I deleted the `PlatformUtil.getJDKModules` method, so now we only support reading JDK library modules via the jrt filesystem support, to limit the number of code paths we need to maintain.

julian-certora:
> @julian-certora @juliandolby FYI in case you see any issues

Quite the opposite; i like the idea of using jrt systematically, and Solidity support currently doesn't use Java stdlibs.

juliandolby (dismissed):
I vaguely recall once trying to make the Java source front end work with JRT, but I never managed it.  Maybe that was an older version of Eclipse...  I like this change.

msridhar:
> I vaguely recall once trying to make the Java source front end work with JRT, but I never managed it. Maybe that was an older version of Eclipse... I like this change.

Yeah, it seems to work now, at least the regression tests pass 🙂 Codex was helpful for figuring out that one.

core/src/testFixtures/java/com/ibm/wala/core/tests/shrike/DynamicCallGraphTestBase.java
liblit: > Is it ok to address this one also in a follow-up?

Certainly!

cast/java/src/testFixtures/java/com/ibm/wala/cast/java/test/IRTests.java
```diff
@@ -72,7 +75,19 @@ public abstract class IRTests {
 
   protected boolean dump = true;
 
-  public static final List<String> rtJar = Arrays.asList(WalaProperties.getJ2SEJarFiles());
+  public static final List<String> rtJar;
+
+  static {
+    rtJar = getJavaRuntimeLibraries();
+  }
```
liblit:
```suggestion
  public static final List<String> rtJar = getJavaRuntimeLibraries();
```

liblit (changes requested):
Just one remaining nit: a `static` block that could probably be a simple field initialization expression as part of the field declaration.

cast/java/src/testFixtures/java/com/ibm/wala/cast/java/test/IRTests.java
msridhar: Done!

liblit (approved):

PR #1880, opened 2026-04-17
Always upload coverage data to Codecov

msridhar:
It seems [based on the documentation](https://docs.codecov.com/docs/merging-reports) that Codecov should just merge coverage reports if we upload from multiple jobs.  So let's give it a try.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -17,29 +17,21 @@ jobs:
         include:
           - os: macos-latest
             java: 17
-            coverage: false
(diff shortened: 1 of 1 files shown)


liblit (approved):
8 lines deleted, 1 line simplified. What's not to like? ✅

msridhar:
Looks like the [CodeCov comment](https://github.com/wala/WALA/pull/1880#issuecomment-4269198416) gets updated as more jobs upload the coverage info, which looks good.  Also looks like [measured coverage improved](https://app.codecov.io/gh/wala/WALA/pull/1880/indirect-changes) due to this PR, as coverage from more jobs is being included.

liblit:
W00t!

PR #1883, opened 2026-04-19
Deprecate redundant `DexConstants`

liblit:
This collection of constants is almost completely redundant with respect to values defined by `org.jf.dexlib2.AccessFlags` and `org.jf.dexlib2.ValueType`, with the one exception of `ACC_UNUSED`. Furthermore, no existing WALA code uses any of these constants.  I see
no reason to keep this code around.

Let's deprecate `DexConstants` for now, just in case some third-party WALA-dependent code uses any of these constants.  We can remove it entirely in some future release.

### DIFF
--- a/dalvik/src/main/java/com/ibm/wala/dalvik/classLoader/DexConstants.java
+++ b/dalvik/src/main/java/com/ibm/wala/dalvik/classLoader/DexConstants.java
@@ -48,6 +48,7 @@
 package com.ibm.wala.dalvik.classLoader;
 
 import org.jf.dexlib2.AccessFlags;
+import org.jf.dexlib2.ValueType;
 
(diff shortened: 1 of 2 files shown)


msridhar (dismissed):

liblit:
FYI, I revised the deprecated `since` attribute from `1.7.1` to `1.7.2` since @msridhar released `1.7.1` just _before_ this change went in. Thus, the first release that will officially declare these deprecations will be whatever release we do next _after_ `1.7.1`. Presumably that will be either `1.7.2` or `1.8.0`. Either way, these symbols were certainly not deprecated in the release that just went out as `1.7.1`.

msridhar (approved):

PR #1884, opened 2026-04-19
Improve type safety of analysis results

liblit:
Previously we used three `int` constants to report various yes/no/maybe analysis results.  Now we extract these constants into their own `enum`. This change improves static type checking and also enables diagnostics about non-exhaustive `switch` statements.


msridhar (approved):
Nice cleanup!  I think the original code may have been written before Java had enums...or maybe written as a low-level optimization that is no longer necessary with modern JVMs.

liblit:
> the original code may have been written before Java had enums

Yeah, that would be my guess.

PR #1887, opened 2026-04-21
Tidy up `SyntheticMethod.getStatements` APIs

liblit:
## Tidy up `SyntheticMethod.getStatements` APIs

Previously we had two overloads of `SyntheticMethod.getStatements`:

* a deprecated overload taking an `SSAOptions` parameter, with a default implementation that returned a special "no statements" value; and

* a non-deprecated overload taking no parameters, with a default implementation that called the deprecated overload with default SSA options.

In practice, no implementation of the unary overload ever actually used the SSA options. So we can simplify things by removing that overload and changing the default implementation of the nullary overload to return the special "no statements" value.

## Deprecate many unused options parameters

The unary `SyntheticMethod.getStatements(SSAOptions)` method has been deprecated in favor of the nullary `SyntheticMethod.getStatements()` method. Anything passing in an `SSAOptions` argument no longer needs to do that. This turns out to transitively make quite a few constructor parameters unnecessary, as well as a few fields. Let's migrate toward alternate constructors that no longer pass these unnecessary parameters. We won't remove the original constructors right away, though. Instead we will just mark them as deprecated and intended for future removal.

### DIFF
--- a/cast/js/rhino/src/main/java/com/ibm/wala/cast/js/examples/hybrid/Driver.java
+++ b/cast/js/rhino/src/main/java/com/ibm/wala/cast/js/examples/hybrid/Driver.java
@@ -91,7 +91,7 @@ public class Driver {
     JavaJavaScriptHybridCallGraphBuilder b =
         new JavaJavaScriptHybridCallGraphBuilder(
             new FakeRootMethod(
-                new FakeRootClass(CrossLanguageCallGraph.crossCoreLoader, cha), options, cache),
+                new FakeRootClass(CrossLanguageCallGraph.crossCoreLoader, cha), cache),
             options,
             cache);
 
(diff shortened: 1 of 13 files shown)


msridhar (commented):
So this is the commit that added the `@Deprecated` annotation:

https://github.com/wala/WALA/commit/933043fccff5d19a57d352c1412513688810290a

Based on that change, it seems the intent was to move all code over to the `getStatements()` overload instead (without the `SSAOptions` parameter).  I'm wondering if it'd be better to delete this overload (including in subclasses), rather than making it abstract and forcing subclasses to implement?

liblit:
> So this is the commit that added the `@Deprecated` annotation:
> 
> [933043f](https://github.com/wala/WALA/commit/933043fccff5d19a57d352c1412513688810290a)

Thank you for doing the extra Git archeology that I neglected, @msridhar!

> Based on that change, it seems the intent was to move all code over to the `getStatements()` overload instead (without the `SSAOptions` parameter). I'm wondering if it'd be better to delete this overload (including in subclasses), rather than making it abstract and forcing subclasses to implement?

Yes, you're absolutely right. Well done.

msridhar (approved):
LGTM!

liblit:
Thanks again for exercising good quality control here, @msridhar. This is a much better change than what I originally proposed, informed by the Git back-story that you brought up.

PR #1892, opened 2026-04-22
Fix for constant folding

juliandolby:
This is a fix to constant folding in the case of an if statement for which the test is folded to a false constant
```
if (...) {
  ...
}
```
The result should be an empty statement, but it was left as an if false.  This change fixes that and also makes a minor correction to `ArgumentSpecialization` which was needed in order to test the fix properly.


cast/js/src/main/java/com/ibm/wala/cast/js/ipa/callgraph/ArgumentSpecialization.java
```diff
@@ -218,7 +218,7 @@ private CAstNode handleArgumentRef(CAstNode n) {
                     return Ast.makeNode(CAstNode.VAR, Ast.makeConstant("$arg" + arg));
                   }
                 } else if ("length".equals(x)) {
-                  return Ast.makeConstant(v.getValue());
+                  return Ast.makeConstant(v.getValue() - 2);
```
msridhar: I see the constant `2` used multiple times in this function.  Can we store it in a local and add a comment describing what it means?

cast/js/src/test/resources/tests/args.js
```diff
@@ -2,16 +2,31 @@
 function a() {
 	if (arguments.length >= 1) {
 		arguments[0]();
-		if (arguments.length >= 2) {
+		if (arguments.length >= 3) {
 			arguments[1]();
 		}
+	    if (arguments.length >= 3) {
+			if (arguments[1] != null) {
+		    	arguments[1]();
```
msridhar, line 10: I'm not sure what this is supposed to test, since `arguments[1]` is invoked unconditionally at line 6 if `arguments.length >= 3`?

cast/js/src/test/resources/tests/args.js
```diff
@@ -2,16 +2,31 @@
...
+			} else {
+				print("null");
+			} 
+	    } else {
+			if (arguments[0] != null) {
+				arguments[0]();
```
msridhar, line 16: Also not sure what this is testing?

cast/src/main/java/com/ibm/wala/cast/ir/translator/ConstantFoldingRewriter.java
```diff
@@ -52,10 +52,12 @@ protected CAstNode copyNodes(
       CAstNode expr = copyNodes(root.getChild(0), cfg, context, nodeMap);
       if (expr.getKind() == CAstNode.CONSTANT && expr.getValue() == Boolean.TRUE) {
         result = copyNodes(root.getChild(1), cfg, context, nodeMap);
-      } else if (expr.getKind() == CAstNode.CONSTANT
-          && root.getChildCount() > 2
-          && expr.getValue() == Boolean.FALSE) {
-        result = copyNodes(root.getChild(2), cfg, context, nodeMap);
+      } else if (expr.getKind() == CAstNode.CONSTANT && expr.getValue() == Boolean.FALSE) {
+        if (root.getChildCount() > 2) {
```
msridhar, line 57: I think the check for child count greater than 2 is checking if there is an `else` case for the `if`?  If that's right could we add a comment to that effect?

msridhar (changes requested):
Just a couple minor style comments

cast/js/src/main/java/com/ibm/wala/cast/js/ipa/callgraph/ArgumentSpecialization.java
juliandolby: These are the standard function object and 'this' pointer arguments to all calls in JS, so we could make a constant like FUNC_AND_THIS_ARGS = 2...

cast/js/src/test/resources/tests/args.js
juliandolby: The only purpose is to make the rewriter to see more cases of IFs with and without constant tests.  Just a hack to improve test coverage.

cast/js/src/test/resources/tests/args.js
juliandolby: Same as above: the only purpose is to make the rewriter to see more cases of IFs with and without constant tests.

cast/src/main/java/com/ibm/wala/cast/ir/translator/ConstantFoldingRewriter.java
juliandolby: yes.  Added.

cast/js/src/main/java/com/ibm/wala/cast/js/ipa/callgraph/ArgumentSpecialization.java
juliandolby: added.

cast/js/src/test/resources/tests/args.js
msridhar: I get it.  But, in this case, the assertions for this test are unaffected, correct?  No matter whether `arguments[1] != null` is interpreted correctly by the rewriter or not, `arguments[1]` will still be a callee?  That's ok (this extra coverage still prevents crashes, etc.), I just wanted to check if you expected this to be connected in some way to the test assertions.

msridhar (approved):
One more comment, but I'm ok with landing this.  I won't auto-merge but @juliandolby you can go ahead and merge it whenever.

cast/js/src/test/resources/tests/args.js
juliandolby: Yes, the idea is that more code in the folder is covered even though the test is not impacted.

PR #1897, opened 2026-05-08
perf(dominators): empty bucket while iterating in step 2

toxamin:
bucket(parent(w)) wasn't getting cleared, so DFS siblings sharing a semidominator kept reprocessing each other's entries. paper has "delete v from bucket(parent(w))" inside the loop. fixes O(n^2) regression on wide flowgraphs.

### DIFF
--- a/util/src/main/java/com/ibm/wala/util/graph/dominators/Dominators.java
+++ b/util/src/main/java/com/ibm/wala/util/graph/dominators/Dominators.java
@@ -299,6 +299,7 @@ public abstract class Dominators<T> {
       Iterator<T> bucketEnum = iterateBucket(getParent(node));
       while (bucketEnum.hasNext()) {
         T node2 = bucketEnum.next();
+        bucketEnum.remove();
 
         // u = EVAL(node2)
         T u = EVAL(node2);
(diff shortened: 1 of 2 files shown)


util/src/main/java/com/ibm/wala/util/graph/dominators/Dominators.java
```diff
@@ -299,6 +299,7 @@ private void step2() {
       Iterator<T> bucketEnum = iterateBucket(getParent(node));
       while (bucketEnum.hasNext()) {
         T node2 = bucketEnum.next();
+        bucketEnum.remove();
```
msridhar, line 304: Can you add a comment to this line explaining why it's important?

msridhar (commented):
Thanks for the contribution!  Could you update the PR description to be a bit more detailed about the issue you have fixed?  Is this line you added from the Tarjan algorithm that was missing in our implementation?

toxamin:
The bucket is a worklist, not persistent state. In a star graph, every leaf has the root as its DFS parent and semidominator, so every leaf is added to bucket(root). Without removal, each sibling iterates all previously added siblings again: 1 + 2 + ... + n, which is quadratic. The Lengauer-Tarjan step is delete v from bucket(parent(w)); this line implements that and keeps each bucket entry processed once.

The reason it has never been noticed is because it doesn't make the results incorrect, nor does it slow down on MOST graphs.

Concretely: imagine a graph with one root node, that has 10000 edges. Every child has the root as its DFS parent and semidominator, so every child is added to bucket(root). Without removal, the first child processes 1 bucket entry, the second processes 2, the third processes 3, and so on. That is 1 + 2 + ... + 10000 = 50,005,000 bucket-loop iterations, instead of 10000. Each of those iterations calls EVAL, so this turns a linear case into a quadratic one.

msridhar (dismissed):
Thanks for this fix!

msridhar (approved):

PR #1899, opened 2026-05-08
Download Java CUP 11a from Maven Central

### LINKED ISSUE #1898: Java CUP test dependency is missing
WALA's test resources include [a specific Java CUP JAR](https://www2.cs.tum.edu/projects/cup/java-cup-11a.jar), downloaded at build time along with other dependencies. Unfortunately, that URL now yields a "404 Not Found" error page. Perhaps that is just a temporary failure, though 404 is not likely to be a transient network outage. If we think that the tests that use this artifact are still important, then we should find some other place from which to download it.

liblit:
Checksums confirm that this JAR is byte-for-byte identical to the one that we were previously downloading using an _ad hoc_ URL.

Fixes #1898.

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -173,11 +173,12 @@ val extractBcel =
 
 ////////////////////////////////////////////////////////////////////////
 //
-//  download "java-cup-11a.jar"
+//  download "java-cup-11a-RELEASE290.jar"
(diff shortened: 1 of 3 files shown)


msridhar (approved):
Thanks for fixing so quickly!

PR #1900, opened 2026-05-12
centralize processing of calls within our CPA'ish dispatch

juliandolby:
Previously, `DispatchOperator` and two places in `visitInvokeInternal` had logic for resolving calls and adding the needed constraints.  This change makes them all use the new `handleCall` function.


core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/SSAPropagationCallGraphBuilder.java
```diff
@@ -551,28 +556,17 @@ protected void rec(final int pi, final int rhsi) {
...
                 x -> {
                   keys[pi] = system.getInstanceKey(x);
                   rec(pi + 1, rhsi + 1);
                 });
-          } /*else {
-              if (!site.isDispatch() || p != 0) {
-                keys[pi] = null;
-                rec(pi + 1, rhsi + 1);
-              }
-            } */
```
msridhar, line 566: Since we're modifying this code, can we delete this commented-out else branch if we know now that we don't need it?  Or, add a comment indicating why we haven't deleted it?

core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/SSAPropagationCallGraphBuilder.java
```diff
@@ -509,23 +510,22 @@ public static List<ProgramCounter> getIncomingPEIs(IRView ir, ISSABasicBlock bb)
...
     private final CallSiteReference site;
     private final InstanceKey[] keys;
+    private final BiFunction<Integer, Integer, IntSet> getParamObjects;
 
     private CrossProductRec(
         InstanceKey[][] invariants,
         SSAAbstractInvokeInstruction call,
         CGNode caller,
-        Consumer<InstanceKey[]> f) {
+        BiConsumer<IClass, InstanceKey[]> f,
```
msridhar: Since we're editing this code, can we rename `f` to a more meaningful name?

core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/SSAPropagationCallGraphBuilder.java
```diff
@@ -509,23 +510,22 @@ public static List<ProgramCounter> getIncomingPEIs(IRView ir, ISSABasicBlock bb)
 
   private class CrossProductRec {
     private final InstanceKey[][] invariants;
-    private final Consumer<InstanceKey[]> f;
-    private final SSAAbstractInvokeInstruction call;
-    private final CGNode caller;
+    private final BiConsumer<IClass, InstanceKey[]> f;
```
msridhar: Since we're editing this code, can we rename `f` to a more meaningful name?

core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/SSAPropagationCallGraphBuilder.java
```diff
@@ -2049,37 +2030,30 @@ public boolean isComplex() {
...
+    final InstanceKey[][] invariants = invs;
+    new CrossProductRec(
+            invariants,
+            instruction,
+            caller,
+            f,
+            (paramIndex, rhsi) -> {
+              int paramVn = instruction.getUse(paramIndex);
+              PointerKey var = getPointerKeyForLocal(caller, paramVn);
+              return system.findOrCreatePointsToSet(var).getValue();
```
msridhar: I noticed that this code is not covered at all by our unit tests according to codecov.  Is there a way to add a unit test that would cover it?

msridhar (changes requested):
Thanks, @juliandolby!  Couple of minor-ish comments

core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/SSAPropagationCallGraphBuilder.java
julian-certora: I _think_ I killed it.  I certainly meant to.

core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/SSAPropagationCallGraphBuilder.java
julian-certora: good idea,  i decided that `handleCallWithSpecificInstanceKeys` is accurate.

core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/SSAPropagationCallGraphBuilder.java
julian-certora: good idea,  i decided that `handleCallWithSpecificInstanceKeys` is accurate.

core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/SSAPropagationCallGraphBuilder.java
julian-certora: this code is actually dead, so i got rid of it and explained my "logic" in some updated comments.

core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/SSAPropagationCallGraphBuilder.java
msridhar: Ah, yes you did!  My bad

msridhar (approved):
LGTM, thanks!

PR #1902, opened 2026-05-12
Target JDK 17

### LINKED ISSUE #1830: Require JDK 17 to run WALA
I think it could be time to impose this requirement.  JDK 17 was released over four years ago.  And being able to use JDK 17 language features enables niceties like records, pattern-matching instance-of, switch expressions, and text blocks.

msridhar:
Fixes #1830 

After this lands, WALA will require a JDK 17+ JVM to run.  It will still be able to analyze programs with pre-JDK-17 bytecodes.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -23,8 +23,6 @@ jobs:
           - os: windows-latest
             java: 17
           - os: ubuntu-latest
-            java: 11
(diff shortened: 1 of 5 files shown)


msridhar:
We'll have to modify the required CI jobs and remove the JDK 11 job before landing this

msridhar:
Closing in favor of #1904

PR #1904, opened 2026-05-12
Require Java 17+ across all of WALA

### LINKED ISSUE #1830: Require JDK 17 to run WALA

liblit:
Java 17 brings many improvements over Java 11 that WALA could potentially use.  However, those can come in future revisions.  This commit only makes the _minimal_ changes needed to require (and be compatible with) Java 17+.

Resolves #1830.

### DIFF
--- a/core/src/testSubjects/java/demandpa/A.java
+++ b/core/src/testSubjects/java/demandpa/A.java
@@ -41,6 +41,7 @@ class A {
 
   Object f;
 
+  @SuppressWarnings("removal")
   Object foo() {
     return new Integer(3);
   }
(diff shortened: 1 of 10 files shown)


core/src/testSubjects/java/demandpa/A.java
```diff
@@ -41,6 +41,7 @@ class A {
 
   Object f;
 
+  @SuppressWarnings("removal")
```
msridhar, line 44: Interestingly, these constructors actually _won't_ be removed after all; see https://bugs.openjdk.org/browse/JDK-8354338.  But I think we'll have to compile with `--release 25` to make these warnings go away.

msridhar (dismissed):
Thanks for putting together all these changes!

msridhar (approved):

PR #1913, opened 2026-06-01
Upgrade code formatters

liblit:
- Spotless 8.0.0 → 8.5.1
- spotless-lib-extra 4.0.0 → 4.6.1

google-java-format is selected at build time based on the Gradle JVM:

  JDK 17  → 1.28.0
  JDK 21+ → 1.35.0

No single version of google-java-format works on both.  Versions 1.29.0+ reference `JCTree$JCAnyPattern`, an inner class added in JDK 21 (JEP 441) that does not exist in JDK 17.  Furthermore, the field `JCCompilationUnit.endPositions` was removed from javac in JDK 21, so versions that access it via bytecode `getfield` (including 1.28.0) fail there too.  Version 1.35.0 was refactored to avoid the `getfield` pattern entirely.

Six `--add-exports` flags and one `--add-opens` flag in `gradle.properties` are needed for the javac internals that google-java-format accesses directly.  These are safe to apply unconditionally on all supported JDK versions.

Also replace `targetExclude("**/*")` with `target(files())` in `cast/java/test/data/build.gradle.kts`.  Both patterns exclude all files from formatting (test assertions depend on source positions), but `target(files())` is the idiomatic form for newer Spotless versions.

### DIFF
--- a/build-logic/src/main/kotlin/com/ibm/wala/gradle/java.gradle.kts
+++ b/build-logic/src/main/kotlin/com/ibm/wala/gradle/java.gradle.kts
@@ -177,4 +177,13 @@ if (gradle.parent != null) {
   }
 }
 
-spotless { java { googleJavaFormat(catalogVersion("google-java-format")) } }
+spotless {
+  java {
+    googleJavaFormat(
+        catalogVersion(
+            if (JavaVersion.current() >= JavaVersion.VERSION_21) "google-java-format-jdk21"
+            else "google-java-format"
+        )
+    )
+  }
+}
--- a/cast/java/test/data/build.gradle.kts
+++ b/cast/java/test/data/build.gradle.kts
@@ -37,10 +37,10 @@ artifacts {
   add(testJavaSourceDirectory.name, testSubjects.map { it.java.srcDirs.first() })
 }
 
-// exclude since various tests make assertions based on
-// source positions in the test inputs.  to auto-format
+// Exclude since various tests make assertions based on
+// source positions in the test inputs.  To auto-format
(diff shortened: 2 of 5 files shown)


msridhar (commented):
Hrm, I wonder if we should require a more recent JDK version for the Gradle JVM to simplify things here?  I don't think it's much to ask to require WALA developers to be running Gradle on JDK 21+.  Of course, I haven't worked out how complicated of a change that would be.  But, that may be preferable to taking on this complexity just to support running Gradle builds on JDK 17?

liblit:
> I wonder if we should require a more recent JDK version for the Gradle JVM to simplify things here? I don't think it's much to ask to require WALA developers to be running Gradle on JDK 21+.

I'm comfortable with that.

> Of course, I haven't worked out how complicated of a change that would be. But, that may be preferable to taking on this complexity just to support running Gradle builds on JDK 17?

Perhaps I should put together an alternative PR that (1) enforces that Gradle use JDK 21+ and (2) upgrades the code formatters in a simpler way based on that restriction. Then you can compare both PRs and decide which one you prefer.

msridhar:
> Perhaps I should put together an alternative PR that (1) enforces that Gradle use JDK 21+ and (2) upgrades the code formatters in a simpler way based on that restriction. Then you can compare both PRs and decide which one you prefer.

This would be great!  I feel like this alternate approach is the better way to go in the long term, if the PR isn't too hard to put together.

liblit:
#1917 shows what an upgrade of code formatters looks like if we also require that Gradle use JDK 21+.

PR #1914, opened 2026-06-02
Upgrade Rhino from 1.8.1 to 1.9.1

liblit:
- Update library version in `gradle/libs.versions.toml`.

In `RhinoToAstTranslator.java`:

- Replace deprecated `ObjectProperty.getLeft()`/`getRight()` with `getKey()`/`getValue()`. `ObjectProperty.getLeft()`/`getRight()` were removed in Rhino 1.9.1.

- Iterate using `AbstractObjectProperty` (new base class in 1.9.1) and skip non-`ObjectProperty` entries since object literals may now contain other element types.

- Handle `Token.UNDEFINED` as a new `KeywordLiteral` type.  `undefined` was previously parsed as a `Name`, but Rhino 1.9.1 introduced a dedicated token for it.

In `TypedNodeVisitor.java`:

- Reorder `instanceof` checks.  `SwitchStatement` must be tested before `Scope` because Rhino 1.9.1 changed `SwitchStatement` from extending `Jump` to extending `Scope`, making every `SwitchStatement` an instance of `Scope`.

### DIFF
--- a/cast/js/rhino/src/main/java/com/ibm/wala/cast/js/translator/TypedNodeVisitor.java
+++ b/cast/js/rhino/src/main/java/com/ibm/wala/cast/js/translator/TypedNodeVisitor.java
@@ -149,6 +149,7 @@ public abstract class TypedNodeVisitor<R, A> {
     } else if (node instanceof ReturnStatement) {
       return visitReturnStatement((ReturnStatement) node, arg);
     } else if (node instanceof SwitchStatement) {
+      // `SwitchStatement` extends `Scope`, so we must check for the former _before_ the latter.
       return visitSwitchStatement((SwitchStatement) node, arg);
     } else if (node instanceof Scope) {
       return visitScope((Scope) node, arg);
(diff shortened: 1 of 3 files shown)


cast/js/rhino/src/main/java/com/ibm/wala/cast/js/translator/TypedNodeVisitor.java
```diff
@@ -148,6 +148,8 @@ public R visit(AstNode node, A arg) {
       return visitRegExpLiteral((RegExpLiteral) node, arg);
     } else if (node instanceof ReturnStatement) {
       return visitReturnStatement((ReturnStatement) node, arg);
+    } else if (node instanceof SwitchStatement) {
+      return visitSwitchStatement((SwitchStatement) node, arg);
```
msridhar, line 153: Can we add a comment explaining why this needs to be checked before `Scope`?  Just to record since it's a bit subtle

msridhar (changes requested):
one minor comment

cast/js/rhino/src/main/java/com/ibm/wala/cast/js/translator/TypedNodeVisitor.java
liblit: Excellent suggestion. Will do.

msridhar (approved):

PR #1918, opened 2026-06-04
Mechanically migrate to enhanced `switch`

liblit:
Also remove many `continue` statements that are no longer needed after this refactor.

Also remove many `default` clauses that are recognized as statically unreachable after this refactor.

IntelliJ IDEA made nearly all changes in this commit.  The only human interventions are as follows:

* remove a `// fall through` comment in `StringStuff.java` that is no longer appropriate after this refactor

* remove `MethodPositions.ERR_UNKNOWN_REASON`, which was previously only used in a statically unreachable `default` clause

* add several `switch`-related Error Prone checks to `java.gradle.kts`


msridhar (approved):
Looks great, thanks!

PR #1922, opened 2026-06-04
Use multiline text blocks for better readability

liblit:


cast/js/rhino/src/test/java/com/ibm/wala/cast/js/test/TestRhinoSourceMap.java
```diff
@@ -42,108 +42,114 @@ public void setUp() {
...
       "Ltests/jquery_spec_test.js/anonymous__0/isEmptyDataObject",
-      "function isEmptyDataObject(obj) {\n"
-          + "        for (var name in obj) {\n"
-          + "            if (name !== \"toJSON\") {\n"
-          + "                return false;\n"
-          + "            }\n"
-          + "        }\n"
-          + "        return true;\n"
-          + "    }"
+"""
```
msridhar: In this file, can we somehow fix the indentation so that the `function` keywords are vertically aligned with the final `}`?  I guess it might change the test results, and they weren't aligned before, but it'd be nice for readability to get things aligned.

msridhar (commented):
one comment

cast/js/rhino/src/test/java/com/ibm/wala/cast/js/test/TestRhinoSourceMap.java
liblit: Will do!

msridhar (approved):

PR #1926, opened 2026-06-05
Avoid array fields in `record` classes

liblit:
For `record` classes, the compiler-generated `equals` and `hashCode` methods treat arrays shallowly. That's rarely what we want. The generated `toString` methods also do not do well with array fields.

I'm not sure whether those methods are actually ever used with `ClassPath` specifically, but I'd rather not wonder. Let's use immutable lists instead of arrays just to be safe. Let's also tell Error Prone to report an error if any instances of this antipattern arise in the future.

Note: all types and type signatures affected by this change are `private`, so this change causes no backward incompatibility problems.

### DIFF
--- a/build-logic/src/main/kotlin/com/ibm/wala/gradle/java.gradle.kts
+++ b/build-logic/src/main/kotlin/com/ibm/wala/gradle/java.gradle.kts
@@ -66,6 +66,7 @@ tasks.withType<JavaCompile>().configureEach {
       error("AnnotationPosition")
       error("AssertEqualsArgumentOrderChecker")
       error("ArgumentSelectionDefectChecker")
+      error("ArrayRecordComponent")
       error("FallThrough")
       error("IfChainToSwitch")
       error("ProtectedMembersInFinalClass")
--- a/cast/java/ecj/src/main/java/com/ibm/wala/cast/java/translator/jdt/ecj/ECJSourceModuleTranslator.java
+++ b/cast/java/ecj/src/main/java/com/ibm/wala/cast/java/translator/jdt/ecj/ECJSourceModuleTranslator.java
@@ -37,6 +37,7 @@
  */
 package com.ibm.wala.cast.java.translator.jdt.ecj;
 
+import com.google.common.collect.ImmutableList;
 import com.ibm.wala.cast.java.loader.JavaSourceLoaderImpl;
 import com.ibm.wala.cast.java.translator.Java2IRTranslator;
 import com.ibm.wala.cast.java.translator.SourceModuleTranslator;
@@ -90,7 +91,9 @@ public class ECJSourceModuleTranslator implements SourceModuleTranslator {
    * JrtModule} entries instead of jar or jmod paths.
    */
   private record ClassPath(
-      String[] sources, String[] libs, boolean includeRunningVMBootclasspath) {}
+      ImmutableList<String> sources,
+      ImmutableList<String> libs,
+      boolean includeRunningVMBootclasspath) {}
 
   protected static class ECJJavaToCAstTranslator extends JDTJava2CAstTranslator<Position> {
     public ECJJavaToCAstTranslator(
@@ -184,8 +187,8 @@ public class ECJSourceModuleTranslator implements SourceModuleTranslator {
 
   protected boolean dump;
   protected ECJSourceLoaderImpl sourceLoader;
-  private final String[] sources;
-  private final String[] libs;
+  private final ImmutableList<String> sources;
+  private final ImmutableList<String> libs;
   private final boolean includeRunningVMBootclasspath;
   private final StringFilter exclusions;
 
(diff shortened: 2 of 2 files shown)


cast/java/ecj/build.gradle.kts
```diff
@@ -7,10 +7,13 @@ plugins {
   id("com.ibm.wala.gradle.publishing")
 }
 
-walaEclipseMavenCentral.implementation(
-    "org.eclipse.equinox.common",
-    "org.eclipse.jdt.core",
-)
+walaEclipseMavenCentral {
+  compileOnly("org.eclipse.jdt.annotation")
```
msridhar: Would this work with JSpecify annotations?  I don't quite understand why we need the Eclipse JDT annotations specifically.  If possible I'd like to use the standardized JSpecify annotations everywhere.

msridhar (commented):
Good catch!  One question

cast/java/ecj/build.gradle.kts
liblit: JSpecify is _years_ overdue. I appreciate the desire to standardize on those annotations. It turns out we can! However, much of ECJ's nullness analysis was previously dormant due to the missing annotation dependency. If we tell ECJ to use JSpecify annotations, and we make those annotations available, then ECJ's nullness analysis comes to life and starts reporting many errors.

Proposal: switch ECJ to use JSpecify across all of WALA but also simultaneously disable ECJ's nullness analysis. We can turn that analysis back on gradually if we improve null-correctness of the code base in the future.

What do you think?

msridhar: > Proposal: switch ECJ to use JSpecify across all of WALA but also simultaneously disable ECJ's nullness analysis.

What does it mean to switch ECJ to use JSpecify across all of WALA?  This PR is adding a dependence to the `cast.java.ecj` module, which I think is different than modifying our ECJ config across WALA.

> We can turn that analysis back on gradually if we improve null-correctness of the code base in the future.

Sure.  Perhaps unsurprisingly I'm inclined to use NullAway 🙂 But, not opposed to also using ECJ's nullness analysis if it adds further value.

liblit: > What does it mean to switch ECJ to use JSpecify across all of WALA?

It means a large number of changes like the following:

```patch
Index: util/.settings/org.eclipse.jdt.core.prefs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/util/.settings/org.eclipse.jdt.core.prefs b/util/.settings/org.eclipse.jdt.core.prefs
--- a/util/.settings/org.eclipse.jdt.core.prefs	(revision 0936842ec40f8e3997b63db6bcee4f7537baefa0)
+++ b/util/.settings/org.eclipse.jdt.core.prefs	(revision 80689c38393ce9be3bd76f03213320fb9a1833aa)
@@ -2,13 +2,13 @@
 instance/org.eclipse.core.net/org.eclipse.core.net.hasMigrated=true
 org.eclipse.jdt.core.compiler.annotation.inheritNullAnnotations=enabled
 org.eclipse.jdt.core.compiler.annotation.missingNonNullByDefaultAnnotation=ignore
-org.eclipse.jdt.core.compiler.annotation.nonnull=org.eclipse.jdt.annotation.NonNull
+org.eclipse.jdt.core.compiler.annotation.nonnull=org.jspecify.annotations.NonNull
 org.eclipse.jdt.core.compiler.annotation.nonnull.secondary=
-org.eclipse.jdt.core.compiler.annotation.nonnullbydefault=org.eclipse.jdt.annotation.NonNullByDefault
+org.eclipse.jdt.core.compiler.annotation.nonnullbydefault=org.jspecify.annotations.NullMarked
 org.eclipse.jdt.core.compiler.annotation.nonnullbydefault.secondary=
-org.eclipse.jdt.core.compiler.annotation.nullable=org.eclipse.jdt.annotation.Nullable
+org.eclipse.jdt.core.compiler.annotation.nullable=org.jspecify.annotations.Nullable
 org.eclipse.jdt.core.compiler.annotation.nullable.secondary=
-org.eclipse.jdt.core.compiler.annotation.nullanalysis=enabled
+org.eclipse.jdt.core.compiler.annotation.nullanalysis=disabled
 org.eclipse.jdt.core.compiler.codegen.inlineJsrBytecode=enabled
 org.eclipse.jdt.core.compiler.codegen.methodParameters=do not generate
 org.eclipse.jdt.core.compiler.codegen.targetPlatform=17
```

> This PR is adding a dependence to the `cast.java.ecj` module, which I think is different than modifying our ECJ config across WALA.

Agreed. I think I'll do the ECJ-uses-JSpecify change first, as its own PR. Then I'll revisit and simplify this PR, which should no longer need the new `compileOnly("org.eclipse.jdt.annotation")` dependency.

msridhar (approved):

PR #1927, opened 2026-06-06
Standardize on JSpecify; tune ECJ null analysis

liblit:
Replace all Eclipse JDT annotation type references (`org.eclipse.jdt.annotation.NonNull`, `Nullable`, `NonNullByDefault`) with their JSpecify equivalents (`org.jspecify.annotations.NonNull`, `Nullable`, `NullMarked`) in ECJ project settings across all subprojects.

Declare `org.jspecify:jspecify:1.0.0` as an `api`-scope dependency in every module that exports annotated APIs, ensuring consumers receive the annotation JAR transitively as JSpecify recommends. In `:ide`, promote from `implementation` to `api` and suppress a spurious configuration warning from the `dependency-analysis` plugin.

Enable ECJ's built-in null analysis only in subprojects where it passes for every source set without errors; disable it everywhere else. This pattern avoids compilation failures while preserving nullness-check coverage where it already works.

Backward compatibility: none of the annotation types referenced in ECJ settings affect runtime behavior. Consumers that declare their own JSpecify dependency at a different version are unaffected since Gradle's conflict resolution favors the consumer's version.

Start gathering notes in a `RELEASE-NOTES.md` file. The intent is that this file should help us write the release notes for the next official release. We could reset this file to empty after each such release, or we could just let it grow over time. I'm undecided on that matter.


build-logic/src/main/kotlin/com/ibm/wala/gradle/java.gradle.kts
```diff
@@ -38,6 +38,14 @@ configurations {
 }
 
 dependencies {
+
+  // Always make JSpecify annotations available at compile time for use by static checkers
+  catalogLibrary("jspecify").let {
+    compileOnly(it)
```
msridhar: The [general recommendation](https://jspecify.dev/docs/using/#gradle) is to put JSpecify in the `api` configuration; that's [what we did](https://github.com/wala/WALA/blob/8812a59bca362fb345cc677ea749f675d4e6c46c/util/build.gradle.kts#L11) before for `wala.util`.  Probably we should at least make this one `api(it)`?

cast/java/test/data/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -2,13 +2,13 @@ eclipse.preferences.version=1
...
+org.eclipse.jdt.core.compiler.annotation.nonnull=org.jspecify.annotations.NonNull
 org.eclipse.jdt.core.compiler.annotation.nonnull.secondary=
-org.eclipse.jdt.core.compiler.annotation.nonnullbydefault=org.eclipse.jdt.annotation.NonNullByDefault
+org.eclipse.jdt.core.compiler.annotation.nonnullbydefault=org.jspecify.annotations.NullMarked
 org.eclipse.jdt.core.compiler.annotation.nonnullbydefault.secondary=
-org.eclipse.jdt.core.compiler.annotation.nullable=org.eclipse.jdt.annotation.Nullable
+org.eclipse.jdt.core.compiler.annotation.nullable=org.jspecify.annotations.Nullable
 org.eclipse.jdt.core.compiler.annotation.nullable.secondary=
-org.eclipse.jdt.core.compiler.annotation.nullanalysis=disabled
+org.eclipse.jdt.core.compiler.annotation.nullanalysis=enabled
```
msridhar, line 11: Do we mean to enable the checking here?

cast/js/nodejs/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -2,13 +2,13 @@ eclipse.preferences.version=1
...
+org.eclipse.jdt.core.compiler.annotation.nonnull=org.jspecify.annotations.NonNull
 org.eclipse.jdt.core.compiler.annotation.nonnull.secondary=
-org.eclipse.jdt.core.compiler.annotation.nonnullbydefault=org.eclipse.jdt.annotation.NonNullByDefault
+org.eclipse.jdt.core.compiler.annotation.nonnullbydefault=org.jspecify.annotations.NullMarked
 org.eclipse.jdt.core.compiler.annotation.nonnullbydefault.secondary=
-org.eclipse.jdt.core.compiler.annotation.nullable=org.eclipse.jdt.annotation.Nullable
+org.eclipse.jdt.core.compiler.annotation.nullable=org.jspecify.annotations.Nullable
 org.eclipse.jdt.core.compiler.annotation.nullable.secondary=
-org.eclipse.jdt.core.compiler.annotation.nullanalysis=disabled
+org.eclipse.jdt.core.compiler.annotation.nullanalysis=enabled
```
msridhar, line 11: Same question here, do we mean to enable the checking

msridhar (commented):
Couple of comments

build-logic/src/main/kotlin/com/ibm/wala/gradle/java.gradle.kts
liblit: > The [general recommendation](https://jspecify.dev/docs/using/#gradle) is to put JSpecify in the `api` configuration

Ah, that makes sense given that JSpecify's annotations are all `@Retention(RUNTIME)`. I'm surprised that our dependency-analysis plugin (via `./gradlew buildHealth`) didn't catch this.

> Probably we should at least make this one `api(it)`?

I don't want to diverge from JSpecify's recommendation. But it feels too blunt to add an `api` JSpecify dependency to _everything_ given that we're disabling ECJ's nullness analysis nearly everywhere anyway.

I propose the following. In every subproject, I'll reconfigure ECJ to use JSpecify's nullness annotations. I'll also add `api` JSpecify dependencies to subprojects where it is required in order for ECJ's nullness analyses to succeed. However, if ECJ's nullness analysis fails even with the JSpecify dependency, then I'll just disable that analysis.

cast/java/test/data/.settings/org.eclipse.jdt.core.prefs
liblit: Yes. ECJ's nullness analysis succeeds here. So we may as well keep running it to catch any future regressions in nullness management.

cast/js/nodejs/.settings/org.eclipse.jdt.core.prefs
liblit: Yes. ECJ's nullness analysis succeeds here. So we may as well keep running it to catch any future regressions in nullness management.

build-logic/src/main/kotlin/com/ibm/wala/gradle/java.gradle.kts
msridhar: > I propose the following. In every subproject, I'll reconfigure ECJ to use JSpecify's nullness annotations. I'll also add `api` JSpecify dependencies to subprojects where it is required in order for ECJ's nullness analyses to succeed. However, if ECJ's nullness analysis fails even with the JSpecify dependency, then I'll just disable that analysis.

What do we mean here by "succeed"?  Does that mean "not crash" or "not report any warnings"?  In any case, one thing is for any subproject where the code actually uses JSpecify annotations (particularly on public APIs), we should be sure to include the JSpecify annotations as an `api` dependence.

liblit: > What do we mean here by "succeed"? Does that mean "not crash" or "not report any warnings"?

I mean that the Gradle ECJ compilation task for a given source set succeeds.

Put differently:

1. If a source set uses JSpecify, I will ensure that this source set has an `api` dependency on JSpecify.

1. Otherwise, I'll add JSpecify `api` dependencies where doing so is both necessary and sufficient for ECJ compilation to succeed with nullness analysis enabled.

1. Otherwise, I'll omit JSpecify `api` dependencies and disable ECJ's nullness analysis.

> In any case, one thing is for any subproject where the code actually uses JSpecify annotations (particularly on public APIs), we should be sure to include the JSpecify annotations as an api dependence.

Agreed.

msridhar (approved):

liblit:
Thanks for holding this change to a high standard of quality (as always), @msridhar!

PR #1930, opened 2026-06-08
Remove redundant `System.arraycopy`

liblit:
In the patch-application loop, every element that `System.arraycopy` copies from `p.code` into `instructions[p.start…]` is immediately overwritten by the `for` loop that follows. The pre-loop copy is unnecessary.


msridhar (approved):
Curious, did IntelliJ find this one?  Nice fix!

liblit:
> Curious, did IntelliJ find this one?

It did not.

> Nice fix!

Thanks!

PR #1932, opened 2026-06-10
Convert many classes to Java 16+ records

liblit:
Convert approximately 100 classes across the codebase to Java 16+ record classes.  Each converted class that is `public` has corresponding deprecated backward-compatible accessor bridge methods:(for example, `getType()` delegating to `type()`), annotated with `@Deprecated(forRemoval = true, since = "1.8.0")`.

Remove explicit `hashCode` and/or `equals` methods where the default `record` implementations are equivalent.

Update `RELEASE-NOTES.md` with a migration table listing all affected types, their old `getXyz()` accessors, and their new record-style accessors.


cast/src/main/java/com/ibm/wala/cast/ir/ssa/AbstractSSAConversion.java
```diff
@@ -267,16 +267,9 @@ private void renameVariables() {
    * avoiding stack overflow.
    *
    * @author yinnonh
+   * @param i iterator o
```
msridhar: maybe just delete this?  i don't understand it

msridhar (commented):
Could you possibly put together a list of cases where arrays have been changed into `List`s?  I'm slightly concerned about potential performance impact there due to extra allocations / indirection (e.g., assuming it's an `ArrayList`).  Sadly, we don't have a performance regression suite to test this properly.  But I'd just like to look those cases specifically to eyeball them more closely

liblit:
> Could you possibly put together a list of cases where arrays have been changed into `List`s?

I believe that the affected classes are:

- `AllocationString`
- `AnnotationsReader.ArrayElementValue`
- `AstMethod.DebuggingInformation`
- `AstTranslator.AstDebuggingInformation`
- `ClassHierarchyStore`
- `ClassHierarchyStore.ClassInfo`
- `Compiler.HelperPatch`
- `DexSSABuilder.SSA2LocalMap`
- `SSABuilder.SSA2LocalMap`
- `StackMapConstants.StackMapFrame`

Disclaimer: this list was machine generated. I've checked it partially, but have not done a full manual review for any omissions.

> I'm slightly concerned about potential performance impact there due to extra allocations / indirection (e.g., assuming it's an `ArrayList`).

That's a perfectly valid concern. My migrations were driven by a blanket "no array components in `record`s" policy. But arrays _can_ be used in `record`s; you just need to be extra careful that you understand the implications on `hashCode` and `equals`. So if we think that some of the above should be left as arrays, then we can do that.

cast/src/main/java/com/ibm/wala/cast/ir/ssa/AbstractSSAConversion.java
liblit: Before this pull request, the code that declared `I` as a field included the same comment:

https://github.com/wala/WALA/blob/d40437a1db02d9c8fca10c8db0a5060b0a5a36d6/cast/src/main/java/com/ibm/wala/cast/ir/ssa/AbstractSSAConversion.java#L271-L273

I think the comment was equally opaque before. OK, we may as well remove it now.

core/src/main/java/com/ibm/wala/ssa/SSABuilder.java
```diff
@@ -1009,47 +1012,41 @@ public ShrikeIndirectionData getIndirectionData() {
...
-     */
-    private final IntPair[] localStoreMap;
-
-    /**
-     * For each basic block i and local j, block2LocalState[i][j] gives the contents of local j at
-     * the start of block i
-     */
-    private final int[][] block2LocalState;
+  private record SSA2LocalMap(
+      ShrikeCFG shrikeCFG, List<IntPair> localStoreMap, List<int[]> block2LocalState)
```
msridhar: I think maybe we should leave this one as using arrays.  I could see this as being performance critical when constructing IRs.  And, if we still have a `List<int[]>` I'm not sure the deep equality issue is solved anyway.  Before this class did not override `SSA2LocalMap` so I think the record conversion should not impact correctness.

dalvik/src/main/java/com/ibm/wala/dalvik/ssa/DexSSABuilder.java
```diff
@@ -1433,35 +1436,27 @@ public ShrikeIndirectionData getIndirectionData() {
...
-    /**
-     * For each basic block i and local j, block2LocalState[i][j] gives the contents of local j at
-     * the start of block i
-     */
-    private final int[][] block2LocalState;
-
-    /** maximum number of locals used at any program point */
-    private final int maxLocals;
+  private record SSA2LocalMap(
+      DexCFG dexCFG, List<IntPair> localStoreMap, List<int[]> block2LocalState, int maxLocals)
```
msridhar: For similar reasons to `SSA2LocalMap` in `SSABuilder`, maybe let's leave this one as using arrays for its components?

core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/cfa/AllocationString.java
```diff
@@ -13,56 +13,55 @@
...
+import java.util.List;
 
 /**
  * This is a {@link ContextItem} that records n allocation sites, the 0th element represents the
  * most recently used receiver obj, which is an {@code AllocationSiteInNode}
  */
-public class AllocationString implements ContextItem {
-
-  private final AllocationSite[] allocationSites;
+public record AllocationString(List<AllocationSite> allocationSites) implements ContextItem {
```
msridhar: Let's leave this one as a class with an array field, for potential perf reasons?  I could see these context objects getting allocated a lot during pointer analysis

msridhar (commented):
I've commented on a couple instances of conversion of arrays to lists.  I haven't looked closely at all of them, but, I wonder if out of a general caution, we could not do any array to list conversions in this PR?  In cases where the containing object didn't override equals / hashCode before, we could still convert to a record; otherwise, we would leave as a class.

Sorry, this might be a bit annoying to (un)do.  I'm just concerned about performance, which I know was tuned very carefully over the years, and I'm a bit scared of unexpected impacts here.  When I get some time, I should really put together some JMH benchmarks at least for for call graph construction so we can measure this stuff.

liblit:
> I wonder if out of a general caution, we could not do any array to list conversions in this PR?

Sure, I can put that together.

> Sorry, this might be a bit annoying to (un)do.

Eh, no worries. This isn't about my convenience; it's about the long-term health of the project.

> When I get some time, I should really put together some JMH benchmarks at least for for call graph construction so we can measure this stuff.

Big +1!

liblit:
FYI, reverting just the array-to-`List` migrations turns out to be fairly easy. I'm glad that you (@msridhar) suggested it.

core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/FilteredPointerKey.java
```diff
@@ -32,31 +32,25 @@ boolean addInverseFiltered(
...
-    }
-
-    @Override
-    public int hashCode() {
-      return concreteType.hashCode();
+      return concreteType();
     }
 
     @Override
     public boolean equals(Object o) {
```
msridhar: Do we still need this `equals` override?

msridhar (dismissed):
Overall looks great, thanks for this!  Noticed one minor thing, but it may not need any fix

core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/FilteredPointerKey.java
msridhar: I disabled auto-merge in case this needs to be addressed

liblit: > Do we still need this `equals` override?

I believe that we do not. Good attention to detail!

msridhar (approved):

liblit:
@msridhar, thank you for your attentive reviewing and excellent judgment. This was a big PR, and I am pleased with how we're landing it. Go team!

PR #1935, opened 2026-06-10
Restructure list of Error Prone checks

liblit:
The order in which we list these checks is arbitrary.  Sorting them alphabetically makes it easier to scan for existing items or to find duplicates.  This ordering also reduces the risk of Git merge conflicts that would arise if always appending to the end.

Also, since Error Prone's `error` method is variadic, use one call for all of them just to reduce visual noise.

Note that this commit does not add or remove any checks.  It just restructures what was already here.


msridhar (approved):
👍

PR #1937, opened 2026-06-11
Use Java 16+ `instanceof` pattern variables

liblit:
These changes were made automatically using IntelliJ IDEA.  If we trust that tool, then we may be willing to accept these changes with only cursory review.

These changes yield a net reduction of 418 lines of code.  Nice.


msridhar (approved):
Looks great!  There is also an [Error Prone PatternMatchingInstanceOf](https://errorprone.info/bugpattern/PatternMatchingInstanceof) check we may want to enable, though that can be done in a follow up.

liblit:
> Looks great!

Thanks!

> There is also an [Error Prone PatternMatchingInstanceOf](https://errorprone.info/bugpattern/PatternMatchingInstanceof) check we may want to enable, though that can be done in a follow up.

Ah, very nice. I'm trying to get in the habit of enabling Error Prone analogues of any IntelliJ IDEA inspections that I correct. But I outright forgot to look for one here.

Even after this PR, a few instances of this pattern remain. IntelliJ IDEA finds them but fails to fix them. So I already planned a follow-up PR to fix those by hand. I'll incorporate this Error Prone check into that PR. Thank you for pointing it out!

PR #1948, opened 2026-06-14
Use `forEachRemaining` instead of explicit loops

liblit:
When we need to do something for each item remaining in an `Iterator`, I prefer to use a high-level operation like `forEachRemaining` instead of the boilerplate `hasNext()`/`next()` loop.

### DIFF
--- a/core/src/main/java/com/ibm/wala/ipa/callgraph/impl/Util.java
+++ b/core/src/main/java/com/ibm/wala/ipa/callgraph/impl/Util.java
@@ -334,9 +334,7 @@ public class Util {
       throw new IllegalArgumentException("Null x");
     }
     Set<T> y = HashSetFactory.make();
-    while (x.hasNext()) {
-      y.add(x.next());
-    }
+    x.forEachRemaining(y::add);
     return y;
   }
 
--- a/util/src/main/java/com/ibm/wala/util/graph/traverse/BFSPathFinder.java
+++ b/util/src/main/java/com/ibm/wala/util/graph/traverse/BFSPathFinder.java
@@ -93,9 +93,7 @@ public class BFSPathFinder<T> {
       throw new IllegalArgumentException("targets is null");
     }
     final Set<T> ts = HashSetFactory.make();
-    while (targets.hasNext()) {
-      ts.add(targets.next());
-    }
+    targets.forEachRemaining(ts::add);
 
     this.G = G;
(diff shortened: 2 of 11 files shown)


core/src/main/java/com/ibm/wala/ipa/callgraph/impl/Util.java
```diff
@@ -10,6 +10,8 @@
  */
 package com.ibm.wala.ipa.callgraph.impl;
 
+import static com.ibm.wala.util.collections.HashSetFactory.make;
```
msridhar: Can we not add this static import?  I'd prefer the call to still show `HashSetFactory.make()`, just seeing `make()` it's a bit hard to see what is going on.

util/src/main/java/com/ibm/wala/util/graph/traverse/BFSPathFinder.java
```diff
@@ -10,8 +10,9 @@
  */
 package com.ibm.wala.util.graph.traverse;
 
+import static com.ibm.wala.util.collections.HashSetFactory.make;
```
msridhar: Same static import comment as above

msridhar (changes requested):
minor comments

core/src/main/java/com/ibm/wala/ipa/callgraph/impl/Util.java
liblit: Ah, right. That was an unnecessary change from the original. Will fix.

util/src/main/java/com/ibm/wala/util/graph/traverse/BFSPathFinder.java
liblit: Ah, right. That was an unnecessary change from the original. Will fix.

util/src/main/java/com/ibm/wala/util/graph/traverse/DFS.java
```diff
@@ -10,9 +10,10 @@
  */
 package com.ibm.wala.util.graph.traverse;
 
+import static com.ibm.wala.util.collections.HashSetFactory.make;
+
```
liblit: I'll fix this static import too, like the others that @msridhar already pointed out.

msridhar (approved):

PR #1950, opened 2026-06-15
Deprecate `Iterator2List` and `Iterator2Set`

liblit:
We retain the static `Iterator2Collection.toList` and `Iterator2Collection.toSet` methods.  However, these now make weaker promises as to the types they return.  Internally, these are now implemented using existing Guava utility methods.  The collections that result are slightly more efficient than before, as they no longer route all `Collection` methods through an unnecessary proxy object.

That being said, the underlying `Collection` types are the same as before: `toList` continues to be based on `ArrayList`, and `toSet` continues to be based on `LinkedHashSet`.  But now we get there using less of our own code and fewer delegation layers.

Note that Guava is now an implementation dependency of our `utils` component.  Other parts of WALA were already using Guava, so this change should be easy for third-party code to handle.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -19,6 +19,31 @@ exported API surfaces](https://jspecify.dev/docs/using/#gradle).
 
 ## API changes
 
+### `Iterator2List` and `Iterator2Set` deprecated for removal
+
+The classes `Iterator2List` and `Iterator2Set`, along with all their methods
+and constructors, are deprecated for removal.
+
+The static methods `Iterator2Collection.toList(Iterator)` and
+`Iterator2Collection.toSet(Iterator)` are retained but now return plain
+`List<T>` and `Set<T>` respectively instead of `Iterator2List<T>` and
+`Iterator2Set<T>` wrappers.
+
+**Effect for third-party consumers:**
+
+- Code that uses `Iterator2List<T>` or `Iterator2Set<T>` as variable types,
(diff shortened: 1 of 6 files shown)


RELEASE-NOTES.md
```diff
@@ -19,6 +19,32 @@ exported API surfaces](https://jspecify.dev/docs/using/#gradle).
...
+
+- Code that calls `Iterator2Collection.toList(...)` or
+  `Iterator2Collection.toSet(...)` continues to work, but may need to
+  update type declarations if they were explicitly typed as
+  `Iterator2List<T>` or `Iterator2Set<T>`.
+
+- The `:util` module now declares Guava as an `implementation`-scope
+  dependency. If your project depends on `:util` and also uses Guava
+  directly, you may need to add an explicit dependency on Guava if it
+  was previously only available via WALA's own dependency chain.
```
msridhar: I don't understand this one.  Did we previously expose a dependency on Guava that we are now hiding?

msridhar (commented):
one question

RELEASE-NOTES.md
liblit: Yeah, that text would only make sense if `:util` previously had an `api` dependency on Guava that was now becoming an `implementation` dependency instead. I'll revise this text and request a new review from you. Thanks for calling this out.

msridhar (approved):

PR #1956, opened 2026-06-22
Avoid labeled break targets

liblit:
[Error Prone's `LabelledBreakTarget` posits that "labels should only be used on loops."](https://errorprone.info/bugpattern/LabelledBreakTarget) Previously we disabled this check, but there aren't very many violations of it. We may as well restructure the few violators and let this check report errors per its default setting.

Also add a few line breaks in the build logic to make it easier to see which comments apply to which bits of Gradle/Kotlin code.

### DIFF
--- a/shrike/src/main/java/com/ibm/wala/shrike/cg/Runtime.java
+++ b/shrike/src/main/java/com/ibm/wala/shrike/cg/Runtime.java
@@ -130,39 +130,37 @@ public class Runtime {
       if (runtime.output != null) {
         String caller = runtime.callStacks.get().peek();
 
-        {
-          //
-          // check for expected caller
-          //
-          boolean handled = false;
-          if (runtime.handleCallback != null) {
-            StackTraceElement[] stack = new Throwable().getStackTrace();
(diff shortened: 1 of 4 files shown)


shrike/src/main/java/com/ibm/wala/shrike/cg/Runtime.java
```diff
@@ -130,11 +130,11 @@ public static void execution(String klass, String method, Object receiver) {
       if (runtime.output != null) {
         String caller = runtime.callStacks.get().peek();
 
-        checkValid:
         {
```
msridhar: Do we still need the block starting with this open brace?

msridhar (commented):
Looks good!  One minor question

shrike/src/main/java/com/ibm/wala/shrike/cg/Runtime.java
liblit: We do not. Thanks for noticing!

msridhar (approved):

PR #1958, opened 2026-06-23
Register summary-modeled class shells before CHA build (#1957)

### LINKED ISSUE #1957: Class hierarchy can't express a source class extending a type modeled only by an XML method summary
WALA materializes bypass/summary synthetic classes ([`BypassSyntheticClassLoader`](https://github.com/wala/WALA/blob/81b08c9df085ca8a4dcc848188079c8d7eea2a1a/core/src/main/java/com/ibm/wala/ipa/summaries/BypassSyntheticClassLoader.java#L57), populated via [`XMLMethodSummaryReader`](https://github.com/wala/WALA/blob/81b08c9df085ca8a4dcc848188079c8d7eea2a1a/core/src/main/java/com/ibm/wala/ipa/summaries/XMLMethodSummaryReader.java#L62)) when bypass logic is added ([`Util.addBypassLogic`](https://github.com/wala/WALA/blob/81b08c9df085ca8a4dcc848188079c8d7eea2a1a/core/src/main/java/com/ibm/wala/ipa/callgraph/impl/Util.java#L99)) during call-graph-builder setup, which runs after [`IClassHierarchy` construction](https://github.com/wala/WALA/blob/81b08c9df085ca8a4dcc848188079c8d7eea2a1a/core/src/main/java/com/ibm/wala/client/AbstractAnalysisEngine.java#L161). Consequently, a source class that `extends` a summary-modeled class cannot carry that inheritance edge: when the hierarchy is built, the summary class does not yet exist, the supertype resolves to a `MissingType`, and the subclass falls back to `Object`.

This rarely surfaces in Java analysis, where subclassable framework types are real classes resolved from the scope's loaders at CHA-build time. (XML method summaries there model the behavior of leaf methods on classes that already exist, not base classes one extends.) It does surface for front-ends that model framework base types purely as summaries and have user code subclass them, such as a source-language front-end that models a framework base class with a summary and analyzes user code that subclasses it.

## Minimal Repro Recipe

1. Register a class only via a bypass `MethodSummary`: a `<class>` whose methods come from `XMLMethodSummaryReader` and land in the `BypassSyntheticClassLoader` when bypass logic is added. Call it `Base`.
2. Analyze a source class `class Sub extends Base {}` whose `Base` resolves only to that summary class.
3. `cha.lookupClass(Sub).getSuperclass()` gives `Object` (actual) rather than `Base` (expected), because the summary `Base` was absent at CHA-build time.

The ordering is the cause: step 1's class is registered after [`buildClassHierarchy()`](https://github.com/wala/WALA/blob/81b08c9df085ca8a4dcc848188079c8d7eea2a1a/core/src/main/java/com/ibm/wala/client/AbstractAnalysisEngine.java#L278) already ran for step 2.

## Manifestation in Practice

As one concrete instance, a Python/TensorFlow analysis registers its framework summaries (via `addSummaryBypassLogic`) only after `buildClassHierarchy()` has already run the source `defineType` over user code, so a user `class GCN(tf.keras.layers.Layer)` falls back to `Object`. The cause is general: any front-end that registers summary bypass logic after the class hierarchy is built hits the same ordering.

## Suggested Remedy

Allow summary-modeled class shells to be registered into a scope loader before the CHA is built, the way stub classes are, so source subclasses resolve their base at definition time. The late method-summary bypass step can keep supplying the method bodies; only the class shells need to exist earlier.

khatchad:
Fixes #1957.

## Problem

A source class that `extends` a type modeled only by an XML method summary cannot carry that inheritance edge. The summary class is materialized into the [`BypassSyntheticClassLoader`](https://github.com/wala/WALA/blob/81b08c9df085ca8a4dcc848188079c8d7eea2a1a/core/src/main/java/com/ibm/wala/ipa/summaries/BypassSyntheticClassLoader.java#L57) only after [`buildClassHierarchy()`](https://github.com/wala/WALA/blob/81b08c9df085ca8a4dcc848188079c8d7eea2a1a/core/src/main/java/com/ibm/wala/client/AbstractAnalysisEngine.java#L278), so at hierarchy-build time the supertype is unresolved and the subclass falls back to the language root.

## Change

A generic way to register class shells into a scope loader before the CHA is built:

- [`XMLMethodSummaryReader`](https://github.com/wala/WALA/blob/81b08c9df085ca8a4dcc848188079c8d7eea2a1a/core/src/main/java/com/ibm/wala/ipa/summaries/XMLMethodSummaryReader.java#L62) parses `<class super="...">` and exposes `getClassSuperclasses()`. Declaring a `super` opts a `<class>` in to being materialized as a standalone shell positioned at that superclass; a `<class>` without a `super` stays method-summary-only (unchanged behavior). Restricting materialization to classes that declare a `super` keeps a summary that declares many non-subclassable `<class>` entries from adding them all to the hierarchy.
- [`SummaryClassShellLoader`](https://github.com/ponder-lab/WALA/blob/5aac30e8561990fec3e131844ef40a21cd5f95be/core/src/main/java/com/ibm/wala/ipa/summaries/SummaryClassShellLoader.java#L32) is a capability for loaders that can introduce shells. The [`BypassSyntheticClassLoader`](https://github.com/wala/WALA/blob/81b08c9df085ca8a4dcc848188079c8d7eea2a1a/core/src/main/java/com/ibm/wala/ipa/summaries/BypassSyntheticClassLoader.java#L57) and the CAst module loaders implement it; an implementation may override the hook to register a shell type of its choosing.
- [`SummaryClassShell`](https://github.com/ponder-lab/WALA/blob/5aac30e8561990fec3e131844ef40a21cd5f95be/core/src/main/java/com/ibm/wala/ipa/summaries/SummaryClassShell.java#L31) is a minimal standalone shell `IClass` (unlike [`BypassSyntheticClass`](https://github.com/wala/WALA/blob/81b08c9df085ca8a4dcc848188079c8d7eea2a1a/core/src/main/java/com/ibm/wala/ipa/summaries/BypassSyntheticClass.java#L30), it wraps no real type), positioned at a declared super resolved live via the CHA.
- [`Util.addSummaryClassShells`](https://github.com/ponder-lab/WALA/blob/5aac30e8561990fec3e131844ef40a21cd5f95be/core/src/main/java/com/ibm/wala/ipa/callgraph/impl/Util.java#L202)`(loader, summary)` materializes the opted-in shells into their declaring loader before the hierarchy is built.

## Scope and Caveats

- This PR fixes the inheritance edge and the generic contract. Making an inherited method body actually execute additionally needs the shell to expose the type's methods; that is left to a follow-up.
- A bytecode source subclass in the Application loader still will not see a shell in the synthetic child loader (parent-only lookup), so the beneficiaries are loaders that resolve superclasses against their own type map (the CAst loaders) and synthetic subclasses. Covering bytecode source subclasses would need a separate `BytecodeClass`/CHA resolution fallback.

### DIFF
--- a/cast/src/main/java/com/ibm/wala/cast/loader/CAstAbstractModuleLoader.java
+++ b/cast/src/main/java/com/ibm/wala/cast/loader/CAstAbstractModuleLoader.java
@@ -34,6 +34,7 @@ import com.ibm.wala.classLoader.SourceFileModule;
 import com.ibm.wala.classLoader.SourceModule;
 import com.ibm.wala.core.util.warnings.Warning;
 import com.ibm.wala.ipa.cha.IClassHierarchy;
+import com.ibm.wala.ipa.summaries.SummaryClassShellLoader;
 import com.ibm.wala.ssa.SSAInstruction;
 import com.ibm.wala.ssa.SymbolTable;
 import com.ibm.wala.types.MethodReference;
@@ -61,7 +62,8 @@ import java.util.Set;
  * abstract class loader that performs CAst and IR generation for relevant entities in a list of
  * {@link Module}s. Subclasses provide the CAst / IR translators appropriate for the language.
  */
-public abstract class CAstAbstractModuleLoader extends CAstAbstractLoader {
+public abstract class CAstAbstractModuleLoader extends CAstAbstractLoader
+    implements SummaryClassShellLoader {
 
   private static final boolean DEBUG = false;
 
@@ -392,6 +394,36 @@ public abstract class CAstAbstractModuleLoader extends CAstAbstractLoader {
         return getDeclaringClass().getClassLoader().getLanguage().getRootType();
       }
     }
+  }
+
+  /**
+   * Register a "shell" {@link IClass} for a summary-modeled type, so that source classes loaded by
+   * this loader can resolve it as a superclass at definition time.
+   *
+   * <p>A CAst class resolves its superclass through this loader's own type map ({@link
+   * #lookupClass(TypeName)}, {@link CoreClass#getSuperclass()}), which is loader-local and does not
+   * consult the {@link IClassHierarchy} or parent loaders. When a framework base type is modeled
+   * only by an XML method summary, the corresponding class is materialized into the {@link
+   * com.ibm.wala.ipa.summaries.BypassSyntheticClassLoader} only after the class hierarchy is built,
+   * which is too late, and in the wrong loader, for a source subclass to pick up the inheritance
+   * edge. Registering a shell here, before the hierarchy is built, closes that gap (see <a
+   * href="https://github.com/wala/WALA/issues/1957">#1957</a>).
+   *
+   * <p>Idempotent: if a class is already registered under {@code name}, it is returned unchanged.
+   *
+   * @param name the type to register a shell for
+   * @param superName the shell's superclass, or {@code null} for this loader's language root type
+   * @return the registered (or pre-existing) class
+   */
+  @Override
+  public IClass defineSummaryClassShell(TypeName name, TypeName superName) {
+    IClass existing = lookupClass(name);
+    if (existing != null) {
+      return existing;
+    }
+    TypeName actualSuperName =
+        superName == null ? getLanguage().getRootType().getName() : superName;
+    return new CoreClass(name, actualSuperName, this, null);
   }
 
   public class CoreClass extends AstDynamicPropertyClass {
--- a/core/src/main/java/com/ibm/wala/ipa/callgraph/impl/Util.java
+++ b/core/src/main/java/com/ibm/wala/ipa/callgraph/impl/Util.java
@@ -11,6 +11,7 @@
 package com.ibm.wala.ipa.callgraph.impl;
 
 import com.ibm.wala.classLoader.IClass;
+import com.ibm.wala.classLoader.IClassLoader;
 import com.ibm.wala.classLoader.IMethod;
 import com.ibm.wala.classLoader.Language;
 import com.ibm.wala.core.util.strings.Atom;
@@ -35,6 +36,7 @@ import com.ibm.wala.ipa.cha.IClassHierarchy;
 import com.ibm.wala.ipa.summaries.BypassClassTargetSelector;
 import com.ibm.wala.ipa.summaries.BypassMethodTargetSelector;
 import com.ibm.wala.ipa.summaries.LambdaMethodTargetSelector;
+import com.ibm.wala.ipa.summaries.SummaryClassShellLoader;
 import com.ibm.wala.ipa.summaries.XMLMethodSummaryReader;
 import com.ibm.wala.types.ClassLoaderReference;
 import com.ibm.wala.types.Descriptor;
@@ -52,6 +54,7 @@ import java.io.InputStream;
 import java.io.UncheckedIOException;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.Map;
 import java.util.Set;
 
 /** Call graph utilities */
@@ -180,6 +183,39 @@ public class Util {
             cha,
             cha.getLoader(cha.getScope().getLoader(Atom.findOrCreateUnicodeAtom("Synthetic"))));
     options.setSelector(cs);
+  }
+
+  /**
+   * Materialize class shells for the summary-modeled classes that opt in by declaring a superclass
+   * (see {@link XMLMethodSummaryReader#getClassSuperclasses()}), registering each into {@code
+   * loader} if it supports {@link SummaryClassShellLoader}. Only the classes whose declaring loader
+   * is {@code loader} are registered.
+   *
+   * <p>Call this before building the {@link IClassHierarchy} (and before translating source classes
+   * that subclass a summary-modeled type), so that those subclasses resolve their base at
+   * definition time rather than falling back to the language root. See <a
+   * href="https://github.com/wala/WALA/issues/1957">#1957</a>. Loaders that do not implement {@link
+   * SummaryClassShellLoader} are silently skipped.
+   *
+   * @throws IllegalArgumentException if {@code loader} or {@code summary} is null
+   */
+  public static void addSummaryClassShells(IClassLoader loader, XMLMethodSummaryReader summary) {
+    if (loader == null) {
+      throw new IllegalArgumentException("loader is null");
+    }
+    if (summary == null) {
+      throw new IllegalArgumentException("summary is null");
+    }
+    if (!(loader instanceof SummaryClassShellLoader registrar)) {
+      return;
+    }
+    for (Map.Entry<TypeReference, TypeReference> entry :
+        summary.getClassSuperclasses().entrySet()) {
+      TypeReference type = entry.getKey();
+      if (type.getClassLoader().equals(loader.getReference())) {
+        registrar.defineSummaryClassShell(type.getName(), entry.getValue().getName());
+      }
+    }
   }
 
   /**
--- a/core/src/main/java/com/ibm/wala/ipa/callgraph/impl/Util.java
+++ b/core/src/main/java/com/ibm/wala/ipa/callgraph/impl/Util.java
@@ -11,7 +11,6 @@
 package com.ibm.wala.ipa.callgraph.impl;
 
 import com.ibm.wala.classLoader.IClass;
-import com.ibm.wala.classLoader.IClassLoader;
 import com.ibm.wala.classLoader.IMethod;
 import com.ibm.wala.classLoader.Language;
 import com.ibm.wala.core.util.strings.Atom;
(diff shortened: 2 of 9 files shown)


core/src/main/java/com/ibm/wala/ipa/summaries/SummaryClassShell.java
```diff
@@ -0,0 +1,142 @@
...
+      IClass superClass = cha.lookupClass(superReference);
+      if (superClass != null) {
+        return superClass;
+      }
+    }
+    return cha.getRootClass();
+  }
+
+  @Override
+  public Collection<IClass> getDirectInterfaces() {
```
msridhar, line 59: Generally speaking, is it valid to invoke these "defaulted" instance methods on a `SummaryClassShell`?  If not, maybe we should have these methods throw an exception instead of returning a default value, to fail fast?

cast/src/main/java/com/ibm/wala/cast/loader/CAstAbstractModuleLoader.java
```diff
@@ -394,6 +396,36 @@ public TypeReference getParameterType(int i) {
...
+   */
+  @Override
+  public IClass defineSummaryClassShell(TypeName name, TypeName superName) {
+    IClass existing = lookupClass(name);
+    if (existing != null) {
+      return existing;
+    }
+    TypeName actualSuperName =
+        superName == null ? getLanguage().getRootType().getName() : superName;
+    return new CoreClass(name, actualSuperName, this, null);
```
msridhar, line 426: Why does this use `CoreClass` and not `SummaryClassShell`?

core/src/main/java/com/ibm/wala/ipa/callgraph/impl/Util.java
```diff
@@ -182,6 +185,39 @@ public static void addBypassLogic(
...
+  public static void addSummaryClassShells(IClassLoader loader, XMLMethodSummaryReader summary) {
+    if (loader == null) {
+      throw new IllegalArgumentException("loader is null");
+    }
+    if (summary == null) {
+      throw new IllegalArgumentException("summary is null");
+    }
+    if (!(loader instanceof SummaryClassShellLoader registrar)) {
+      return;
+    }
```
msridhar: Why not just make the declared type of the parameter `SummaryClassShellLoader`?

core/src/main/java/com/ibm/wala/ipa/summaries/XMLMethodSummaryReader.java
```diff
@@ -72,6 +72,25 @@ public class XMLMethodSummaryReader implements BytecodeConstants {
   /** Set of TypeReferences that are marked as "allocatable" */
   private final HashSet<TypeReference> allocatable = HashSetFactory.make();
 
+  /**
+   * Set of TypeReferences for classes declared via a {@code <class>} element. These are the classes
+   * the summary <em>models</em> (as opposed to the classes whose methods it merely overrides).
+   */
+  private final HashSet<TypeReference> classes = HashSetFactory.make();
```
msridhar: Using `Set` as the declared type is better I think

core/src/main/java/com/ibm/wala/ipa/summaries/XMLMethodSummaryReader.java
```diff
@@ -72,6 +72,25 @@ public class XMLMethodSummaryReader implements BytecodeConstants {
...
+   * Only classes that declare a {@code super} appear here: declaring a superclass opts the class in
+   * to being materialized as a standalone class shell, positioned in the hierarchy at the given
+   * superclass. A front-end (or {@link com.ibm.wala.ipa.callgraph.impl.Util#addSummaryClassShells})
+   * may register these shells into the corresponding scope loader before the {@link
+   * com.ibm.wala.ipa.cha.IClassHierarchy} is built, so that source classes subclassing a
+   * summary-modeled type can resolve their base at definition time (see <a
+   * href="https://github.com/wala/WALA/issues/1957">#1957</a>). Classes declared without a {@code
+   * super} stay method-summary-only, exactly as before.
+   */
+  private final HashMap<TypeReference, TypeReference> classSuperclasses = HashMapFactory.make();
```
msridhar: Use `Map` as the declared type

core/src/test/java/com/ibm/wala/core/tests/cha/SummaryClassSuperclassOrderingTest.java
```diff
@@ -0,0 +1,239 @@
...
+      return cha.lookupClass(subRef);
+    }
+  }
+
+  /**
+   * A minimal {@link SyntheticClass} that resolves its superclass through its class loader and
+   * caches the first resolution, mirroring the contract of {@link
+   * com.ibm.wala.classLoader.BytecodeClass#getSuperclass()}.
+   */
+  private static final class SummaryClass extends SyntheticClass {
```
msridhar: Why do we need this?  Can't we just use `SummaryClassShell` directly?

msridhar (commented):
Thanks for this!  Some comments below

core/src/main/java/com/ibm/wala/ipa/summaries/SummaryClassShell.java
khatchad: I think these are valid to invoke. The CHA and downstream analyses query a class's interfaces, fields, and methods for every class they process, and a shell is a real participant in the hierarchy, so for a member-less shell the empty collections and `null`s are the correct answers rather than placeholders. The exception is `getModifiers()`, which has no sensible value for a shell and contradicted `isPublic()` returning `true`, so I made it throw `UnsupportedOperationException` to fail fast.

cast/src/main/java/com/ibm/wala/cast/loader/CAstAbstractModuleLoader.java
khatchad: The shell in a CAst loader should be a real CAst class. `CoreClass` is an `AstClass` that registers itself in the loader's `types` map, so loader-local superclass resolution ([`CAstAbstractLoader.lookupClass`](https://github.com/wala/WALA/blob/81b08c9df085ca8a4dcc848188079c8d7eea2a1a/cast/src/main/java/com/ibm/wala/cast/loader/CAstAbstractLoader.java#L96), [`CoreClass.getSuperclass`](https://github.com/wala/WALA/blob/81b08c9df085ca8a4dcc848188079c8d7eea2a1a/cast/src/main/java/com/ibm/wala/cast/loader/CAstAbstractModuleLoader.java#L432)) finds it, and a front-end can extend it (for example `PythonClass extends CoreClass`) instead of getting a bare placeholder. [`SummaryClassShell`](https://github.com/ponder-lab/WALA/blob/5aac30e8561990fec3e131844ef40a21cd5f95be/core/src/main/java/com/ibm/wala/ipa/summaries/SummaryClassShell.java#L31) is the minimal stand-in for the [`BypassSyntheticClassLoader`](https://github.com/wala/WALA/blob/81b08c9df085ca8a4dcc848188079c8d7eea2a1a/core/src/main/java/com/ibm/wala/ipa/summaries/BypassSyntheticClassLoader.java#L57), which has no real-class machinery, so each capability implementation produces a loader-appropriate shell.

core/src/main/java/com/ibm/wala/ipa/callgraph/impl/Util.java
khatchad: Will do. I made `SummaryClassShellLoader` extend `IClassLoader` (both implementations already are class loaders) so the parameter can be typed as `SummaryClassShellLoader` while still using `getReference()` for the declaring-loader filter, and dropped the `instanceof` guard.

core/src/main/java/com/ibm/wala/ipa/summaries/XMLMethodSummaryReader.java
khatchad: Done, changed the field type to `Set<TypeReference>`.

core/src/main/java/com/ibm/wala/ipa/summaries/XMLMethodSummaryReader.java
khatchad: Done, changed to `Map<TypeReference, TypeReference>`.

core/src/test/java/com/ibm/wala/core/tests/cha/SummaryClassSuperclassOrderingTest.java
khatchad: Good point. `SummaryClassShell` resolves its superclass live through the CHA, so this ordering scenario self-heals with it and would not reproduce the durable fallback. I refocused the test on the subclass-graph snapshot, which stays stale even for a live-resolving class, and used `SummaryClassShell` directly so the helper goes away.

khatchad:
For motivation: this comes from the WALA-based Ariadne analysis (wala/ML), where user code subclasses a framework base modeled only by an XML method summary, for example `class GCN(tf.keras.layers.Layer)`. Today that subclass loses its base at CHA-build time and falls back to `object`, which blocks the inherited `__call__` from running and breaks downstream tensor analysis (wala/ML#118, wala/ML#106, wala/ML#595). This PR is the generic WALA piece that lets the base resolve, while the front-end-specific shell type and method wiring stay in the front-end.

msridhar (approved):
Thanks!

msridhar:
@khatchad this is still marked as a draft.  Please mark it as ready when appropriate and then we can land it.

khatchad:
> @khatchad this is still marked as a draft. Please mark it as ready when appropriate and then we can land it.

Sorry! Done. Thanks again!!

PR #1959, opened 2026-06-23
Expose a summary class shell's own methods (#1957 follow-up)

khatchad:
Depends on #1958 and is stacked on it. **Do not merge before #1958.** Until #1958 merges, the diff here also includes its commits; once #1958 lands, this reduces to the single follow-up commit. Toward #1957.

## What

#1958 registers a class shell before the CHA is built so a source subclass resolves its summary-modeled base. This follow-up lets that shell also carry the base's own methods, so a walk of the shell (for example, resolving an inherited method on the subclass) finds them. The method bodies are still bound late by the usual bypass mechanism; only the declarations need to exist on the shell.

## Change

- [`SummaryClassShell`](https://github.com/ponder-lab/WALA/blob/3cd29c2918b965892a9148a73bde7ab03e1c7b5e/core/src/main/java/com/ibm/wala/ipa/summaries/SummaryClassShell.java#L97) exposes the method summaries declared by its `<class>` (its `<method>` children) as its own declared methods, materialized lazily as `SummarizedMethod`s.
- [`SummaryClassShellLoader`](https://github.com/ponder-lab/WALA/blob/3cd29c2918b965892a9148a73bde7ab03e1c7b5e/core/src/main/java/com/ibm/wala/ipa/summaries/SummaryClassShellLoader.java#L63) gains a `defineSummaryClassShell(name, super, methods)` overload. It is a `default` that ignores `methods` and delegates to the no-methods version; the `BypassSyntheticClassLoader` overrides it. The no-methods 2-arg overload remains for a pure base shell.
- [`Util.addSummaryClassShells`](https://github.com/ponder-lab/WALA/blob/3cd29c2918b965892a9148a73bde7ab03e1c7b5e/core/src/main/java/com/ibm/wala/ipa/callgraph/impl/Util.java#L218) groups the summary's methods by declaring class and hands each shell its own.

A loader whose shells cannot carry methods (the CAst module loaders' default) keeps the prior behavior; such a loader may override the hook to build a shell type that does.

### DIFF
--- a/cast/src/main/java/com/ibm/wala/cast/loader/CAstAbstractModuleLoader.java
+++ b/cast/src/main/java/com/ibm/wala/cast/loader/CAstAbstractModuleLoader.java
@@ -34,6 +34,7 @@ import com.ibm.wala.classLoader.SourceFileModule;
 import com.ibm.wala.classLoader.SourceModule;
 import com.ibm.wala.core.util.warnings.Warning;
 import com.ibm.wala.ipa.cha.IClassHierarchy;
+import com.ibm.wala.ipa.summaries.SummaryClassShellLoader;
 import com.ibm.wala.ssa.SSAInstruction;
 import com.ibm.wala.ssa.SymbolTable;
 import com.ibm.wala.types.MethodReference;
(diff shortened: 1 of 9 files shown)


khatchad:
Reverting: keeping this as a stacked PR on the fork (ponder-lab/WALA#9) for a clean follow-up-only diff. Will retarget upstream once #1958 merges.

PR #1961, opened 2026-06-24
Enforce or explicitly disable many Error Prone checks

liblit:
Remove the blanket `disableAllWarnings = true` that suppressed all warning-level Error Prone checks.  Instead,

- promote ~200 individual checks from warning to error, since none of our current code violates them;

- explicitly disable ~60 checks that would still detect violations in our current code; and

- keep `MissingSummary` and `ReferenceEquality` separately disabled with documented rationale for not addressing them in the near term.

See also discussion #1951.


msridhar (approved):
Looks great!  My only concern is keeping these lists up to date as we upgrade Error Prone versions.  I guess for any new checks we should see warnings in the build log to know they exist if we have any violations?  In any case, we can probably study release notes and figure that out as we go.

liblit:
> My only concern is keeping these lists up to date as we upgrade Error Prone versions.

That consideration informed my decision to remove the catch-alll `disableAllWarnings = true` setting. With that removed, if we upgrade Error Prone and it adds new default-on errors or warnings that we violate, then the build will fail. (Even on warnings? [Yes, even on warnings.](https://github.com/wala/WALA/blob/946edd6db1c440f50a5ad2a9432d24ec76ef8a41/build-logic/src/main/kotlin/com/ibm/wala/gradle/java.gradle.kts#L181)) Whomever is doing the upgrade will then have to decide what to fix, suppress, or disable. That seems reasonable.

PR #1962, opened 2026-06-24
Expose a summary class shell's own methods

khatchad:
Follow-up to #1958. Toward #1957.

## What

#1958 registers a class shell before the CHA is built so a source subclass resolves its summary-modeled base. This follow-up lets that shell also carry the base's own methods, so a walk of the shell (for example, resolving an inherited method on the subclass) finds them. The method bodies are still bound late by the usual bypass mechanism; only the declarations need to exist on the shell.

## Change

- [`SummaryClassShell`](https://github.com/ponder-lab/WALA/blob/92f62ace6e185a4e6f6f2976efe831daaa52ba5f/core/src/main/java/com/ibm/wala/ipa/summaries/SummaryClassShell.java#L97) exposes the method summaries declared by its `<class>` (its `<method>` children) as its own declared methods, materialized lazily as `SummarizedMethod`s.
- [`SummaryClassShellLoader`](https://github.com/ponder-lab/WALA/blob/92f62ace6e185a4e6f6f2976efe831daaa52ba5f/core/src/main/java/com/ibm/wala/ipa/summaries/SummaryClassShellLoader.java#L63) gains a `defineSummaryClassShell(name, super, methods)` overload. It is a `default` that ignores `methods` and delegates to the no-methods version; the `BypassSyntheticClassLoader` overrides it. The no-methods 2-arg overload remains for a pure base shell.
- [`Util.addSummaryClassShells`](https://github.com/ponder-lab/WALA/blob/92f62ace6e185a4e6f6f2976efe831daaa52ba5f/core/src/main/java/com/ibm/wala/ipa/callgraph/impl/Util.java#L213) groups the summary's methods by declaring class and hands each shell its own.

A loader whose shells cannot carry methods (the CAst module loaders' default) keeps the prior behavior; such a loader may override the hook to build a shell type that does.


khatchad:
@msridhar this is the stacked follow-up to #1958 (now merged). It lets a summary class shell expose its `<class>`'s own methods, so a walk of the shell (for example, resolving an inherited method on a subclass) finds them; the bodies are still bound late by the usual bypass mechanism. Single-commit diff. Whenever you have a moment, a review would be appreciated.

msridhar (approved):
LGTM!

khatchad:
@msridhar the merge queue removed this for "no response for status checks", i.e. the queue's CI run stalled and timed out rather than failing. It is approved and `mergeStateStatus` is `CLEAN`. Could you re-add it to the queue when you get a chance? (Same thing happened to #1958's first queue attempt and it went through cleanly on the retry.)

PR #1966, opened 2026-06-25
Fix `Language`/`JavaLanguage` initialization deadlock

liblit:
Use the initialization-on-demand holder idiom for lazy, thread-safe singleton initialization of `JavaLanguage`.  The deadlock-tempting `Language.JAVA` field is deprecated in favor of a new deadlock-free `JavaLanguage.get()` method.

**Reviewer guidance:** most of these changes are systematic replacements of `Language.JAVA` with `JavaLanguage.get()` and the corresponding `import` updates. The _interesting_ parts of this change are in `Java.gradle.kts`, `JavaLanguage.java`, and `Language.java`.


msridhar (approved):
Nice!

PR #1972, opened 2026-06-29
Keep summary parameter names that begin with "arg"

khatchad:
`XMLMethodSummaryReader` seeds a summary method's value-name table with synthetic positional symbols (`arg0`, `arg1`, ...) and then drops them so they are not mistaken for real local names. The filter used is `startsWith("arg")`, which is over-broad: it also discards legitimate declared parameter names that merely begin with `arg`, such as `args` and `argv`.

A parameter whose name is dropped has no value name. Keyword arguments bind to a callee's parameters by matching the keyword against the callee's value names (see how Python front-end keyword binding consults `getLocalNames`), so a parameter named `args` can never receive a keyword argument `args=...`; the binding silently no-ops.

This narrows the filter to the exact synthetic pattern `arg\d+`, so `arg0`/`arg1` stay filtered while declared names like `args` and `argv` are retained.

Surfaced in Ariadne (https://github.com/wala/ML), where [`tf.distribute.Strategy.run(fn, args=(...))`](https://www.tensorflow.org/api_docs/python/tf/distribute/Strategy#run) is modeled with an `args` tuple parameter whose keyword form could not bind until the name was restored.


khatchad:
@msridhar would you have a chance to review this? It narrows the summary reader's `startsWith("arg")` parameter-name filter to the synthetic `arg\d+` symbols, so legitimately-named parameters like `args` and `argv` keep their value names (and can therefore receive keyword arguments). Full rationale and a regression test are in the description.

msridhar (approved):
Thanks for the fix!

PR #1982, opened 2026-07-01
Update release notes file for 1.8.0

msridhar:
Add a version number to the file; we will plan to add to this file for each future release.  Update indentation accordingly.  Call out some key functionality changes and pull requests.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -4,24 +4,36 @@
 
 ### Functionality changes
 
-WALA now requires a Java 17+.  With this new requirement, WALA has internally adopted various new
-Java language features, including `switch` expressions and records.  Beyond this change, there
-have been various cleanups and bug fixes.
+WALA now requires a Java 17+. With this new requirement, WALA has
+internally adopted various new Java language features, including `switch`
+expressions and records. Beyond this change, there have been various cleanups
+and bug fixes.
 
-Key pull requests
------------------
-* Require Java 17+ across all of WALA by @liblit in https://github.com/wala/WALA/pull/1904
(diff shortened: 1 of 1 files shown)


RELEASE-NOTES.md
```diff
@@ -1,8 +1,43 @@
...
-## Dependency changes
-
-### `:core` and `:util` now declare Guava as `implementation`-scope dependencies
+## Version 1.8.0
+
+### Functionality changes
+
+WALA now requires a Java 17+. With this new requirement, WALA has
+internally adopted various new Java language features, including `switch`
+expressions and records. Beyond this change, there have been various cleanups
```
liblit: I've been assuming that internal changes are of no interest to someone reading the release notes. I don't _mind_ covering internal changes, but my assumption was that WALA users would treat WALA itself as a black box.

liblit (dismissed):
I've been assuming that internal changes are of no interest to someone reading the release notes. If that's true, then the fact that we require Java 17 _is_ relevant, but the fact that we now use `switch` expressions is not. Likewise, many of the listed pull requests would not be of interest if treating WALA as a black box.

I'm approving this PR in case @msridhar thinks it makes sense to cover internal changes. I don't _mind_ doing that, it's just not what I was assuming. I turned off "Merge when ready" for this PR in case @msridhar wants to reconsider. But if @msridhar prefers to have this info here, that's 100% OK with me too!

RELEASE-NOTES.md
msridhar: I've now removed the sentence on the internal change

liblit (approved):

PR #1986, opened 2026-07-01
Remove an unused constructor

liblit:


liblit:
**Reviewer question:** do we think that someone _might_ use this in the future, and that the right way to chain to the other constructor will be non-obvious? If so, then we could keep it and either comment it out or annotate it with `@SuppressWarnings("unused”)`.

msridhar:
> **Reviewer question:** do we think that someone _might_ use this in the future, and that the right way to chain to the other constructor will be non-obvious? If so, then we could keep it and either comment it out or annotate it with `@SuppressWarnings("unused”)`.

I think it's fine to delete this; I don't expect anyone to be looking for this method anytime soon

msridhar (approved):

PR #1993, opened 2026-07-08
Bound `NodeImpl.toString()` against cyclic contexts

### LINKED ISSUE #1992: Context.toString() can recurse without bound on cyclic scope-mapping/receiver contexts
`Context.toString()` and the instance-key `toString()`s it transitively calls form a reference cycle with no cycle detection, so on cyclic context structures the render grows until the JVM runs out of memory.

The chain:

- `BasicCallGraph$NodeImpl.toString()` renders `context.toString()`.
- `CallerContext.toString()` renders its `caller` (a `CGNode`, back into `NodeImpl.toString()`).
- `CallerSiteContextPair.toString()` renders `super.toString()` plus its `baseContext`.
- `ReceiverInstanceContext.toString()` renders its `ik`, which may be a `ScopeMappingInstanceKey`.
- `ScopeMappingInstanceKeys$ScopeMappingInstanceKey.toString()` renders its `creator` (a `CGNode`).

When a node's context transitively references that same node, which arises naturally from mutually-recursive closures (e.g. a closure whose `ScopeMappingInstanceKey.creator` is the node whose receiver context is that key), the render never terminates. None of these `toString()` methods carry a visited-set or depth limit.

Observed with the Ariadne Python front-end analyzing a closure-heavy program: a single `CallGraph` IR-dump line reached 19 MB before OOM, and any `LOGGER.fine("..." + pointsToVariable)` at a fine level exhausted a 16 GB heap.

Suggested fix: add cycle detection (an identity visited-set) or a depth bound so a diagnostic render can never loop or OOM. Note a design constraint: `toString()` takes no parameters, so a visited-set cannot be threaded through the call chain, and the cycle spans four classes across two modules (`core` and `cast`), so no single method can own the set. A per-method `Set` is therefore a dead end. The viable routes are a `ThreadLocal` visited-set or depth counter installed and removed around the render, or breaking the cycle at the `CGNode` boundary by having the context renderers print embedded nodes shallowly (method plus a stable id, without expanding their context).

khatchad:
## Problem

Fixes #1992. `BasicCallGraph$NodeImpl.toString()` and the instance-key `toString()`s it transitively calls form a reference cycle with no cycle detection. On cyclic context structures, which arise naturally from mutually-recursive closures (e.g. a closure whose `ScopeMappingInstanceKey.creator` is the node whose receiver context is that key), the render never terminates and the JVM runs out of memory. Observed with the Ariadne Python front-end: a single `CallGraph` IR-dump line reached 19 MB before OOM.

## Fix

Guard the single re-entry point, `NodeImpl.toString()`, with a per-thread identity set of the nodes currently rendering. A re-entrant call on a node already being rendered returns a shallow `Context: ...` instead of recursing. This defends every cyclic render that passes through a `CGNode`, not just the two edges enumerated in the issue, and leaves acyclic output byte-for-byte unchanged.

## Test

`NodeImplToStringTest` builds a real `NodeImpl` whose caller context renders back to that same node and asserts `toString()` terminates with a bounded string. It fails with a `StackOverflowError` without the guard and passes with it.

## Alternative Considered

A narrower fix that shallow-renders the two `CGNode` boundaries (`CallerContext`, `ScopeMappingInstanceKey`) is smaller but only closes the two documented edges, changes output for all callers even in the acyclic case, and spans two modules. The `ThreadLocal` guard was chosen for generality and for preserving existing output.


core/src/test/java/com/ibm/wala/ipa/callgraph/impl/NodeImplToStringTest.java
```diff
@@ -0,0 +1,155 @@
...
+   */
+  private static IMethod fakeMethod() {
+    return (IMethod)
+        Proxy.newProxyInstance(
+            NodeImplToStringTest.class.getClassLoader(),
+            new Class<?>[] {IMethod.class},
+            (proxy, m, args) -> {
+              if (m.getName().equals("toString")) return "fakeMethod()";
+              return defaultValue(m.getReturnType());
+            });
```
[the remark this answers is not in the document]
khatchad: Done in c70766d8e. The IMethod proxy now handles equals (identity) and hashCode (System.identityHashCode) explicitly rather than falling through to false/0.

core/src/test/java/com/ibm/wala/ipa/callgraph/impl/NodeImplToStringTest.java
```diff
@@ -0,0 +1,155 @@
...
+        Proxy.newProxyInstance(
+            NodeImplToStringTest.class.getClassLoader(),
+            new Class<?>[] {CGNode.class},
+            (proxy, m, args) ->
+                switch (m.getName()) {
+                  case "toString" -> holder[0].toString();
+                  case "getMethod" -> fakeMethod();
+                  case "getGraphNodeId" -> 0;
+                  case "equals" -> proxy == args[0];
+                  case "hashCode" -> 0;
```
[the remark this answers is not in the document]
khatchad: Done in c70766d8e. The caller proxy now returns System.identityHashCode(proxy) for hashCode. On the equals point: the args array passed to an invocation handler for equals is always non-null with exactly one element (the compared object, which may itself be null), so proxy == args[0] cannot NPE.

core/src/test/java/com/ibm/wala/ipa/callgraph/impl/NodeImplToStringTest.java
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  /**
+   * Returns a type-appropriate default for an unstubbed proxy method: {@code false}, 0, or null.
+   */
+  private static Object defaultValue(Class<?> returnType) {
+    if (returnType == boolean.class) return false;
+    if (returnType == int.class) return 0;
+    return null;
+  }
```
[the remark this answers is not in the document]
khatchad: Done in c70766d8e. defaultValue now returns the zero value for every primitive return type (boolean, char, byte, short, int, long, float, double) and null only for reference and void returns.

core/src/main/java/com/ibm/wala/ipa/callgraph/impl/BasicCallGraph.java
```diff
@@ -187,12 +188,30 @@ public IMethod getMethod() {
     @Override
     public abstract int hashCode();
 
+    /**
+     * Tracks the nodes whose {@link #toString()} is in progress on the current thread. Cyclic
+     * context structures (e.g. a closure whose context transitively references this same node)
+     * would otherwise cause {@code toString()} to recurse without bound and exhaust the heap. See
+     * <a href="https://github.com/wala/WALA/issues/1992">issue #1992</a>.
+     */
+    private static final ThreadLocal<Set<CGNode>> RENDERING_IN_PROGRESS =
```
liblit, line 197: Thank you for thinking ahead about thread-safety.

core/src/main/java/com/ibm/wala/ipa/callgraph/impl/BasicCallGraph.java
```diff
@@ -187,12 +188,30 @@ public IMethod getMethod() {
...
+        ThreadLocal.withInitial(() -> Collections.newSetFromMap(new IdentityHashMap<>()));
```
liblit: This code could be made more concise using Guava, which is already a dependency:

```java
ThreadLocal.withInitial(Sets::newIdentityHashSet);
```

core/src/main/java/com/ibm/wala/ipa/callgraph/impl/BasicCallGraph.java
```diff
@@ -187,12 +188,30 @@ public IMethod getMethod() {
...
+
     /**
      * @see java.lang.Object#toString()
      */
     @Override
     public String toString() {
-      return "Node: " + method.toString() + " Context: " + context.toString();
+      Set<CGNode> inProgress = RENDERING_IN_PROGRESS.get();
+      if (!inProgress.add(this)) {
+        // Re-entered while already rendering this node: break the cycle with a shallow render.
```
liblit, line 207: While this strategy avoids cycles, it does not avoid potential exponential blowup in the result `String` length if there is a lot of non-cyclic sharing. Is that a concern too? If you're not sure whether that's a concern or not, then I favor sticking with the solution you have here: let's not over-complicate without good reason.

core/src/main/java/com/ibm/wala/ipa/callgraph/impl/BasicCallGraph.java
```diff
@@ -187,12 +188,30 @@ public IMethod getMethod() {
...
+        return "Node: " + method.toString() + " Context: ...";
```
liblit: String conversation is implicit in string concatenation. Therefore, simplify: `method.toString()` → `method`.

core/src/main/java/com/ibm/wala/ipa/callgraph/impl/BasicCallGraph.java
```diff
@@ -187,12 +188,30 @@ public IMethod getMethod() {
...
+      }
+      try {
+        return "Node: " + method.toString() + " Context: " + context.toString();
```
liblit: String conversation is implicit in string concatenation. Therefore, simplify: `method.toString()` → `method` and `context.toString()` → `context`.

core/src/test/java/com/ibm/wala/ipa/callgraph/impl/NodeImplToStringTest.java
```diff
@@ -0,0 +1,166 @@
...
+import java.lang.reflect.Proxy;
+import java.time.Duration;
+import java.util.Iterator;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Regression test for <a href="https://github.com/wala/WALA/issues/1992">issue #1992</a>: {@link
+ * BasicCallGraph.NodeImpl#toString()} must terminate on a cyclic context rather than recurse
+ * without bound and exhaust the heap.
+ */
```
liblit, line 35: Thank you for taking the time to write tests for your change.

core/src/test/java/com/ibm/wala/ipa/callgraph/impl/NodeImplToStringTest.java
```diff
@@ -0,0 +1,166 @@
...
+    if (returnType == boolean.class) return false;
+    if (returnType == char.class) return '\0';
+    if (returnType == byte.class) return (byte) 0;
+    if (returnType == short.class) return (short) 0;
+    if (returnType == int.class) return 0;
+    if (returnType == long.class) return 0L;
+    if (returnType == float.class) return 0f;
+    if (returnType == double.class) return 0d;
+    return null;
+  }
```
liblit: Can we replace this method with [`Defaults.defaultValue`](https://guava.dev/releases/snapshot/api/docs/com/google/common/base/Defaults.html#defaultValue(java.lang.Class)) from Guava?

liblit (changes requested):

core/src/main/java/com/ibm/wala/ipa/callgraph/impl/BasicCallGraph.java
khatchad: Done in 9b1541127. Much nicer, thanks.

core/src/main/java/com/ibm/wala/ipa/callgraph/impl/BasicCallGraph.java
khatchad: Done in 9b1541127.

core/src/main/java/com/ibm/wala/ipa/callgraph/impl/BasicCallGraph.java
khatchad: Done in 9b1541127.

core/src/test/java/com/ibm/wala/ipa/callgraph/impl/NodeImplToStringTest.java
khatchad: Done in 9b1541127. The helper is gone; the proxies now call Guava's Defaults.defaultValue directly, which has the same semantics (zero value for primitives, null for reference and void returns).

core/src/main/java/com/ibm/wala/ipa/callgraph/impl/BasicCallGraph.java
khatchad: I considered that. Exponential blowup would need a node whose context fans out to multiple distinct nodes that each share substructure, e.g. a DAG of delegating contexts. The contexts WALA builds in practice (CallerContext, CallStringContext, receiver contexts) form chains or small trees per node, and issue #1992 was specifically about the cyclic case, so I have not seen the DAG-sharing pattern arise. Agreed on not over-complicating; if blowup ever shows up in practice, the same guard set could be extended to a depth or length budget.

liblit (approved):

PR #1994, opened 2026-07-14
Remove a redundant condition clause

liblit:
If `attr` starts with `"on"`, then it must be at least two characters long.  No need to check the latter if we're also going to check the former.

### DIFF
--- a/cast/js/src/main/java/com/ibm/wala/cast/js/html/DefaultSourceExtractor.java
+++ b/cast/js/src/main/java/com/ibm/wala/cast/js/html/DefaultSourceExtractor.java
@@ -167,7 +167,7 @@ public class DefaultSourceExtractor extends DomLessSourceExtractor {
         ITag tag, String attr, String value, String varName, String varName2) {
       // There should probably be more checking to see what the attributes are since we allow things
       // like: ; to be used as attributes now.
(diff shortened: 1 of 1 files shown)


cast/js/src/main/java/com/ibm/wala/cast/js/html/DefaultSourceExtractor.java
```diff
@@ -167,7 +167,7 @@ protected void writeEventAttribute(
         ITag tag, String attr, String value, String varName, String varName2) {
       // There should probably be more checking to see what the attributes are since we allow things
       // like: ; to be used as attributes now.
-      if (attr.length() >= 2 && attr.startsWith("on")) {
+      if (attr.startsWith("on")) {
```
msridhar, line 170: This is the commit that introduced the length check:

https://github.com/wala/WALA/commit/11c6619c7f1791d00e8ba2de464f81b61aa680f9

It seems like it was purposely added.  I wonder if the intention was to not match `"on"` itself as a valid attribute, in which case the check should be `attr.length() > 2`?  I can't really recall.  In any case, if all the tests are passing, the test is clearly redundant, so we can get rid of it.

msridhar (approved):

PR #1998, opened 2026-07-15
Allow `Assertions.UNREACHABLE(...)` in expressions

liblit:
Previously, the various overloads of `Assertions.UNREACHABLE(...)` returned `void`.  Therefore, calls to these methods could be used as statements but not expressions.  That restriction led to many code fragments like the following:

```java
Assertions.UNREACHABLE();
return null;
```

An unfortunate side effect of that style was to make many methods appear to return nullable values even though the `return null` statement could never be reached.

Now the `Asserions.UNREACHABLE(...)` methods statically claim to return a value of any arbitrary type.  In reality, they will never return at all; each such method always throws an exception.  However, by _pretending_ to return an arbitrarily typed value, we can now replace the above code pattern with:

```java
return Assertions.UNREACHABLE();
```

This change should help us recognize that some methods can never return `null`, even though they contain unreachable code.

This change was inspired by several standard Kotlin features.  The `TODO()`, `TODO(String)`, and `error(Any)` functions all return type `Nothing`, which is a subtype of every other type.  Kotlin's `throw ...` construct is also an expression (not a statement) with static type `Nothing`.

Also add JetBrains `@Contract` annotations to `Asserions.UNREACHABLE(...)` methods declaring that they always "fail" (i.e., throw exceptions) regardless of the arguments passed to them. Some static analysis tools may benefit from this information.

### DIFF
--- a/.markdownlint-cli2.yaml
+++ b/.markdownlint-cli2.yaml
@@ -2,3 +2,6 @@ $schema: https://raw.githubusercontent.com/DavidAnson/markdownlint-cli2/main/sch
 noBanner: true
 noProgress: true
 showFound: false
+config:
+  no-duplicate-heading:
+    siblings_only: true
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,36 @@
 # WALA Release Notes
 
+## Version 1.9.0
+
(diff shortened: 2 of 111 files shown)


msridhar (approved):
Nice!

PR #2000, opened 2026-07-19
Update to NullAway 0.13.8 and address new warning

msridhar:
`MultiMap.removeAll` now returns the empty set when a key is absent, a safer and more consistent behavior.

### DIFF
--- a/util/src/main/java/com/ibm/wala/util/collections/AbstractMultiMap.java
+++ b/util/src/main/java/com/ibm/wala/util/collections/AbstractMultiMap.java
@@ -110,7 +110,7 @@ abstract class AbstractMultiMap<K, V> implements Serializable, MultiMap<K, V> {
   }
 
   @Override
-  public Set<V> removeAll(K key) {
+  public @Nullable Set<V> removeAll(K key) {
     return map.remove(key);
   }
 
--- a/util/src/main/java/com/ibm/wala/util/collections/AbstractMultiMap.java
+++ b/util/src/main/java/com/ibm/wala/util/collections/AbstractMultiMap.java
@@ -110,8 +110,9 @@ abstract class AbstractMultiMap<K, V> implements Serializable, MultiMap<K, V> {
   }
 
   @Override
-  public @Nullable Set<V> removeAll(K key) {
-    return map.remove(key);
+  public Set<V> removeAll(K key) {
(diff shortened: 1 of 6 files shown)


util/src/main/java/com/ibm/wala/util/collections/MultiMap.java
```diff
@@ -59,7 +60,7 @@ public interface MultiMap<K, V> {
 
   boolean putAll(K key, Collection<? extends V> vals);
 
-  Set<V> removeAll(K key);
+  @Nullable Set<V> removeAll(K key);
```
liblit: Can we document the circumstances under which a this method returns `null`? My guess is that it returns `null` when `key` was not in the map. But one could also consider returning the empty set in that case. Is it _guaranteed_ that removing an absent key must always return `null`, i.e., that this method _never_ returns the empty set?

liblit (changes requested):

util/src/main/java/com/ibm/wala/util/collections/MultiMap.java
msridhar: Good catch!  I think it's better and more consistent to just return an empty set when the key is absent; I've updated the `AbstractMultiMap` implementation to do so.

util/src/main/java/com/ibm/wala/util/collections/AbstractMultiMap.java
```diff
@@ -111,7 +111,8 @@ public boolean remove(K key, V val) {
 
   @Override
   public Set<V> removeAll(K key) {
-    return map.remove(key);
+    Set<V> result = map.remove(key);
+    return result != null ? result : Collections.emptySet();
```
liblit, line 115: Does `MultiMap` appear in any `public` WALA API? If yes, then returning an empty set instead of `null` for missing keys is a behavioral change that should be mentioned in `RELEASE-NOTES.md`.

liblit (changes requested):

util/src/main/java/com/ibm/wala/util/collections/AbstractMultiMap.java
msridhar: Now documented in release notes

liblit (approved):

PR #2006, opened 2026-07-19
Make `DFSFinishTimeIterator` more `null`-strict

liblit:
Each `DFSFinishTimeIterator` is an `ArrayList`.  I'm trying to establish an invariant that the elements of that `ArrayList` are never `null`.  If we can convince ourselves that this is true, then that will help some other improvements I have pending.

This change adds two dynamic `requireNonNull(...)` checks in places where we cannot statically establish non-nullability.  With those two checks added, NullAway is satisfied that the rest of what we are doing here is valid.  But do we believe that each of those dynamic checks must always succeed?  They do always succeed in our tests, but perhaps our tests just don't cover some important cases.

### DIFF
--- a/util/src/main/java/com/ibm/wala/util/graph/traverse/DFS.java
+++ b/util/src/main/java/com/ibm/wala/util/graph/traverse/DFS.java
@@ -48,7 +48,7 @@ public class DFS {
         new SlowDFSFinishTimeIterator<>(G, C.iterator()) {
 
           @Override
-          protected Iterator<T> getConnected(@Nullable T n) {
+          protected Iterator<T> getConnected(T n) {
             return new FilterIterator<>(G.getSuccNodes(n), filter);
           }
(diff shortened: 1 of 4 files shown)


msridhar (commented):
The `requireNonNull` calls look valid to me.  And I'm very much in favor of getting rid of the `@Nullable` annotations that this PR removes.

Could we add a brief comment near each `requireNonNull` call documenting why we think it should never fail?

liblit:
> Could we add a brief comment near each `requireNonNull` call documenting why we think it should never fail?

Good idea; [done](https://github.com/wala/WALA/pull/2006/commits/7cbafd3e782ec0b77a123034da9b653642d1b12b). Thank you for challenging me to justify these invariants more rigorously.

msridhar (approved):
Thanks @liblit looks great!

PR #2010, opened 2026-07-20
Stop extending `ArrayList` in DFS iterator classes

liblit:
Replace exposed `ArrayList` backing store with an internal `ArrayDeque` in `DFSDiscoverTimeIterator` and `DFSFinishTimeIterator`.  These classes now implement only `Iterator<T>`, which better reflects their actual usage: they are consumed as iterators, never as lists.

This is a source-incompatible change for any caller that relied on `ArrayList`/`List` methods (`size()`, `get()`, `add()`, etc.) on instances of these types.  Code that only uses the `Iterator<T>` contract is unaffected.

Also move the `IterableAndIterator` Error Prone check from the suppressed-to-warn list to the suppressed-to-error list, since the refactored code no longer triggers it.

Update `RELEASE-NOTES.md` to document these API changes for some future WALA v1.9.0 release.


msridhar (dismissed):

msridhar:
Nice fix!

msridhar (approved):

PR #2011, opened 2026-07-20
Read `StringBuilder` instead of short-lived `String`

liblit:
Replace `StringReader` and `ByteArrayInputStream` wrappers in `JavaScriptConstructorFunctions` with `CharSequenceReader` and `ReaderInputStream` from `commons-io`.  This change lets us read directly from the `StringBuilder` without creating an intermediate `String`.

Note that `commons-io` is already a dependency here, so this change does not affect dependencies in any way that third-party WALA users would notice.

### DIFF
--- a/cast/js/src/main/java/com/ibm/wala/cast/js/ipa/summaries/JavaScriptConstructorFunctions.java
+++ b/cast/js/src/main/java/com/ibm/wala/cast/js/ipa/summaries/JavaScriptConstructorFunctions.java
@@ -519,7 +519,9 @@ public class JavaScriptConstructorFunctions {
                 @Override
                 @SuppressWarnings("resource")
                 public InputStream getInputStream() {
-                  // No leak here: `ReaderInputStream.close()` closes the underlying `Reader`.
+                  // `ReaderInputStream` takes ownership of the underlying `Reader`.  Therefore, if
+                  // whoever called `getInputStream()` eventually closes the returned `InputStream`,
+                  // then that will also close the `Reader` that `getInputReader()` provided.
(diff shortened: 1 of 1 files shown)


cast/js/src/main/java/com/ibm/wala/cast/js/ipa/summaries/JavaScriptConstructorFunctions.java
```diff
@@ -518,8 +517,10 @@ public boolean isSourceFile() {
                 }
 
                 @Override
+                @SuppressWarnings("resource")
                 public InputStream getInputStream() {
-                  return new ByteArrayInputStream(fun.toString().getBytes(StandardCharsets.UTF_8));
+                  // No leak here: `ReaderInputStream.close()` closes the underlying `Reader`.
```
msridhar: I don't quite follow this comment; where does `ReaderInputStream.close()` get called?

msridhar (commented):
One question

cast/js/src/main/java/com/ibm/wala/cast/js/ipa/summaries/JavaScriptConstructorFunctions.java
liblit: Good point. My "clarifying" comment was confusing in its own right. [I have improved it.](https://github.com/wala/WALA/pull/2011/commits/237348c69f3d4e720d481d339f4bdc9144b2e80c)

msridhar (approved):
Looks good now, thanks!

PR #2015, opened 2026-07-22
Switch IntelliJ JDK name to temurin-25 and add docs

### LINKED ISSUE #2014: Can we update IntelliJ config to not mention Zulu JDK?
See here:

https://github.com/wala/WALA/blob/ca4c34b1367eef8b952bc45400db9758c9dffa12/.idea/misc.xml#L65

I've switched over to Temurin JDKs locally.  Also ideally I'd just use any JDK version as long as it supports language level 17.

@liblit do you know if we can update our committed config to allow such flexibility?  Right now, when I use a different JDK, IntelliJ updates this file and I need to be careful not to commit the changes.

msridhar:
Fixes #2014

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -62,7 +62,7 @@
       </value>
     </option>
   </component>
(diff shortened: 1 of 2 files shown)


liblit (approved):
Thanks for including documentation!

PR #2016, opened 2026-07-22
Add parentheses to clarify operator precedence

liblit:
I generally prefer to assume that developers are skilled professionals who know how to use complex language constructs correctly.  But I have to admit that the operator precedence in these specific cases can be non-obvious.


msridhar (approved):
Yeah these are better with parens!

PR #2021, opened 2026-07-22
Remove redundant guards, dead assertions, and tautologies

liblit:
Several code paths had assertions or conditional branches that were redundant with immediately preceding checks, unreachable logic, or conditions already guaranteed by enclosing context.

### DIFF
--- a/core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/PointerAnalysisImpl.java
+++ b/core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/PointerAnalysisImpl.java
@@ -418,11 +418,10 @@ public class PointerAnalysisImpl extends AbstractPointerAnalysis {
     MutableSparseIntSet s = MutableSparseIntSet.makeEmpty();
     for (InstanceKey element : ik) {
       int index = instanceKeys.getMappedIndex(element);
-      if (index != -1) {
-        s.add(index);
-      } else {
-        assert index != -1 : "instance " + element + " not mapped!";
+      if (index == -1) {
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/PointerAnalysisImpl.java
```diff
@@ -418,11 +418,8 @@ private OrdinalSet<InstanceKey> toOrdinalSet(InstanceKey[] ik) {
     MutableSparseIntSet s = MutableSparseIntSet.makeEmpty();
     for (InstanceKey element : ik) {
       int index = instanceKeys.getMappedIndex(element);
-      if (index != -1) {
-        s.add(index);
-      } else {
-        assert index != -1 : "instance " + element + " not mapped!";
-      }
+      assert index != -1 : "instance " + element + " not mapped!";
+      s.add(index);
```
msridhar: This causes a change in behavior when assertions are off, yes?  I'd be fine with not using `assert` and instead throwing an exception when `index != -1` here.

msridhar (commented):
One comment

core/src/main/java/com/ibm/wala/ipa/callgraph/propagation/PointerAnalysisImpl.java
liblit: Fair point. I'll change this code in the pull request so that it throws an exception when `index` is -1.

msridhar (approved):
One comment

PR #2022, opened 2026-07-24
Enable `AlmostJavadoc` Error Prone check and fix all violations

liblit:
Move `AlmostJavadoc` from the disabled-by-default list to the error-level list, then remove every existing non-Javadoc `@see` comment (`/* (non-Javadoc) @see ... */`) across 183 files.

These comments are auto-generated stubs from Eclipse's Override/Implement Methods code generator.  Beyond being unnecessary boilerplate, many were stale:  their `@see` targets had gone invalid after renamings and refactorings, and some were truncated or empty outright.  Even when correct, the comments would add no value as real Javadoc: the HTML renderer already links to the overridden method automatically, and modern IDEs provide easy navigation to the same. Removing these unhelpful comments declutters the code and lets Error Prone enforce the absence of almost-Javadoc going forward.


msridhar (approved):
Nice!

PR #2023, opened 2026-08-02
Fix zizmor warnings

msridhar:
Fix warnings from [zizmor](https://zizmor.sh/) in our GH Actions config.  Specifically, this pins various actions to a specific SHA, and runs `actions/checkout` with `persist-credentials: false`, reduces the permissions available with the GH token, and changes how some env vars are expanded.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -31,9 +31,11 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - name: Check out WALA sources
-        uses: actions/checkout@v6
+        uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6.1.0
+        with:
+          persist-credentials: false
       - name: 'Set up JDKs'
-        uses: actions/setup-java@v5
+        uses: actions/setup-java@b6effb05e454b25005698d916606bdc6ffcbf961 # v5.7.0
         with:
           # Install all the JDK versions we use in the matrix.  We do JDK 26 last to use that
           # to run Gradle.
@@ -44,7 +46,7 @@ jobs:
             26
           distribution: 'temurin'
       - name: Setup Gradle
-        uses: gradle/actions/setup-gradle@v6
+        uses: gradle/actions/setup-gradle@9c971963bec38e04b3d30dcc455b5382be2fdbfb # v6.3.0
         with:
           build-scan-publish: true
           build-scan-terms-of-use-url: https://gradle.com/terms-of-service
@@ -64,7 +66,7 @@ jobs:
           aggregatedJavadocs javadoc build ${{ env.OS_SPECIFIC_GRADLE_TASKS }} checkGitCleanliness
       - name: Upload test results
         if: always()
-        uses: actions/upload-artifact@v7
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: Test results for JDK ${{ matrix.java }} on ${{ matrix.os }}
           path: |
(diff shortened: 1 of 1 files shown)


liblit:
Should zizmor checks run along with our other build-time linters like Spotless, ShellCheck, LintMarkdown, etc.?

liblit:
Or if running zizmor via Gradle on developer machines is inconvenient, there's also [a zizmor GitHub action](https://docs.zizmor.sh/integrations/#via-zizmorcorezizmor-action).

msridhar:
Here are the two remaining `zizmor` warnings:

```
help[template-injection]: code injection via template expansion
  --> .github/workflows/continuous-integration.yml:64:15
   |
63 |         run: >-
   |         --- this run block
64 |           ${{ env.XVFB_RUN }}
   |               ^^^^^^^^^^^^ may expand into attacker-controllable code
   |
   = note: audit confidence → High
   = help: audit documentation → https://docs.zizmor.sh/audits/#template-injection

help[template-injection]: code injection via template expansion
  --> .github/workflows/continuous-integration.yml:68:48
   |
63 |         run: >-
   |         --- this run block
...
68 |           aggregatedJavadocs javadoc build ${{ env.OS_SPECIFIC_GRADLE_TASKS }} checkGitCleanliness
   |                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ may expand into attacker-controllable code
   |
   = note: audit confidence → High
   = help: audit documentation → https://docs.zizmor.sh/audits/#template-injection

7 findings (5 suppressed): 0 informational, 2 low, 0 medium, 0 high
```

I think the risk here is pretty low (zizmor marks them as low).  The env vars are used basically to avoid some duplication in our config and have just one "Build and test using Gradle" step:

https://github.com/wala/WALA/blob/cc22260990af558027be54f769be5622816a251a/.github/workflows/continuous-integration.yml#L52-L64

We could get rid of the env vars by just duplicating the "Build and test using Gradle" step, with one for Linux and one for non-Linux.  Or maybe there is a way to suppress these warnings.  @liblit thoughts on this one?

msridhar:
> Or if running zizmor via Gradle on developer machines is inconvenient, there's also [a zizmor GitHub action](https://docs.zizmor.sh/integrations/#via-zizmorcorezizmor-action).

I think this is probably the easiest way to go and the right approach once we get the warnings fixed.

liblit:
> Here are the two remaining `zizmor` warnings:
> 
> ```
> help[template-injection]: code injection via template expansion
>   --> .github/workflows/continuous-integration.yml:64:15
> [...]
> help[template-injection]: code injection via template expansion
>   --> .github/workflows/continuous-integration.yml:68:48

Do these Zizmor warnings go away if we follow the advice given in the first green "Tip" box in [the Zizmor Remediation advice](https://docs.zizmor.sh/audits/#remediation_29) for this rule? In other words, is Zizmor content if we

* replace `${{ env.XVFB_RUN }}` with `$XVFB_RUN`
* replace `${{ env.OS_SPECIFIC_GRADLE_TASKS }}` with `$OS_SPECIFIC_GRADLE_TASKS`

If not, then I suggest that we figure out how to suppress these warnings. I consider that preferable to duplicating steps.

msridhar:
> Do these Zizmor warnings go away if we follow the advice given in the first green "Tip" box in [the Zizmor Remediation advice](https://docs.zizmor.sh/audits/#remediation_29) for this rule? In other words, is Zizmor content if we
> 
> * replace `${{ env.XVFB_RUN }}` with `$XVFB_RUN`
> * replace `${{ env.OS_SPECIFIC_GRADLE_TASKS }}` with `$OS_SPECIFIC_GRADLE_TASKS`
> 
> If not, then I suggest that we figure out how to suppress these warnings. I consider that preferable to duplicating steps.

This mostly worked, thanks!  I also had to add `shell: sh` to avoid having to special case for Powershell on Windows.

liblit:
> This mostly worked, thanks!  I also had to add `shell: sh` to avoid having to special case for Powershell on Windows.

Excellent! Thanks for putting in that extra effort to fix these warnings the right way.

liblit (approved):

PR #2024, opened 2026-08-03
Use different site for Kawa 3.0 download

liblit:
`ftpmirror.gnu.org` has been failing with HTTP error 502 (Bad Gateway) for the last few days.  Download the artifact we need from a different site that still works.

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -88,7 +88,7 @@ interface GenerateHelloHashServices {
 //  download and extract kawa 3.0 "kawa.jar"
 //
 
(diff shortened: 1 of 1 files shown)


msridhar (approved):
I was hoping it would come back, but it hasn't. Hopefully with build caching this won't put too much load on the main site

liblit:
[This Reddit post](https://www.reddit.com/r/gnu/comments/1vbpesg/issues_with_ftpmirrorgnuorg/) described the same problem three days ago, but there are no comments providing additional information yet.

PR #2027, opened 2026-08-04
Pin / update uses in publish-test-results.yml

msridhar:
Discovered by zizmor


msridhar:
Will remove this workflow instead

PR #2031, opened 2026-08-06
Add JMH microbenchmarks for the `Atom` class

liblit:
These benchmarks allowed me to evaluate two experimental `Atom` changes: (1) using a Guava `Interner` instead of a `HashMap` and (2) retaining the `HashMap` but adding items to it using `HashMap.computeIfAbsent`. Both regress the common-case interning path compared with the baseline `HashMap` with `get`/`put` operations, so we won't be adopting either change at this time.

This change does not retain JMH benchmark results: data is stored in a Git-ignored `core/build` subdirectory.  There is no support for detecting performance regressions as a gate for pull requests.  Any of that could potentially be added in the future.  For now, though, the JMH infrastructure is just available for individual developers to use in their local development environments.

### DIFF
--- a/README-Gradle.md
+++ b/README-Gradle.md
@@ -218,6 +218,31 @@ by shortening each camel-case-delimited word in its name.  For
 example, the `processTestResources` task can probably be abbreviated
 as `procTeRes` or even `pTR`.
 
+#### JMH Microbenchmarks
+
+[JMH](https://github.com/openjdk/jmh) microbenchmarks are kept alongside
+the unit tests for the code they benchmark, under `src/jmh/java` in the
+relevant sub-project.  For example, the `Atom` microbenchmarks live in
+`core/src/jmh/java/com/ibm/wala/core/util/strings/AtomBenchmark.java`.
+
+Run all microbenchmarks of a sub-project with the `jmh` task.  For
+example:
+
+```shell
+./gradlew :core:jmh
+```
+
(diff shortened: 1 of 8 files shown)


README-Gradle.md
```diff
@@ -218,6 +218,31 @@ by shortening each camel-case-delimited word in its name.  For
...
+[JMH](https://github.com/openjdk/jmh) microbenchmarks are kept alongside
+the unit tests for the code they benchmark, under `src/jmh/java` in the
+relevant sub-project.  For example, the `Atom` microbenchmarks live in
+`core/src/jmh/java/com/ibm/wala/core/util/strings/AtomBenchmark.java`.
+
+Run all microbenchmarks of a sub-project with the `jmh` task.  For
+example:
+
+```shell
+./gradlew core:jmh
```
msridhar: Should this be?

```suggestion
./gradlew :core:jmh
```

msridhar (dismissed):
Looks great!  One suggested fix (if it's actually an issue, not sure)

README-Gradle.md
liblit: `:core:jmh` is an absolute task path; `core:jmh` is relative. They are equivalent in context, but I'll change to the absolute path for clarity and for consistency with a comment in `AtomBenchmark.java`. Good eye!

msridhar (approved):

msridhar (approved):

PR #2036, opened 2026-08-08
Allow scope files to reference classes and sources in JARs

liblit:
In an analysis scope file, 'classFile' and 'sourceFile' entries previously could only name files that exist on the filesystem.  They now also work when the named class or source file is a classpath resource, including a resource packaged inside a JAR.

The broader context here is that we plan to add JMH macrobenchmarks that directly call existing test methods. Previously these tests ran with their test resources available on the classpath as unpacked directories. However, when JMH runs those benchmarks, everything (including the test resources) is bundled up into one or more JAR archives. This change lets those tests find and use their resources using the standard JMH classpath without messy Gradle gymnastics.

`FileProvider.getURLFromClassLoader()` resolves a scope-file pathname to a URL, searching the classloader first and then falling back to the filesystem. `AnalysisScopeReader` now uses `ClassFileURLModule` and `SourceURLModule` for 'classFile' and 'sourceFile' entries, respectively, reading the class or source bytes from the resolved URL instead of requiring a real file. `SourceURLModule` gains a `getClassName()` derived from the URL's entry name or pathname, as `ClassLoaderImpl.loadAllSources` needs it.

Adds `SourceMapTest.testHelloFromJar`, which loads the 'hello' test subject only from a JAR.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -84,6 +84,18 @@ outside WALA, please [let the WALA maintainers
 know](https://github.com/wala/WALA/issues/new) so that we don't remove them in a
 future release.
 
+#### Analysis scope files may reference classes and sources packaged in JARs
+
+In an analysis scope file, `classFile` and `sourceFile` entries previously
+could only name files that exist on the filesystem.  They now also work when
+the named class or source file is a classpath resource, including a resource
+packaged inside a JAR.  For example, an entry such as
+`Application,Java,classFile,hello/Hello.class` now resolves
+`hello/Hello.class` via the classloader, whether it is an unpacked file or an
+entry of `hello.jar`.  This is particularly useful when running an analysis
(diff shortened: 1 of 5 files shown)


msridhar (approved):
Nice!  Do you think there's a need to also update the wiki docs at https://github.com/wala/WALA/wiki/Analysis-Scope?  If so, maybe you can take a look?  That can be done after this PR lands

liblit:
> Do you think there's a need to also update the wiki docs at https://github.com/wala/WALA/wiki/Analysis-Scope?

Ah yes, good point. [Done.](https://github.com/wala/WALA/wiki/Analysis-Scope/_compare/825b769f4be127f205e3363179d3ecaac3d3eeda...1f555297b90d15bee3b8dbfaf270fbe948044e2f) Thanks!

PR #2038, opened 2026-08-09
Tune JMH microbenchmark settings for regression detection

liblit:
Empirically tune the `AtomBenchmark` JMH configuration so that repeated runs are precise enough to detect relative regressions.  With a single fork, several methods reported errors of 5-11%, too noisy to catch a meaningful slowdown.  A single fork's iteration-to-iteration noise averages out, but between-fork variance dominates for several methods, so each method now runs in three forks, with three one-second warmup and three one-second measurement iterations.  That cuts reported errors to roughly 1-2%.

A few methods get their own overrides.  `findOrCreateFromByteArray` and `concatAtoms` each use four forks because their means drifted most between forks in repeated runs.  `atomHashCode` uses six measurement iterations because its throughput periodically dips, apparently due to garbage collection, even after warmup.

`findOrCreateMiss` is a single-shot benchmark whose per-operation cost is not steady until roughly eight or more iterations into each fork, so it now uses ten warmup and ten measurement iterations instead of three of each.

### DIFF
--- a/core/src/jmh/java/com/ibm/wala/core/util/strings/AtomBenchmark.java
+++ b/core/src/jmh/java/com/ibm/wala/core/util/strings/AtomBenchmark.java
@@ -286,8 +286,9 @@ public class AtomBenchmark {
    * Reads an {@link Atom}'s cached hash code.
    *
    * <p>More measurement iterations than the class default, because this benchmark's throughput
-   * periodically dips (apparently GC-related) even after warmup, and extra iterations average those
-   * dips out.
+   * periodically dips (for an unidentified reason, possibly an OS or scheduler transient; the hot
+   * path allocates nothing, so GC is unlikely) even after warmup, and extra iterations average
+   * those dips out.
    */
   @Benchmark
   @Measurement(iterations = 6, time = 1)
--- a/core/src/jmh/java/com/ibm/wala/core/util/strings/AtomBenchmark.java
+++ b/core/src/jmh/java/com/ibm/wala/core/util/strings/AtomBenchmark.java
@@ -42,15 +42,25 @@ import org.openjdk.jmh.infra.Blackhole;
  * interns a bounded fresh pool of content for each measurement iteration, resetting the dictionary
  * first so that every iteration starts from the same empty state.
  *
+ * <p>Even so, the per-operation cost is not steady from the very first iterations: in repeated runs
+ * the per-operation time only settles to its steady state after roughly eight or more iterations in
+ * each fork. {@link #findOrCreateMiss} therefore uses more warmup and measurement iterations than
+ * the class default.
+ *
  * <p>Run with {@code ./gradlew :core:jmh}. Results are written to {@code core/build/results/jmh/}.
- * Increase the fork count (e.g., by changing the {@code @Fork} annotations or passing {@code -f} on
- * the JMH command line) for more precise numbers.
+ * These settings trade total run time for precision, so that they can detect relative regressions:
+ * each method runs in three forks, with three one-second warmup iterations and three one-second
+ * measurement iterations. A single fork is measurably noisier (the per-fork mean is stable, but
+ * between-fork variance dominates for several methods), so three forks cut the reported error from
+ * up to ~10% down to ~1-2%. A few methods with idiosyncratic noise get their own overrides; see
+ * their Javadoc. Pass {@code -f}, {@code -wi}, {@code -i}, and similar JMH command-line flags for
+ * one-off adjustments.
  */
 @BenchmarkMode(Mode.Throughput)
+@Fork(3)
+@Measurement(iterations = 3, time = 1)
 @OutputTimeUnit(TimeUnit.MICROSECONDS)
-@Warmup(iterations = 5, time = 1)
-@Measurement(iterations = 5, time = 1)
-@Fork(1)
+@Warmup(iterations = 3, time = 1)
 public class AtomBenchmark {
 
   /** Number of distinct, never-before-interned byte arrays in each miss-benchmark iteration. */
@@ -174,8 +184,14 @@ public class AtomBenchmark {
     }
   }
 
-  /** Looks up an already-interned {@link Atom} from its bytes. */
+  /**
+   * Looks up an already-interned {@link Atom} from its bytes.
+   *
+   * <p>One more fork than the class default, because this benchmark's mean has drifted more between
+   * forks than the other microbenchmarks in repeated runs.
+   */
   @Benchmark
+  @Fork(4)
   public Atom findOrCreateFromByteArray(AtomFixture fixture) {
     return Atom.findOrCreate(fixture.nextByteArray());
   }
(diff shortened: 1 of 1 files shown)


msridhar:
What kind of machine are we doing these measurements and seeing this noise on?  If it's a Mac, I wonder if CPU power management settings / core assignment could be playing a role?  Would be great to have a bit of a better understanding of what is going on.  Setting up a machine environment to have as little noise as possible is a black art as far as I understand, and we don't want to go too far down that road.  I just don't quite grok the differences here.

msridhar:
If it's helpful, here's a ChatGPT transcript on this, with links to a couple of documents:

https://chatgpt.com/share/6a78ebe0-e3f0-83e8-b44b-4f32c7d54fc7

It does seem like reducing noise is more painful on Mac vs Linux.  But, if we're mostly on Mac, we'll just have to do our best.

liblit:
My tuning experiments were largely delegated to the OpenCode agent using the OpenCode Zen/Big Pickle model. From the same session that ran these experiments and suggested the changes, here's a recap of what it did, what it saw, and what measures it took to control for confounding factors:

---

# Recap: `AtomBenchmark` JMH configuration changes

### Goal
Make the `AtomBenchmark` JMH config suitable for **relative regression detection** — i.e., reliably catching a real slowdown when the `Atom` implementation changes, not measuring absolute performance. The constraint is total wall-clock time: enough precision per run to trust a change, without making `./gradlew :core:jmh` impractically slow.

### Hardware & environment
- Apple M4 Pro, 12 cores (8 performance + 4 efficiency), 48 GB RAM, macOS 26.5.2, arm64.
- JMH 1.37 running on Zulu OpenJDK 17.0.19 (the toolchain the Gradle JMH task forks; the Gradle daemon itself ran on JDK 27-ea but never inside the measured region).
- Benchmarks run with JMH defaults for threads (1 thread per fork).

### Starting point (baseline)
The then-default config was `@Fork(1) @Warmup(5,1s) @Measurement(5,1s)`, `Mode.Throughput`, `MICROSECONDS`. One full run took ~3:51. Reported relative errors across the 14 throughput methods ranged from **0.5% (`toString`) to 11% (`rIndex`)**, with `startsWith` 8%, `atomHashCode` 6%, `findOrCreateFromByteArray` 5%, `findOrCreateUnicodeAtom` 4.6%. Five of fourteen methods were too noisy to detect a small regression; `findOrCreateMiss` was worse (0.031 ± 0.238 µs/op — error ~8× the mean).

### Experiments and findings
I drove JMH directly against the fat `com.ibm.wala.core-1.9.0-SNAPSHOT-jmh.jar` to sweep `-f/-wi/-i/-w/-r` without Gradle overhead.

1. **`-f 4 -wi 3 -i 3` (12 samples/benchmark)** collapsed errors to 0.5–3.3%. Decomposing per-fork vs per-iteration variance showed two distinct noise structures:
   - **`getVal`, `rIndex`, `startsWith`, `findOrCreateFromByteArray`**: between-fork variance dominates (e.g., `getVal` fork means 1021.7, 1038.3, 1039.2, 1039.0). More forks are the fix.
   - **`atomHashCode`**: fork means were extremely tight (1470.7–1473.0) but iteration-level noise was large (σ≈25) — a *periodic dip* pattern (1484, 1487, 1486, 1451, 1462, …) recurring every ~4–5 s. More iterations are the fix.
2. **`-f 1 -i 20`** showed flat trajectories within a fork — no JIT ramp and no drift after warmup — so 5×1 s warmup was already overkill; 3×1 s suffices.
3. **1 s vs 2 s iterations** roughly halved error for `startsWith`/`getValArray`/`rIndex` but at double the runtime; forks buy more precision per minute.
4. **`findOrCreateMiss`** was fundamentally non-steady-state: the interning `HashMap` is cleared and refilled every iteration, and I found *reproducible* step functions (e.g., 0.055 → 0.12 µs/op mid-measurement) in every fork, with the "settled" value depending on warmup length (0.014 vs 0.115 µs/op — an 8× gap). With `@Warmup(10)` it finally reached a steady 0.015 ± 0.001 µs/op across all forks.
5. I also learned the printed JMH error is `t(n−1, 99.9%) × stderr` — with n=3 that t-multiplier is ≈31.6, which is why tiny single-shot runs print absurd margins.

### How findings map to changes
- **`@Fork(1)` → `@Fork(3)`**, **`@Warmup/Measurement(5)` → `(3)`**: three forks directly attack the dominant between-fork variance, cutting errors to ~1–2% at ~2× the old throughput runtime (validated: 4:19).
- **`@Fork(4)` on `concatAtoms` and `findOrCreateFromByteArray`**: the two methods whose means drifted most between forks across repeated runs.
- **`@Measurement(6)` on `atomHashCode`**: extra iterations average out the periodic dips; 6% → 1.2% error.
- **`@Warmup(10)`/`@Measurement(10)` on `findOrCreateMiss`**: gets past the reproducible ~8-iteration transient to a steady state (error 767% → 5.6%).

### Concerns: power management, core selection, thermal throttling
Honest assessment of what could have affected the numbers and what the evidence shows:

- **Power management.** I did *not* verify the Mac's power/energy setting, did not use `caffeinate`, and did not pin anything. Default macOS behavior can drop clocks under light load or on battery. That said, the *within-session* numbers were stable: between-fork means sat within ~1–2%, and no run showed a bimodal or chaotic distribution. A large power-state swing would have inflated exactly the between-fork spread the config now samples against. JMH's forking is what makes the mean robust to individual "bad" forks, and the higher fork count directly hedges this.
- **Core selection (P vs E cores).** The M4 Pro has 8 P + 4 E cores; with `threads=1` each benchmark thread could theoretically land on an E-core (macOS exposes no userland affinity). Had some forks run on E-cores, these compute-bound methods would have measured roughly half the throughput — producing a clearly bimodal fork-mean distribution. **No bimodality was observed** (tight fork means everywhere, e.g., `atomHashCode` spanning 0.16%), which is strong evidence the scheduler consistently kept benchmark threads on P-cores for this workload. This is not guaranteed on other machines or under load, though.
- **Thermal throttling.** I did not monitor temperature or frequency (`powermetrics`) during the runs. However, the flat 20-iteration trajectories (no downward drift over ~20 s) and the fact that means for the same method matched across the hour-plus experiment (e.g., `toUnicodeString` ≈111 µs/op in both the first and last runs) argue against sustained throttling. The periodic `atomHashCode` dips are transient ~2% events; notably, that benchmark allocates nothing in its hot path, so "apparently GC-related" (my committed comment) is probably wrong — transient OS/power/scheduler events are a more plausible cause. Either way, extra iterations average them out regardless of cause, so the config choice is robust.

**Bottom line:** the changes improve *relative* precision — more independent forks and more iterations reduce the variance that power/scheduler/thermal effects contribute, without depending on controlling the machine. For strongest regression-detection practice I'd additionally recommend: run on an idle, AC-powered machine with `caffeinate -dimsu`, keep the macOS/power setting constant, and interleave A/B versions (e.g., old-then-new-then-old) to cancel slow drift; optionally verify frequency stability with `powermetrics`. If you'd like, I can also amend the `atomHashCode` comment to avoid the GC attribution since the cause is actually uncertain.

msridhar:
Cool!  Not an expert here, but if the noise is down that's a good thing 🙂 That's an interesting distinction between tuning to maximize regression detection vs accurately measuring peak performance.  The tradeoffs seem reasonable.

msridhar (approved):

PR #2041, opened 2026-08-10
Adapt two dataflow tests as dataflow macrobenchmarks

liblit:
Split `DataflowTest.testContextSensitive` and `DataflowTest.testContextInsensitive` into `prepare...` phases (analysis scope, class hierarchy, call graph, and pointer analysis) and `compute...` phases (the reaching-definitions analyses themselves), so that a new JMH benchmark can run all the setup once per fork and time only the analyses.  Set `ContextInsensitiveReachingDefs.VERBOSE` to `false` so the timed `analyze()` doesn't print to standard output on every invocation.

Each analysis takes roughly a quarter of a millisecond on the `dataflow.StaticDataflow` subject (measured means of about 0.25 ms for the context-insensitive analysis and 0.23 ms for the context-sensitive one).  These are sub-millisecond workloads, so single invocations are noisy:  allocation and GC jitter are on the same scale as the work itself, and the per-invocation time keeps dropping until about the sixtieth invocation as the JIT progressively compiles the analysis code. Benchmark accordingly:  single-shot time mode, sixty warmup invocations, twenty measured invocations, and four forks.  This reproduces the means to within about 5–9%, and repeated runs agree within that noise (e.g., 0.255 and 0.250 ms for the context-insensitive mean and 0.236 and 0.224 ms for the context-sensitive mean across two consecutive runs).

### DIFF
--- a/core/src/jmh/java/com/ibm/wala/benchmarks/dataflow/DataflowBenchmark.java
+++ b/core/src/jmh/java/com/ibm/wala/benchmarks/dataflow/DataflowBenchmark.java
@@ -1,0 +1,103 @@
+package com.ibm.wala.benchmarks.dataflow;
+
+import com.ibm.wala.examples.analysis.dataflow.DataflowTest;
+import com.ibm.wala.ipa.cha.ClassHierarchyException;
+import com.ibm.wala.util.CancelException;
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+import org.openjdk.jmh.annotations.Benchmark;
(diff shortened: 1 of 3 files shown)


msridhar (approved):
If it's helpful in the future I'm sure we can find a way to make the analysis time longer for the benchmarks by involving the standard library somehow

liblit:
> If it's helpful in the future I'm sure we can find a way to make the analysis time longer for the benchmarks by involving the standard library somehow

Certainly! These are the _first_ versions of WALA macrobenchmarks, not the _last_ versions.

PR #2043, opened 2026-08-14
Intern `Atom`s using a lock-free map

liblit:
Use a JCTools `NonBlockingHashMap` to canonicalize `Atom`s, so the common "already interned" lookup path no longer needs a class-level monitor. `putIfAbsent` preserves the one-canonical-instance invariant when threads race to intern the same content.

See #2042 for alternatives considered and performance evaluation.


msridhar (approved):
Very nice improvement!

liblit:
> Very nice improvement!

Thank you for challenging me to consider performance more carefully in my changes. 🤝

PR #2048, opened 2026-08-26
Cache `ExplicitCallGraph.ExplicitNode.hashCode`

liblit:
The old implementation recomputed `getMethod().hashCode() * 8681 + getContext().hashCode()` on every call.  However, `CGNode` objects are canonical per (method, context) pair, both fields are `final` and never change, and `equals()` is identity-based.  Therefore, caching the hash in a `final` field is safe and avoids repeated (and re-derived) hash computations.  Every node is created through `findOrCreateNode`, which rejects `null` methods, so eager computation in the constructor cannot NPE.

Effect on JMH benchmarks (JDK 17; 3 forks × 6 one-second iterations per configuration after warm-up; comparisons against `master` using per-fork means as the unit of analysis, Welch's t-test, one-sided in the improvement direction):

* `DataflowBenchmark.analyzeTestContextInsensitive`
  * baseline: 0.2934 ms/op (99% CI [0.2792, 0.3077])
  * patched:  0.2158 ms/op (99% CI [0.2027, 0.2288])
  * 26.5% faster, p = 0.005
* `DataflowBenchmark.analyzeTestContextSensitive`
  * baseline: 0.2233 ms/op (99% CI [0.2145, 0.2321])
  * patched:  0.2136 ms/op (99% CI [0.1993, 0.2278])
  * 4.4% faster, p = 0.046

No other benchmark showed a statistically significant difference in either direction.

A note on multiple comparisons: we ran 19 hypothesis tests per branch, each at the nominal alpha = 0.05 level.  Even if this change had no real effect anywhere, uncorrected testing would give roughly a 62% chance of at least one spurious "significant" result (1 - 0.95^19), so raw p-values overstate our confidence.  Applying a Holm step-down correction across those 19 tests, the context-insensitive improvement has an adjusted p ("q") of 0.19:  above the conventional 0.05 cutoff, though its large ~27% effect size and disjoint 99% confidence intervals leave little practical doubt that it is real.  The context-sensitive result does not survive any correction and should be treated as suggestive rather than conclusive.

These gains are consistent with async-profiler data showing `ExplicitNode.hashCode` as a significant fraction of dataflow-analysis CPU; the same hash chain also feeds `LocalPointerKey.hashCode` during pointer analysis.

No API change; backward compatible.


msridhar (approved):
Nice!

PR #2050, opened 2026-08-31
Tune JMH configurations for stable measurements

liblit:
Increase forks, warmup iterations, and measurement iterations across all JMH benchmarks so the suite can reliably detect ~2% performance changes.  Remove per-method fork overrides whose concerns are now covered by the higher class defaults.  Update Javadoc to match.

Suite wall-clock time increases from ~7 minutes to ~20 minutes: +170% or ~2.7×.  In my opinion, that extra time is acceptable in exchange for letting us make more trustworthy decisions about performance impacts of future changes.


msridhar (approved):
100% agreed that more time for more accurate measurements is worthwhile.  NullAway's JMH suite takes almost 1 hour to run, as a comparison.

PR #2051, opened 2026-08-31
Reduce `int[]` growth churn in `MutableSparseIntSet`

liblit:
Double the initial backing array size (2 → 4) and increase the growth factor (1.5 → 2).  Both changes reduce the number of reallocations when a `MutableSparseIntSet` grows, which is the dominant cost in points-to-set and dataflow analysis.

This change yields large gains on `IntSet`-heavy workloads:

- `SlicerBenchmark.computeBackwardSliceInTestList`: ~19% faster
- `DataflowBenchmark.analyzeTestContextSensitive`: ~11% faster


msridhar (approved):
Yeah, for modern JVMs and heap sizes this probably makes more sense.  Any impact on the call graph benchmark?

liblit:
> Any impact on the call graph benchmark?

No clear impact that stands out from the noise.

PR #2052, opened 2026-08-31
Optimize binary search in `MutableSparseIntSet.add`

liblit:
Replace the linear scan used to find the insertion point in `MutableSparseIntSet.add` with a binary search via the new `IntSetUtil.binarySearchInsertionPoint` helper, which delegates to `java.util.Arrays.binarySearch`.


msridhar (commented):
Looks good!  I'm curious, are the performance improvement numbers given here in addition to those from #2051?  Or were these numbers in comparison to the master branch without #2051?

msridhar:
I'm also curious about the reduction in code coverage for `IntSetUtil`, though sadly codecov is not able to show me the differences right now...

liblit:
> are the performance improvement numbers given here in addition to those from #2051? Or were these numbers in comparison to the master branch without #2051?

Astute question! The latter, and that's a problem. I expected that this PR would show even better improvements when applied to the larger arrays introduced by #2051. However, I'm seeing evidence that the opposite is true. Please hold off on approving this PR until I can do a more thorough investigation. We may discover that each change is an improvement by itself, but not when taken together.

liblit:
OpenCode helped me with a more extensive analysis of this PR by itself, PR #2052 by itself, and both taken together. In brief, **neither of these changes has a significant impact on performance, and both changes taken together are likewise without impact**. I propose that we merge this PR anyway: it doesn't _hurt_ benchmark performance, and its reduced asymptotic complexity could (in theory) benefit large analyses where the `IntSet`s get much bigger.

<details>
<summary>OpenCode's analysis summary</summary>

# Performance Analysis: Optimizations to `MutableSparseIntSet`

## Background

[WALA](https://github.com/wala/WALA) is a static analysis framework for Java bytecode. One of its core data structures, `MutableSparseIntSet`, is used for integer set operations during analysis (e.g., atom interning, call graph construction). Two optimizations were proposed and benchmarked:

1. **Reduce array growth churn**: When the set resizes, instead of allocating a new array and copying, the implementation now reuses the existing array when possible. This reduces garbage collection pressure.

2. **Optimize binary search**: The `add` method uses binary search to find insertion points; this was optimized for better cache behavior.

Both optimizations were tested individually and combined.

## Benchmark Methodology

- **Benchmark**: JMH (Java Microbenchmark Harness), throughput mode (`ops/us`, higher = better)
- **Target**: `atomHashCode` benchmark — a micro-benchmark that exercises the core `IntSet` operations
- **Configuration**: 4 forks, 5 warmup iterations, 6 measurement iterations, 1s each
- **GC configurations**: Default GC (with `-Xbatch`) and Serial GC (`-XX:+UseSerialGC`)
- **JVM**: Azul Zulu OpenJDK 17.0.19, aarch64 (Apple Silicon)

## Results Summary

| Configuration | Baseline | Array growth optimization | Binary search optimization | Both optimizations |
|---|---|---|---|---|
| **Default GC** | 1491 ± 12 | 1397 ± 45 ⚠ | 1313 ± 274 ⚠ | 1492 ± 12 |
| **Serial GC** | 1483 ± 18 | 1484 ± 17 | 1419 ± 84 ⚠ | 1486 ± 14 |

⚠ = Results contaminated by external interference (see below)

## Key Findings

1. **No statistically significant performance difference**: All four implementations (baseline, array growth, binary search, and both combined) perform within measurement noise of each other across all 16 benchmarks tested.

2. **No regressions detected**: The optimizations do not introduce performance overhead in any measured scenario.

3. **Possible minor improvement**: The combined implementation showed a 3% faster score on the end-to-end `CallGraph` benchmark (8171 ms vs 8436 ms baseline), but this is within the confidence interval overlap and not yet conclusive.

4. **Data quality note**: The default GC runs for the individual optimizations (array growth and binary search) were contaminated by external system interference (likely thermal throttling or background processes), producing bimodal score distributions. The Serial GC runs and the combined-optimization default GC runs were clean and reliable.

## Conclusion

**All three optimization candidates are performance-equivalent to the baseline.** The optimizations are safe to merge from a performance perspective — they neither help nor hurt in these benchmarks. The decision to merge should be based on other factors (code quality, maintainability, theoretical benefits for other workloads).

</details>

msridhar (approved):

